PR #7, opened 2022-11-30
Feature: add Fonts based on nextjs13

### LINKED ISSUE #6: Fonts using Nextjs 13
Next comes with a new way of using google fonts  [@next/font](https://nextjs.org/docs/api-reference/next/font)

Currently app just uses one global font, we need to see and use multiple fonts dynamically and take advantage of the package provided by Nextjs

ggwilin:
## Pull Request type

- [x] :sparkles: Feature

Fonts didn't load using Nextjs13 Font Optimization  [@next/font](https://nextjs.org/docs/api-reference/next/font)

## What is the new behavior?

- Fonts load from a single class 
- Tailwind theme is extended
- We only need to use `font-(my-preferred-font)` in the utility classes

## Close issue

resolved #6

- [ ] Yes
- [x] No


components/Fonts.tsx
```diff
@@ -0,0 +1,39 @@
+import { Caveat_Brush as cBrush, Inter, Nunito } from '@next/font/google'
+
+const cbrush = cBrush({
+    weight: '400',
+    variable: '--cbrush-font',
+    preload: true,
+    display: 'swap',
+})
+
+const inter = Inter({
```
GBKS, line 10: I don't think we'll need. We'll need Space Mono though, we're using that for rendering code.

GBKS:
Looks good to me. I have not input on how to include fonts in this project, really your call. 

Only thing I saw is that we don't need the "Inter" font, but we need "Space Mono" for rendering code.

Issue #1, opened 2022-11-01
Login and sharing identity


adamjonas:
As a learner, I can access my account by logging in with my 'private key'.

As a learner, I can share my identity with other players via my 'public key'.

GBKS:
We will have the first ask in V1. Sharing public keys is a bigger design and tech process we'll have to tackle separately at a later stage.

aureleoules:
The way I would implement this is to store the hash of the private key in the localStorage to serve a 'public key' when the user signs up, along with the user progress and avatar.
I don't think it's worth it to derive the actual secp256k1 public key and not sure how easy that is in javascript also.

When the user tries to login, they will provide their private key and we will hash it to make sure it matches our public key.

And is there a need for the public key (in a chapter for example)? Maybe we can just store the private key as is.

aureleoules:
I'll work on "As a learner, I can access my account by logging in with my 'private key'.".

PR #37, opened 2022-12-19
Save progress and login

### LINKED ISSUE #31: Add account access to the `NavBar` component
image

image

The nav bar should show a blank avatar icon when logged out, and the avatar image the user has picked when logged in.

Clicking this icon should show the account overlay to log in or out.

### LINKED ISSUE #32: Make avatars selectable in signup modal
image

The user should be able to pick one of those avatars to identify them in the future.

This modal component is currently called `SaveProgressButton`, which is not quite right. I'd rename to to `SignupModal` or something like it.

[Design in Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=393%3A10917&t=Ir9L2sg7RUjtqKyc-1).

### LINKED ISSUE #33: Develop the login modal
This allows users to log back in and load their progress using their personal code ([designs](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=393%3A10941&t=Ir9L2sg7RUjtqKyc-1)).

image

### LINKED ISSUE #36: Check that personal code (private key) is within secp256k1 range
Users are able to save their progress with a personal code (aka a private key) that is generated in [`components/chapters/SaveProgressButton.tsx`](https://github.com/saving-satoshi/saving-satoshi/blob/master/components/chapters/SaveProgressButton.tsx#L41)

Screenshot from 2022-12-16 14-04-18

This key is a random 64 digit hex number (aka 256 bits).

As a user, I would like to use this code to generate addresses of my own in chapter 3. In order for this to work, the secret key must be smaller than 1.1578 * 10^77 (a number that is slightly less than 2^256). 

I'm not sure what the chances are of a random 256 bit number being larger than the acceptable range. That will probably determine the priority of this ticket. However, if we don't add a check, it will result in some kind of instability in the chapter 3 challenges, because those challenges have the user generate a public key (aka a point on the secp256k1 elliptic curve) from this secret key. 

Possibly a good first issue?

aureleoules:
Fixes #32.
Fixes #36.
Fixes #33.
Fixes #31 
Addresses #1.

- Moved `SaveProgressButton.tsx` to `SignUpModal.tsx`
- Avatar is now selectable
- Callback function to retrieve the selected avatar and code
- Generate secp256k1 private key instead of generic random hex number
- Store progress, avatar and public key in localStorage
- Login modal 
- Added avatar in Navbar

## Pull Request checklist

- [ ] Lint has passed locally and any fixes were made for failures


components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,112 @@
+'use client'
+
+import Image from 'next/image'
+import { useState } from 'react'
+import Modal from 'react-modal'
+import crypto from 'crypto';
+import Secp256k1 from '@lionello/secp256k1-js';
+
+export const LoginModal = (props) => {
+  const customStyles = {
```
GBKS, line 10: Let's use Tailwind instead of these custom styles.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+            <div className="flex w-full mt-4">
+              <button onClick={loadProgress} className="border-white border-2 w-full p-1 text-xl">Load my progress</button>
+            </div>
+          </div>
+        </>
+      ) : (
+        <>
+          <Image
+            alt="avatar"
+            src="/assets/avatars/1.png"
```
GBKS, line 96: This should probably be the current users avatar and not always the same one.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+            width={80}
+            height={80}
+            className="rounded-full mb-2"
+          />
+          <h2 className="mb-2 text-3xl font-bold">Progress loaded</h2>
+          <p className="mb-5 text-xl">
+            {"You're all set to continue with Chapter 3, Challenge 2."}
```
GBKS, line 103: How can we determine the users progress? Since we also need to show this on the chapters overview page (and probably elsewhere), is it best to create a utility class for this?

components/chapters/SignUpModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+    const publicKey = Secp256k1.generatePublicKeyFromPrivateKeyData(privateKey);
+
+    return {
+      privateKey,
+      publicKey
+    };
+  }
+
+  const customStyles = {
```
GBKS, line 22: Also change to Tailwind.

components/chapters/SignUpModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  return (
+    <Modal
+      isOpen={props.open}
+      style={customStyles}
+      contentLabel="Example Modal">
+
+        <div className="float-right flex justify-end">
+          <button onClick={props.onClose}>
```
GBKS, line 57: Eventually we want proper aria attributes for buttons to make the UI accessible. For now, a title tag should be enough to explain what this button does..

components/ui/NavBar.tsx
```diff
@@ -13,9 +30,9 @@ export const Navbar = ({ items }: { items: NavItem[] }) => {
         >
           <h1 className="text-xl md:text-3xl">{siteConfig.name}</h1>
         </Link>
+        <ul className="flex text-xl md:text-2xl">
```
GBKS, line 33: The `<ul>` tag can be a `<nav>`, and then we don't need the `<li>` tags. A cleaner structure.

components/ui/NavBar.tsx
```diff
@@ -24,10 +41,44 @@ export const Navbar = ({ items }: { items: NavItem[] }) => {
                   {item.title}
                 </Link>
               </li>
-            ))}
-          </ul>
+            ))
         ) : null}
+        {}
+        {!loggedIn && <li onClick={() => setOpenSignInModal(true)} className="text-grey-300 w-10 h-10 ml-4 cursor-pointer">
```
GBKS, line 47: Should be a `<button>` tag since it's clickable.

components/ui/NavBar.tsx
```diff
@@ -24,10 +41,44 @@ export const Navbar = ({ items }: { items: NavItem[] }) => {
...
+            <rect width="30" height="30" fill="white" transform="translate(0 0.5)"/>
+            </clipPath>
+            <clipPath id="clip1_393_10135">
+            <rect y="0.5" width="30" height="30" rx="15" fill="white"/>
+            </clipPath>
+            </defs>
+          </svg>
+        </li>}
+        
+        {loggedIn && <li className="text-grey-300 w-10 h-10 ml-4 cursor-pointer">
```
GBKS, line 67: Should also be a `<button>` tag. Also, we need to figure out what happens when I am logged in and click the avatar - maybe another modal to show my key and the option to log out?

GBKS (changes requested):
Just a few points I came across when reviewing. I need to design another modal state for when you're logged in already, and the error state for when your key doesn't match.

components/chapters/LoginModal.tsx
aureleoules: I've updated it to show the user's progress.

image
But I don't think this is ideal, we should probably have a utility class for this in deed.

components/ui/NavBar.tsx
aureleoules: > another modal to show my key and the option to log out?

Yes I agree.

components/chapters/SignUpModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+import Modal from 'react-modal'
+const crypto = require('crypto')
+const Secp256k1 = require('@lionello/secp256k1-js')
+
+export const SignUpModal = (props) => {
+  function generateKeyPair() {
+    const privateKeyBuf = crypto.randomBytes(32)
+    const privateKey = Secp256k1.uint256(privateKeyBuf, 16)
+
+    const publicKey = Secp256k1.generatePublicKeyFromPrivateKeyData(privateKey);
```
satsie: This isn't really actionable feedback, but I want to make note (really a note to myself) of storing the public key. In chapter 3 I have an open question about manually deriving the public key from the private key. 

Recall that the public key is actually a point on the elliptic curve. Public key = private key * Generator point

I believe for ECDSA signatures, the public key is 1 byte (indicating if the y coordinate is even or odd) plus 32 bytes for the x coordinate in full.

However for Schnorr, I am reading that the public key is just the x coordinate (32 bytes). Someone please correct me if my understanding is wrong!

This all may lead to a slight discrepancy in what happens in lesson 3 and what we have stored for the public key. Anyways this is not really related to PR review, but something I need to keep in mind!

components/chapters/SignUpModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+export const SignUpModal = (props) => {
+  function generateKeyPair() {
+    let publicKey = null;
+    let privateKey = null;
+    do {
+      const privateKeyBuf = crypto.randomBytes(32)
+      privateKey = Secp256k1.uint256(privateKeyBuf, 16)
+      publicKey = Secp256k1.generatePublicKeyFromPrivateKeyData(privateKey);
+    } while (!publicKey || privateKey.gte(Secp256k1.uint256("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141", 16)))
```
satsie, line 17: Thanks for adding this check! I would assume that the library you are using has this covered since it's specific to secp256k1, but I think it's still a good idea to have this safety net, just to be double sure.

satsie:
While I can't speak to any of the front end stuff, I am able to ACK the public/private key login stuff for 83a13ed

components/chapters/SignUpModal.tsx
aureleoules: Public keys for ECDSA and Schnorr are both an X and a Y coordinate (32 bytes each).
But yes you are correct, in Bitcoin ECDSA public keys are compressed into 33 bytes: the X coordinate and the parity byte (whether Y is odd or not)
For Schnorr the compressed key is 32 bytes, only the X coordinate is kept and the tie-breaker is implicit.

For lesson 3, since we are deriving ECDSA keys, the user should decide what parity byte to use (either `0x02` if Y is even, otherwise `0x03`. Or we let the user generate uncompressed keys.

components/chapters/SignUpModal.tsx
aureleoules: I originally thought this too but after looking at the code (https://github.com/lionello/secp256k1-js/blob/master/src/secp256k1.js), the curve order doesn't seem to be checked so I added it.

components/chapters/SignUpModal.tsx
satsie: oooo thank you for the clarification! Would there be any reason to let the user generate a Schnorr public key? 

I am leaning towards ending the lesson at the uncompressed keys, just to avoid this altogether :sweat_smile:

aureleoules: > Would there be any reason to let the user generate a Schnorr public key?

I would love to support Schnorr but I believe it is much more complex to derive a P2TR address from a Schnorr public key than a legacy P2PKH from an ECDSA public key. So that may not be very fun for the user.

It shouldn't be too much trouble to ask the user which parity byte to use. Maybe allow the user to run `Y % 2` in the terminal to check if Y is odd or not, or let the user decide randomly until the challenge accepts the solution.

satsie (approved):

satsie:
@GBKS gentle reminder to approve this if you feel the changes you requested have been resolved. While we can bypass the branch protections on the merge, I'd like to avoid if possible :)

components/chapters/SignUpModal.tsx
adamjonas: I think we stick with uncompressed to start and work our way down in complexity.

aureleoules:
I added a few more states, based on the figma file. https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=1112%3A14225&t=8GV6lHISj7LrNNnh-0

Logged in:
image
image

Load button disabled when no input:
image

Incorrect key:
image

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+              <input
+                className="bg-transparent outline-none text-white border-dotted border-white border-2 p-1 w-full text-lg"
+                type="text"
+                placeholder="Enter your code"
+                value={userPrivateKey}
+                onChange={(e) => setPrivateKey(e.target.value)}
+              />
+            </div>
+            <div className="flex w-full mt-4">
+              <button disabled={!userPrivateKey} onClick={loadProgress} className={`border-white border-2 w-full p-1 text-xl ${!userPrivateKey && 'opacity-50'}`}>Load my progress</button>
```
GBKS, line 117: This can be a BoxButton.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  return (
+    <Modal
+      isOpen={props.open}
+      className="top-1/2 left-1/2 outline-none p-5 transform -translate-x-1/2 -translate-y-1/2 absolute w-[32rem] rounded-lg shadow-lg bg-[#32455D] text-white"
+      contentLabel="Login Modal"
+    >
+      <div className="float-right flex justify-end">
+        <button onClick={props.onClose} aria-label="Close">
+          <svg
```
GBKS, line 80: In another comment, you had mentioned an `Icon` class. Could we use this approach here?

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+      {!loaded && !props.signedIn && (
+        <>
+          <img src="/assets/avatars/1.png" width={80} height={80} />
+          <h2 className="mb-4 mt-5 text-3xl font-bold">Load your progress</h2>
+          <p className="mb-5 text-xl">
+            Enter your personal code below to restore your chapter and challenge progress, as well as your avatar. <a href="#" className='text-white underline'>No code yet?</a>
+          </p>
+
+          <div className="flex flex-col justify-between">
+            <div className="flex w-full">
```
GBKS, line 107: Why do you need these wrapper div tags around every UI element?

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+            <div className='flex items-center mt-auto'>
+                <p>
+                  Want to start again? <button onClick={clearProgress} className='underline'>Clear saved progress</button>.
+                </p>
+            </div>
+          </div>
+        </>)}
+      {props.signedIn && (
+        <>
+          <Image
```
GBKS, line 141: Is it worth putting creating an avatar component? We'll show it quite often in the mining lesson in a few sizes and could bake those options in.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+          />
+          <h2 className="mb-2 text-3xl font-bold">Load your progress</h2>
+          <p className="mb-5 text-xl">
+            Your progress is stored in this browser. Use the code below to restore it after taking a break.
+          </p>
+
+        <pre className="mb-5 flex flex-col rounded-md border-2 border-dotted border-white/25 p-4">
+          <code className="whitespace-pre-wrap break-all">{props.user.privateKey}</code>
+          {copied ? (
+            <button
```
GBKS, line 156: This can be a BoxButton. May need to add a "success" or "clicked" state though.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+              className="mt-4 w-full rounded-md bg-green-500 px-4 py-2 text-green-800"
+              onClick={copy}
+            >
+              Copied!
+            </button>
+          ) : (
+            <button
```
GBKS, line 163: This can be a BoxButton.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+              className="mt-4 w-full rounded-md bg-[#233346] px-4 py-2 text-white"
+              onClick={copy}
+            >
+              Copy
+            </button>
+          )}
+        </pre>
+        <div className='flex items-center mt-auto'>
+          <button onClick={signOut} className="border-red-500 bg-red-500 border-2 w-full p-1 text-xl text-white">Sign out</button>
```
GBKS, line 172: This can be a BoxButton.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+            width={80}
+            height={80}
+            className="rounded-full mb-2"
+          />
+          <h2 className="mb-2 text-3xl font-bold">Progress loaded</h2>
+          <p className="mb-5 text-xl">
+            {`You're all set to continue with Chapter ${user.progress.chapter}, Challenge ${user.progress.lesson}.`}
+          </p>
+
+          <button onClick={props.onLogin} className="border-white border-2 w-full p-1 text-xl">Continue</button>
```
GBKS, line 190: This can be a BoxButton.

components/ui/NavBar.tsx
```diff
@@ -13,21 +44,60 @@ export const Navbar = ({ items }: { items: NavItem[] }) => {
...
+                className="text-white text-opacity-75 transition duration-150 ease-in-out hover:text-opacity-100 mr-2"
+              >
+                {item.title}
+              </Link>
+            ))
         ) : null}
+        {}
+        {!loggedIn && <button onClick={() => setOpenSignInModal(true)} className="text-grey-300 w-10 h-10 ml-4 cursor-pointer">
+          <svg width="30" height="31" viewBox="0 0 30 31" fill="none" xmlns="http://www.w3.org/2000/svg">
+            <g clip-path="url(#clip0_393_10135)">
```
GBKS, line 62: It's a bit messy to have these clip-paths in inline SVGs. I need to clean up the vectors in Figma so we can export clean ones. Also, can we refer to SVG asset files and have them automatically pulled in during compile/build?

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+        </button>
+      </div>
+
+      <div className="p-4">
+      {!loaded && !props.signedIn && (
+        <>
+          <img src="/assets/avatars/1.png" width={80} height={80} />
+          <h2 className="mb-4 mt-5 text-3xl font-bold">Load your progress</h2>
+          <p className="mb-5 text-xl">
+            Enter your personal code below to restore your chapter and challenge progress, as well as your avatar. <a href="#" className='text-white underline'>No code yet?</a>
```
GBKS, line 103: The regular body text should be Nunito Semi-Bold.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+          </p>
+
+          <div className="flex flex-col justify-between">
+            <div className="flex w-full">
+              <input
+                className="bg-transparent outline-none text-white border-dotted border-white border-2 p-1 w-full text-lg"
```
GBKS, line 109: The font for code input should be Space Mono.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+  function signOut() {
+    setLoaded(false);
+    props.onLogout();
+    props.onClose();
+  }
+
+  return (
+    <Modal
+      isOpen={props.open}
+      className="top-1/2 left-1/2 outline-none p-5 transform -translate-x-1/2 -translate-y-1/2 absolute w-[32rem] rounded-lg shadow-lg bg-[#32455D] text-white"
```
GBKS, line 75: Instead of hard-coding colors, we need to find a way to assign dynamic colors via CSS variables, based on the color theme of the lesson ([example](https://github.com/GBKS/bitcoin-designathon/blob/main/assets/css/variables.scss#L10)). Seems like Tailwind has [support](https://tailwindcss.com/docs/adding-custom-styles#resolving-ambiguities), but we need to figure out how to assign these variables based on the current lesson.

For example, each content .mdx file could have a meta attribute called theme. That theme could be applied to the whole page as a class. That class could then adjust the variables accordingly.

GBKS (changes requested):
Great progress. There are still various things to tweak, which we can do in this PR, or separately.
- Some fonts are not right
- Should we introduce an `Icon` class or a way to reference SVG assets instead of copying SVG code inline?
- Some buttons can be `BoxButtons`
- Do we need an `Avatar` class?
- We need to find a way to use CSS variables based on the color theme of the current lesson

Let me know if you have questions, and want to get everything in this PR or merge the main functionality and then update things in new PRs.

components/chapters/LoginModal.tsx
aureleoules: I like to keep UI elements separated by divs but I can change it if you want.

aureleoules:
Thanks for your feedback @GBKS! Yes I suggest we clean up the code in separate follow-ups.
I agree we should :
- Add an avatar class
- Add an Icon class
- Add more states to the BoxButton

aureleoules:
I updated the fonts.

components/chapters/LoginModal.tsx
aureleoules: I'm not familiar with next and the tooling used in this project so I haven't been able to import svgs from files, which would be better than inlining svgs in a big Icon class I think.
I created #47.

GBKS: In #49 I added SVGR to allow us to keep SVG code in separate files. Nice and easy.

components/chapters/LoginModal.tsx
GBKS: Why?

components/chapters/LoginModal.tsx
GBKS: I introduced theme color variables in #45.

components/ui/NavBar.tsx
aureleoules: Done in ed38633f9b207f81ef94253542245d3e6c91529a

components/chapters/LoginModal.tsx
aureleoules: Done in 3d0cd738f917ab419773aee1fd932256d118f13c.

components/chapters/LoginModal.tsx
aureleoules: I have difficulties making it work with the hover class and can't override it easily.
I think a follow-up PR should improve this class to allow for other background colors and other hover states.

components/chapters/LoginModal.tsx
aureleoules: Sometimes it's easier to wrap form components with a div to add styles in css so it became a habit.

components/chapters/LoginModal.tsx
GBKS: 100%. I created issue #50 for this.

GBKS:
Great updates. Let me know if the PR is ready for another review or if you're still working on things.

components/chapters/LoginModal.tsx
aureleoules: Done

aureleoules:
Ready for another review round! I rebased on top of master to use the color variables.

adamjonas:
Avatars not highlighting and modal not collapsing once I've copied


https://user-images.githubusercontent.com/755825/211050636-7697b356-dfbf-41e9-b36e-d2a48806abb8.mov

aureleoules:
Weird, both work for me. Tested on firefox and google-chrome.
It looks like this
image

Do you have any errors in the console? 

@GBKS can you reproduce the issue?

components/ui/NavBar.tsx
```diff
@@ -1,9 +1,42 @@
+'use client';
+
 import Link from 'next/link'
 
 import { siteConfig } from 'config/site'
 import { NavItem } from 'types'
+import { useEffect, useState } from 'react'
+import { LoginModal } from 'components/chapters/LoginModal';
+import Image from 'next/image';
+import User from 'public/assets/icons/user.svg';
```
GBKS, line 10: We already have avatar.svg in the repo, you can use that.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  function signOut() {
+    setLoaded(false);
+    props.onLogout();
+    props.onClose();
+  }
+
+  return (
+    <Modal
```
GBKS, line 77: Need to find a way to change the background color of the modal.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,177 @@
...
+      <div className="float-right flex justify-end">
+        <button onClick={props.onClose} aria-label="Close">
+          <Close className="h-6 w-6" />
+        </button>
+      </div>
+
+      <div className="p-4">
+      {!loaded && !props.signedIn && (
+        <>
+          <Avatar avatar={1} size={80} />
```
GBKS, line 91: The avatar here should be dynamic and use the one in the local storage.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,177 @@
...
+          <h2 className="mb-4 mt-5 text-3xl font-bold">Load your progress</h2>
+          <p className="mb-5">
+            Enter your personal code below to restore your chapter and challenge progress, as well as your avatar. <a href="#" className='text-white underline'>No code yet?</a>
```
GBKS, line 94: Need to hook up the link here to switch to the SignupModal.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+      {loaded && !props.signedIn && (
+        <>
+          <Avatar
+            avatar={user.avatar}
+            size={80}
+          />
+          <h2 className="mt-2 mb-2 text-3xl font-bold">Progress loaded</h2>
+          <p className="mb-5">
+            {`You're all set to continue with Chapter ${user.progress.chapter}, Challenge ${user.progress.lesson}.`}
```
GBKS, line 168: Chapter and lesson should be numbers, not IDs.

components/ui/NavBar.tsx
```diff
@@ -1,9 +1,42 @@
...
+  }, [])
+
+  useEffect(() => {
+    const user = window.localStorage.getItem('user');
+    if (user) {
+      setUser(JSON.parse(user));
+    }
+  }, [loggedIn])
+
+  function onLogin() {
```
GBKS, line 30: We should navigate user to their latest lesson when clicking continue.

components/chapters/LoginModal.tsx
```diff
@@ -0,0 +1,177 @@
...
+                onChange={(e) => setPrivateKey(e.target.value)}
+              />
+            </div>
+            <div className="flex w-full mt-4">
+              <BoxButton full size='small' style='outline' disabled={!userPrivateKey || !checkValidKey()} onClick={loadProgress} classes={`border-white border-2 p-1 text-xl md:w-full ${!userPrivateKey && 'opacity-50'}`}>Load my progress</BoxButton>
+            </div>
+            <div className="flex w-full mt-4 mb-4">
+              {userPrivateKey && !checkValidKey() && (
+                <div className='flex items-center'>
+                  <Warning />
```
GBKS, line 113: `warning.svg` should use `currentColor` for color definitions, so the value can be overwritten with CSS. Makes the icons more flexible.

GBKS (approved):
Great work with this PR.

I left a few comments and have a follow-up list of visual tweaks around font sizes and spacing. We'll also need to switch out buttons with the updated BoxButton with the green style that I'm working on.

Let me know if you want to keep tweaking this PR, or feel free to merge it and we can tweak things in follow-ups.

components/chapters/LoginModal.tsx
aureleoules: Done

components/ui/NavBar.tsx
aureleoules: Done

aureleoules:
I addressed some of the nits, I suggest doing the rest as follow-ups.

PR #44, opened 2022-12-22
Fixing a few CSS object propery names

GBKS:
The console keeps complaining about these.


satsie (approved):
Yay for bug squashing!

aureleoules (approved):
ACK 528b1a86c6e19d3a4094a57bf67c4199203575c3

But shouldn't we replace the style object by tailwind classes?

GBKS:
@aureleoules sure thing, that would be better. The Terminal needs a lot of work still. I am going to merge this one so we get rid of the console errors for now.

PR #45, opened 2023-01-02
Introduce themes using CSS variables

GBKS:
Chapters have unique background colors that get picked up by certain UI elements. This PR introduces a basic system using a `back` (= background) variable that gets changed by appending a theme class to the current page. Related to [this comment](https://github.com/saving-satoshi/saving-satoshi/pull/37#discussion_r1060170784).

Themes also include a background pattern, which is not included here. Something to add separately, we need to figure out the best way (optimizing file size and a fade on load) to implement this.

Here's a mock-up with an example of how the theme is expressed. See how the label on the button picks up the background color. Not that this is not needed in many places and it's fine to just overlay white or black with some opacity.
image

PR #37 includes various hardcoded color values that instead should use theme variables.

### DIFF
--- a/components/ui/BoxButton.tsx
+++ b/components/ui/BoxButton.tsx
@@ -29,9 +29,9 @@ export const BoxButton = ({ href, children, title, style, size, disabled, extern
     className += ''
 
     if(disabled) {
-      className += ' bg-white/50 text-black/50'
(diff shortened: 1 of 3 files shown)


aureleoules (approved):
Looks good to me.

PR #48, opened 2023-01-03
Add tabs to genesis lesson on mobile

GBKS:
This PR introduces a tabbed layout on mobile where the user can switch between seeing the lesson info or code. Both at once can only be seen on desktop (no tabs there).

👳🏻[Check the preview](https://saving-satoshi-git-feature-genesis-respon-44c3c2-saving-satoshi.vercel.app/chapters/chapter-1/genesis)👲🏻

- Adds the 'react-responsive' module to more easily handle screen-size specific layouts
- Adds tabs to the genesis page on mobile only
- Splits up the genesis lesson into sub-components so they can be individually shown and hidden based on which tab is selected
- Lots of style tweaks to more closely match the Figma file

Designs:
image

It's a very common thing to show slightly different UI on mobile and desktop, but it seems to take a lot of extra code overhead to do this. First we need the `react-responsive` library. Then we need to hide stuff until the page is rendered once before showing content to avoid hydration errors (server-rendered HTML won't match client-rendered HTML because the server doesn't know what the client environment is). Would be awesome if this code could be packaged into a helper tool or so, but I don't know React well enough.

I am also curious if we can re-use layout styling. We have a few page templates for lessons. Ideally we can avoid copy/pasting the same layout Tailwind classes (e.g. `flex grow md:items-center max-w-[840px] px-4 py-8`) all over the place and instead have a single class like `-lesson-wrap` that bundles these definitions. Will have to see what a good way is to do that. That can be a separate PR.

### DIFF
--- a/app/chapters/[slug]/genesis/info.tsx
+++ b/app/chapters/[slug]/genesis/info.tsx
@@ -1,0 +1,31 @@
+import { BoxButton } from 'components/ui/BoxButton'
+
+export const Info = ({
+  isSmallScreen,
+  genesis
+}: {
+  isSmallScreen: boolean,
+  genesis: any
+}) => {
+  return (
+    <div className="
+      flex
+      flex-col
+      gap-6
+      grow
+      px-4
+      py-6
+      max-w-[840px]
+      md:justify-center
+    ">
+      <div
+        className="genesis text-white font-nunito"
+        dangerouslySetInnerHTML={{ __html: genesis.body.html }}
+      ></div>
+      <BoxButton
+        href="https://blockstream.info/block/000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f"
+        external={true}
+      >View Block 0</BoxButton>
+    </div>
+  )
+}
--- a/app/chapters/[slug]/genesis/info.tsx
+++ b/app/chapters/[slug]/genesis/info.tsx
@@ -1,10 +1,8 @@
 import { BoxButton } from 'components/ui/BoxButton'
 
 export const Info = ({
-  isSmallScreen,
   genesis
 }: {
-  isSmallScreen: boolean,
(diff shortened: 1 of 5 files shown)


app/chapters/[slug]/genesis/info.tsx
```diff
@@ -0,0 +1,31 @@
+import { BoxButton } from 'components/ui/BoxButton'
+
+export const Info = ({
+  isSmallScreen,
+  genesis
+}: {
+  isSmallScreen: boolean,
```
Comment (aureleoules) [inline app/chapters/[slug]/genesis/info.tsx:7]: This doesn't seem used.

aureleoules (approved):
Looks good.

I'm a bit skeptical about adding a js lib to handle to responsiveness but it does seem to make it easier to conditionally display some elements so I'm ok with it.

app/chapters/[slug]/genesis/info.tsx
Comment (GBKS) [reply app/chapters/[slug]/genesis/info.tsx]: I removed that line.

I also think it's a bit overkill to add a dedicated library for this. Just trying to do things the react way, and this is what was recommended in the blog posts I found.

app/chapters/[slug]/genesis/page.tsx
```diff
@@ -23,103 +21,57 @@ function getGenesis() {
...
+          items={tabData}
+          activeId={activeTab}
+          onChange={setActiveTab}
+          classes="px-4 py-2 w-full"
+          stretch={true}
+        />
+      )}
+      {(!isSmallScreen || (isSmallScreen && activeTab == 'info')) && (
+        <Info
+          isSmallScreen={isSmallScreen}
```
Comment (aureleoules) [inline app/chapters/[slug]/genesis/page.tsx:63]: should also remove this

Comment (GBKS) [reply app/chapters/[slug]/genesis/page.tsx]: Done.

aureleoules (approved):

PR #72, opened 2023-01-09
Add Finish Chapter view

### LINKED ISSUE #53: Add 'You did it' view after a chapter completion
image

aureleoules:
Fixes #53.
This PR adds the view for the chapter completion.
After completing chapter 1 - transacting-2, the save progress modal is no more shown. Instead there is a "Next" button that leads to this page where the progress can be saved.
image

image
image


app/chapters/[slug]/done/page.tsx
```diff
@@ -0,0 +1,48 @@
...
+                <h1 className="text-4xl font-bold text-white">
+                    You did it!!!
+                </h1>
+                <p className="text-xl text-white font-nunito">
+                    Amazing. You completed the first chapter and learned a lot about hashes and transactions. Are you ready for more?
+                </p>
+
+                <div className="flex flex-col gap-4 mt-4">
+                    <BoxButton
+                        size='small'
```
Comment (GBKS) [inline app/chapters/[slug]/done/page.tsx:36]: Why is this one small?

app/chapters/[slug]/done/page.tsx
```diff
@@ -0,0 +1,48 @@
...
+          progress: {
+            chapter: 'transacting',
+            lesson: 'transacting-2',
+          }
+        }))
+        setOpenModal(false)
+    }
+
+    return (
+        <div className="flex w-screen grow bg-fixed bg-cover bg-no-repeat" style={{backgroundImage: "url('/assets/images/chapter-finish.jpg')"}}>
```
Comment (GBKS) [inline app/chapters/[slug]/done/page.tsx:24]: The layout breaks on mobile - the screenshot below is on 390x844 (iPhone 12 Pro size). The image should also center on mobile, so the character remains visible.

localhost_3000_chapters_chapter-1_done(iPhone 12 Pro)

GBKS (changes requested):
Thank you for tackling this one. Had a bit of feedback, and also a question.

Is the file location of this page in `chapters/[slug]/done` appropriate for when we have more chapters? This may also be relevant to the other pages. I wonder if we won't end up getting away from `chapters/[slug]` using wildcards towards `chapters/chapter-1` folders, so we can better distinguish the chapter-specific code. What do you think?

app/chapters/[slug]/done/page.tsx
Comment (aureleoules) [reply app/chapters/[slug]/done/page.tsx:36]: Both are small, would you prefer them not to be?

aureleoules:
I made the page a reusable component so that each chapter end can be customized and made the page responsive.

aureleoules:
> I wonder if we won't end up getting away from chapters/[slug] using wildcards towards chapters/chapter-1 folders, so we can better distinguish the chapter-specific code. What do you think?

I absolutely agree, I just opened #79 for this.

app/chapters/[slug]/done/page.tsx
Comment (GBKS) [reply app/chapters/[slug]/done/page.tsx]: As, sorry, somehow did not see the small tag on the second one.

GBKS (approved):
Nice work, let's get it merged.

I'll do a follow-up with some minor styling tweaks. A little bit spacing, and also placing a subtle dark gradient underneath the text area to create better contrast.

There's also this nice full-screen effect [in Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=71%3A957&t=ETY1a7RtUfXSnqyp-1), where the images goes behind the nav bar. We can look into that at some point in the future. 

Question, do we want to use numbers of chapter/challenge IDs to store progress locally? I believe I've seen both used in different places.

aureleoules:
> Question, do we want to use numbers of chapter/challenge IDs to store progress locally? I believe I've seen both used in different places.

I don't really know what is best. Maybe IDs are more practical because we can map them easily to titles and metadata.

Merging this PR, thanks for the reviews!

GBKS:
Let's use IDs then.

PR #73, opened 2023-01-09
Improve Terminal colors

### LINKED ISSUE #60: Add color coding to the Terminal component
User input should be white, Terminal output should be a theme-based highlight color (for the default theme it's #3DCFEF). [See in Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=1056%3A14226&t=nxCE26gS2fD24CVh-1).

Current:

image

Design:

image

aureleoules:
Fixes #60.

image

### DIFF
--- a/app/chapters/[slug]/genesis/genesis-pt2/page.tsx
+++ b/app/chapters/[slug]/genesis/genesis-pt2/page.tsx
@@ -27,13 +27,13 @@ export default function Genesispt2() {
   const [answer, setAnswer] = useState('')
 
   function onInput(input) {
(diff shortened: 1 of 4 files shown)


GBKS (approved):
Looks great, nice addition.

Once merged, I'll do a follow up with some small tweaks on the `>` character. For the current input it's added via a pseudo element. For completed lines of code it's in the HTML. Result is slightly different spacing that can cause lines to break differently - causing a jumpy effect when hitting enter. I might also turn this into a SVG, will have to see how it looks.

PR #75, opened 2023-01-09
Add <> to the "scriptSigHex" in the example command.

shaavan:
## Pull Request checklist

- [ ] Lint has passed locally and any fixes were made for failures

## Pull Request type

- [ ] :bug: Bugfix
- [ ] :sparkles: Feature
- [ ] :zap: Performance
- [x] :pencil: Documentation content changes

- From how the example command is written, it is not evident that the "scriptSigHex" needs to be replaced by the message found in the last exercise to complete this exercise correctly.

## What is the new behavior?

- This commit adds the carets to the scriptSigHex and makes subsequent changes to the preceding text to make the exercise's goal clearer to understand.

- [ ] Yes
- [x] No

### DIFF
--- a/content/lessons/genesis-pt2.mdx
+++ b/content/lessons/genesis-pt2.mdx
@@ -5,8 +5,8 @@ contentType: 'challenge'
 
 # Let's make sense of this
 
-Use the following command to decode Satoshi's secret HEX message into ASCII, a more human readable format:
+To decode Satoshi's secret hidden in the message you saw in the last exercise, use the following command and replace \<scriptSigHex\> with the message to interpret it.
 
 ```shell
-echo scriptSigHex | xxd -r -p
+echo <scriptSigHex> | xxd -r -p
 ```
--- a/content/lessons/transacting-2.mdx
+++ b/content/lessons/transacting-2.mdx
@@ -12,5 +12,5 @@ Once you've found it, decode the hidden message, just like we did in the previou
 And remember, 'Bitcoin is for everyone.'
 
 ```shell
-echo scriptSigHex | xxd -r -p
+echo <scriptSigHex> | xxd -r -p
 ```


content/lessons/genesis-pt2.mdx
```diff
@@ -5,8 +5,8 @@ contentType: 'challenge'
 
 # Let's make sense of this
 
-Use the following command to decode Satoshi's secret HEX message into ASCII, a more human readable format:
+To decode Satoshi's secret hidden in the message you saw in the last exercise, use the following command and replace \<scriptSigHex\> with the message to interpret it.
```
GBKS, line 8: I like how you're trying to remove some of the technical terms, but it's not consistent right now with the previous screen. The term `message` only appears here, the previous screen does not refer to any messages. Maybe review both screens together (as well as the next challenge which is also about decoding messages) and see what works across them all?

`ScriptSig HEX Representation` really is a mouthful, I hope we can find something simpler.

satsie: Appreciate the effort here to make things more clear to the user :) Regarding discussions that involve copy, what does everyone think about moving those to Figma? We're lucky for now that we have a designer that reviews PRs, but that may not always be the case.

I don't have an issue with the original copy, but the proposed change is a bit long. Is there a way we can break it up into two, maybe three smaller sentences? I also agree with Christoph on the use of the term 'message'.

Something like: "Remember the message from the last exercise? Let's decode the secret that Satoshi left in it. Use the following command and replace \<scriptSigHex\> with the message. We need to do this because that message is currently in HEX format. This command will decode it to ASCII, something that is much easier for humans to read."

Another way we can mitigate some confusion is also through creative use of tool tips, which is more than just a simple copy change. This is another reason why I think Figma might be a better place for these kinds of discussions.

GBKS: Do we need to propose a process for submitting copy edits? I also think it's better to keep those discussion out of GitHub.

satsie: yeah! let's put it on the agenda for tomorrow's call

adamjonas:
Converting to draft while we answer some of the questions posed in the above comments.

satsie:
[Per last week's call](https://www.notion.so/January-2023-Meeting-Notes-40780d3fc94a425ca4712f5b17f6e238#9e0086c0b28d46d8998000c95c7dbc2f), we are going to move discussion regarding copy changes to Figma

This PR has two options:

1. Remove the copy changes and we'll work through those in Figma
2. Leave the copy changes in and we can continue discussing them here in GitHub. Since this is the first time a copy change has come up in a PR, I'm okay with making the process exception.

content/lessons/genesis-pt2.mdx
```diff
@@ -5,8 +5,8 @@ contentType: 'challenge'
...
 
 ```shell
-echo scriptSigHex | xxd -r -p
+echo <scriptSigHex> | xxd -r -p
```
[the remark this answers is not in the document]
ecurrencyhodler: I totally agree. If someone could hover over the <scriptSigHex> portion and have a tool tip appear that said something like "Replace this text with the scriptSigHex you copied in the last challenge", it would be a lot clearer.

GBKS:
Quick heads-up that I am looking into a different approach where variables can be pre-defined in the terminal (just like in "normal" coding). 

image

afterburn:
What's the status of this PR @shaavan, is it still relevant after some of the other things that were merged?

PR #76, opened 2023-01-11
Minor color and spacing tweaks to the Terminal

GBKS:
- Adds a first tip line to the Terminals in the two lessons that use it ("Enter your commands here and press Enter...")
- Tweaks colors and spacing of Terminal lines
- Changes ">" to always be in a pseudo-element to avoid glitches when submitting

[Check the preview](https://saving-satoshi-git-feature-terminal-color-tweaks-saving-satoshi.vercel.app/chapters/chapter-1/genesis/genesis-pt2)

### DIFF
--- a/app/chapters/[slug]/genesis/genesis-pt2/page.tsx
+++ b/app/chapters/[slug]/genesis/genesis-pt2/page.tsx
@@ -23,7 +23,10 @@ function getGenesis() {
 export default function Genesispt2() {
   const genesis = getGenesis()
 
-  const [lines, setLines] = useState([])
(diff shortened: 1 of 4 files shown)


aureleoules (commented):
Nice that's better.
Just a nit: I would prefer the user input lines to be white and the current input greyish instead of the other way around. What do you think?

GBKS:
@aureleoules why do you prefer that? My logical reasoning is that the active input should visually be stronger.

aureleoules:
Maybe both white then? I don't want to slow down the development for this.

aureleoules (approved):

PR #77, opened 2023-01-11
Adding green and tiny styles to the BoxButton

GBKS:
Both styles are needed for the copy button in the Signup modal.

Learned a bit about Tailwind here that results in some tweaks. To get opacity with color variables, there needs to be a color transform function in the Tailwind config, so I added that. This way we also don't need to use color variables in Tailwind classes anymore, as the transform adds those.

Also replaced two existing color definitions with our new variable-based ones.


aureleoules:
Nice, looks good and is more practical. It conflicts with master though.

GBKS:
Fixed, thanks.

aureleoules (approved):

PR #85, opened 2023-01-17
Bug: Keep chapter height consistent when tabbing

### LINKED ISSUE #83: Keep chapter height consistent when tabbing (Chapters page)
The height of each tab content is different. So when tabbing, the layout changes (medium-width screens). It should not. Best way to solve this is to place add a container with a the min-height set to the taller height of both tab contents. This requires us to render both (also the one not shown to the user at the moment). Not elegant (let me know if you have a better proposal - fixing the height does not work across responsive screen sizes and dynamic content), but the current layout jump is pretty unpleasant.

**Current behavior:**
localhost_3000_chapters (2)

localhost_3000_chapters (3)

Expected behavior is that the height remains the same, even when tabbing to the challenges.

vinayaksh42:
## Pull Request checklist

## Pull Request type

- [x] :bug: Bugfix
- [ ] :sparkles: Feature
- [ ] :zap: Performance

Chapter height changing upon switching tab in medium-width screens

## What is the new behavior?
issue83before
issue83 after

## Close issue

resolved #83

- [ ] Yes
- [x] No


components/partials/OverviewInfo.tsx
```diff
@@ -33,13 +33,13 @@ export const OverviewInfo = (chapter: Chapter) => {
         onChange={setActiveTab}
         classes="mt-8"
       />
-      <div className="flex grow">
-        {activeTab == 'info' &&
+      <div className="flex">
+        <div className={'tab-content ' + (activeTab != 'info' ? '' : 'tab-content-active')}>
```
GBKS, line 37: Could you please also add `aria-hidden` attributes? Right now, the invisible tab content is still being picked up by screen readers. With `aria-hidden=true`, it will be ignored. [More info here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden).

GBKS (changes requested):
Thank you for working on this. That's a nice trick using `margin-right`. I left a comment about `aria-hidden` attributes, and could you please also only add this logic for medium-large sized screens? On mobile, we want the layout to adjust because the image height is not getting distorted either way.

components/partials/OverviewInfo.tsx
vinayaksh42: I have created a ternary statement based on screen width (1023px), for greater than 1023px it will use the invisible logic and for less than 1023px it will use the existing tab-switching logic.

styles/globals.css
```diff
@@ -196,3 +196,14 @@ div.react-terminal-line {
 div.react-terminal-active-input {
   color: white;
 }
+
+.tab-content {
```
GBKS, line 200: Can you please handle these styles via Tailwind?

GBKS (approved):
ACK. Thank you!

PR #86, opened 2023-01-18
Fix progress tracking logic and show status on lessons

GBKS:
[Here's a short video](https://youtu.be/O_ELxQvR8ps) about this PR that explains it. And [check the preview](https://saving-satoshi-git-feature-fix-progress-l-e74ed0-saving-satoshi.vercel.app/chapters).

This turned into a bigger PR out of necessity. My main goal was to make lessons/challenges display their status of whether the user has unlocked or completed them in the chapters overview page and in the nav bar in lesson content.

This required making the progress tracking consistent based on IDs.

Then I saw that various components were overwriting the user object at different times. Also, we need to track progress before the user has officially "registered" at the end of chapter 1, so we need an extra property to track that. I centralized all the user-related logic in lib/user.ts. That is now the only class that messes with the user object, simplifying logic in various other places.

In order to make the challenge components figure out their status, I created a new function in lib/content.ts called getUserLessonStatus. That way we also don't need to repeat this in multiple components.

I hope this isn't too crazy to review and give feedback. Main thing to test is the register, login, logout, clear functionality, and to go through chapter 1 and seeing whether the chapters overview page and the nav bar in the lesson accurately shows your status (lock means you can't access the lesson yet, a check means you completed it).


aureleoules:
Master
image

This PR
image

image


Also, after completing `chapter-1/transacting/transacting-2`, the transaction challenge is not saved after clicking on "Continue without saving", which kinda makes sense but the genesis challenge is saved on the other hand.
Maybe we should remove the "Continue without saving" and always save the progress in localStorage.

GBKS:
Thanks, feedback addressed. For saving, we just need to clearly differentiate between saving locally, which we want to do at all times, and the user registering by seeing their code and stating that they saved it. Right now, this user functionality is really basic and you could argue that it's not very useful, but it might be more important in the future to have it.

lib/content.ts
```diff
@@ -1,9 +1,64 @@
...
+        result.completed = true
+      }
+
+      if(userLessonIdIndex >= lessonIdIndex) {
+        result.unlocked = true
+      }
+    }
+  } else if(chapterId == 'chapter-1' && lessonId == 'genesis') {
+    result.unlocked = true
+  }
```
aureleoules, line 60: This seems a bit too hardcoded no?

aureleoules (approved):
Looks good to me.

lib/content.ts
GBKS: Yes. It's only because this is an exception to ensure the first lesson in the first chapter can always be accessed no matter what. A loop over chapters/lessons could would be cleaner. Same goes for setting the initial user progress in `user.ts`. Let's clean up in a separate PR.

PR #92, opened 2023-01-19
Add link to bitcoin.design to README, link inline to channel

satsdev:
The direct link to the Slack channel only works if the user is a member of bitcoindesign.slack.com. Otherwise they get an error and have to track down the invite link on the https://bitcoin.design/ website. I added a link to hopefully make it more intuitive.

Alternative changes could be:

- Link directly with the Slack invite URL. Not ideal if the invite link expires or changes in the future.
- Explain how to join the Bitcoin Design Slack in this section. This may be overkill.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,8 @@ Now be quick, Satoshi needs you!
 
 ## Contributing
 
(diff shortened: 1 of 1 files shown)


adamjonas:
ACK

PR #96, opened 2023-01-21
Load default avatar for not logged in users

### LINKED ISSUE #94: Site crashes on first page load
**Describe the bug**
Site crashes on first page load. Refreshing the page fixes the problem.

**To Reproduce**

1. Open home page in incognito mode

**Screenshots and/or logs**

Screenshot 2023-01-20 at 11 34 56 AM

Screenshot 2023-01-20 at 11 40 01 AM

satsdev:
Fix for site crashing immediately after load

Resolves #94 

## Pull Request checklist

## Pull Request type

- [x] :bug: Bugfix
- [ ] :sparkles: Feature
- [ ] :zap: Performance

Site crashes shortly after load.

## What is the new behavior?

Site does not crash.

- [ ] Yes
- [x] No

### DIFF
--- a/components/chapters/LoginModal.tsx
+++ b/components/chapters/LoginModal.tsx
@@ -94,7 +94,7 @@ export const LoginModal = (props) => {
       <div className="p-4">
       {!isLoggedIn && (
         <>
-          <Avatar avatar={user.avatar} size={80} />
+          <Avatar avatar={(user && user.avatar)} size={80} />
           <h2 className="mb-4 mt-5 text-3xl font-bold">Load your progress</h2>
           <p className="mb-5">
             Enter your personal code below to restore your chapter and challenge progress, as well as your avatar. <a href="#" className='text-white underline'>No code yet?</a>


vinayaksh42:
Tested the vercel-preview link in incognito mode, and the site is not crashing. Great fix!

GBKS:
This PR does not address the core issue and breaks the dynamic avatar display for registered users.

The main problem is that the login modal should only ever be shown if the user has actually registered - by going through the signup modal. That means that an avatar was selected and saved, and should be available in the login modal.

I am curious what the root cause here is...

satsdev:
@GBKS Thank you for the catch. I can work on the logic. But, for your assumption, my understanding is that LoginModal will still get rendered even if it isn't displayed since it is included in NavBar. Is that not expected?

image

satsdev:
I amended the commit with better logic that I believe preserves the saved user avatar state as you expected. Instead of passing 1, I let that decision be determined in the Avatar component.

If there is still more to this, let me know.

GBKS (approved):
ACK. Thanks for the update.

We still need to find out why the login modal is shown in the first place without an `avatar` being defined in the `user` object in local storage. I wonder if this has to do with us having a bit of a broken user management logic until recently, maybe this is already fixed and can't happen for completely new visitors to the site... well, we'll find out.

PR #97, opened 2023-01-21
Input area interaction

### LINKED ISSUE #65: Improve interaction model of input area on Copy challenges
The component currently uses the [react-individual-character-input-boxes](https://www.npmjs.com/package/react-individual-character-input-boxes) module. The expected behavior is that of a regular text box, so it feels a bit off when trying to do anything more than a simple paste. 

Let's try a simple text box approach that is styled to still show the underlines (maybe another hidden text area behind the input that intelligently shows the underscored for blank characters).

This affects the two copy challenge pages we have:
- [Genesis](https://savingsatoshi.com/chapters/chapter-1/genesis)
- [Transacting](https://savingsatoshi.com/chapters/chapter-1/transacting/transacting-1)

Screenshot:
image

benalleng:
Adds better userInput with standard textarea styled to mimic the react individual character input boxes 'RICIBs'

Resolves #65

## Pull Request checklist

## Pull Request type

- [ ] :bug: Bugfix
- [x] :sparkles: Feature
- [x] :lipstick: :art: Code style update (formatting, renaming)
- [ ] :zap: Performance

RICIBs gives a lot of footguns to the user to type the answers incorrectly. While this is not a problem in Chapter 1 as the user is only required to paste it could become a problem in higher challenges.

## What is the new behavior?

- Typing and Paste-ing behave like a standard texarea

- [ ] Yes
- [x] No

## Other information

I think it is worth considering utilizing RICIBs or perhaps another custom text area when the user is building their own transactions. The data is broken into chunks and RICIBs are good for allowing the user to type it in however the please as loosely demonstrated below. I still belive in most cases, the new input system is better in most cases

Before:

image

After: 

image
image


benalleng:
I would like to know if we prefer to still have the underscores disappear? just to make it as similar to the previous input field as possible.

satsdev:
Tested on desktop and mobile (IOS). Nicely improved usability on desktop. I did not see any issues besides what you already have listed in TODO.

GBKS:
Thank you a ton for tackling this one, several people have complained about the interaction model of this component. I also realized that I never provided full design specs for this component, so I put something together here (and [in Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=1393%3A15989&t=fF6Oe8NLCx1yFm5D-1)).
image

To your question, I think it's much easier on the eyes when the underscores disappear. It might also be best if we use color to indicate incorrect input (blue) and valid input (green). I prefer blue over red, because it's not really an error (red is always so serious and makes you feel bad).

For mobile, I reduced the font size to 18px (from 30px), line height is at 180% at letter-spacing at 5%. The green and blue should use our standardized theme CSS color variables (they may slightly differ for each theme).

Hope that's useful, let me know how it looks.

components/UserInput.tsx
```diff
@@ -0,0 +1,67 @@
...
+    value: string;
+    userInput: (value: string) => void
+}
+
+export const UserInput: React.FC<UserInputProps> = ({ inputAmount, value, userInput }) => {
+    const [textAreaValue, setTextAreaValue] = useState(value);
+    
+    return (
+        <form style={{position:'relative'}} >
+              <br/>
```
GBKS, line 24: What do you need the line breaks for?

components/UserInput.tsx
```diff
@@ -0,0 +1,67 @@
...
+    inputAmount: number;
+    value: string;
+    userInput: (value: string) => void
+}
+
+export const UserInput: React.FC<UserInputProps> = ({ inputAmount, value, userInput }) => {
+    const [textAreaValue, setTextAreaValue] = useState(value);
+    
+    return (
+        <form style={{position:'relative'}} >
```
GBKS, line 23: Would be amazing if you could use Tailwind instead of styles.

components/UserInput.tsx
```diff
@@ -0,0 +1,67 @@
...
+              <br/>
+              <textarea 
+                onChange={(e) => {setTextAreaValue(e.target.value)
+                    userInput(e.target.value);
+                }}
+                value={textAreaValue}  
+                id="text" name="text"
+                rows={4} cols={40}
```
GBKS, line 31: We may need these to be customizable. Do you know if browser strictly enforce rows and cols or if they can be overwritten with CSS?

components/UserInput.tsx
```diff
@@ -0,0 +1,67 @@
...
+                    top: '0',
+                    left: '0',
+                    zIndex: '1',
+                    outline: 'none',
+                    resize: 'none'
+                }}
+                />
+              <p style={{
+                    display: 'block',
+                    width: '900px',
```
GBKS, line 52: Width should be adjustable per page (maybe via a max-width), it currently causes a horizontal scroll-bar. Ideally, we have roughly equal width rows (e.g. we have 150 characters = 3 rows of 50 on desktop and 5 rows of 30 on mobile) for a consistent look, but I am not sure if we can easily get full control over this and can use max-width as a shortcut.

components/UserInput.tsx
```diff
@@ -0,0 +1,67 @@
...
+                    height: '120px',
+                    lineBreak: 'anywhere',
+                    fontSize: '20px',
+                    color:'white',
+                    letterSpacing: '10px',
+                    position: 'relative',
+                    top: '-20px',
+                    zIndex: '0'
```
GBKS, line 60: You may not need `position:relative` and the `z-index` here if you use `pointer-events: none`, which makes it non-selectable.

components/UserInput.tsx
```diff
@@ -0,0 +1,67 @@
...
+    
+    return (
+        <form style={{position:'relative'}} >
+              <br/>
+              <textarea 
+                onChange={(e) => {setTextAreaValue(e.target.value)
+                    userInput(e.target.value);
+                }}
+                value={textAreaValue}  
+                id="text" name="text"
```
GBKS, line 30: Do we need `id` and `name` attributes?

components/UserInput.tsx
```diff
@@ -0,0 +1,67 @@
+'use client'
+
+import { Space_Mono } from '@next/font/google'
+import { useState } from 'react';
+
+const smono = Space_Mono({
```
GBKS, line 6: Don't think you need this import if you use the `font-space-mono` Tailwind class.

GBKS (changes requested):
Thanks again for tackling this. I reviewed the code and had a few comments. Some around consistency with the rest of the code, and some about making this component more flexible for various inputs and screen dimensions. Let me know if you have questions.

components/UserInput.tsx
benalleng: not needed

benalleng:
>     * Weird offset when removing parts of the last line
>       image

This is currently one of my main blockers. While i can set the test-align-last: center for the underscores I am unsure how to achieve this with a textarea or any kind of editable input. Perhaps this is where Christoph mentioning changing it to be a content editable div would give me better functionality.

ecurrencyhodler:
Another issue is that after pasting in the hex, I am unable to move onto the next chapter.
test

ecurrencyhodler:
I tested this on mobile. Seems like the sizing is off as well. Not sure if we want to address that in this PR. 

But like the updates I see on the web!
IMG_1334

afterburn:
Hey, just sharing my thoughts on this. I think the approach you've chosen is promising. I hope you can resolve the remaining issues using this approach, but if that turns out to be rather difficult we might need to consider using `contenteditable`. In my opinion, the problem with regular inputs and textareas is that they do not provide fine-grained control over individual characters/tokens, which is what we seem to be trying to achieve here.

I'm sure you'll be able to get different colors working for correct/incorrect characters, and get the underscores lined up correctly, but I just wanted to share this [code](https://jsfiddle.net/afterburner/rejbty21/1/) I had laying around from some old, forgotten project of mine, which I modified to more closely resemble the input we're trying to create here. While it's by no means ready for production, it uses a similar approach (overlaying two elements) but I think it provides more control over individual tokens. I don't know, perhaps it could be useful to you.

On the flip side, it's also worth mentioning that `contenteditable` is often way more involved (as you can see from the code example) and it's very prone to bugs, so it might not be worth the effort. I think if we can avoid using `contenteditable` and still meet requirements that would be best.

benalleng:
@ecurrencyhodler Hey I'm having some trouble with the reproduceability of your issues, perhaps you have the settings of an older preview cached? 

image


image

benalleng:
@afterburn I think you're right on the contenteditable div being a better approach afterall, and I appreciate the code you provided!  I had actually begun to refactor to contenteditable div already I just feel that it isn't quite ready to be reviewed. some of your code will certainly be helpful! I think unfortunately creating using a textarea is causing the last couple of blockers to match the figma GBKS posted

benalleng:
I played around with the fiddle a little bit today, got it right here now i just have to refactor it for react :sweat_smile: 
https://jsfiddle.net/0nb8g7mc/

GBKS:
@benalleng is it possible to restrict the characters a user can input to avoid some of those edge cases (and save you a headache)? Seems like it [should be](https://gist.github.com/hygull/356d1c5f043b1564d9fb59e58a490272).

afterburn:
Hey @benalleng, you should be able to prevent certain characters from being input on the keypress event. You can see me doing the same in my fiddle. 😁

ecurrencyhodler:
@benalleng you may need to rebase this with https://github.com/saving-satoshi/saving-satoshi/pull/118 now merged.

benalleng:
@GBKS, @ecurrencyhodler  or @afterburn I have rebased this following #118 and I think its ready for y'all to take a look for review!

ecurrencyhodler:
Hmmm not that I can think of. I just copied and pasted the hex into prod and it worked. Which hex are you using?
Screenshot 2023-02-01 at 10 10 37 AM

content/lessons/genesis/genesis.tsx
```diff
@@ -13,7 +13,6 @@ import {
   LessonPrompt,
 } from 'ui'
 
-const inputAmount = 154
 const answer =
   '04ffff001d0104455468652054696d65732030332f4a616e2f32303039204368616e63656c6c6f72206f6e206272696e6b206f66207365636f6e64206261696c6f757420666f722062616e6b73'
```
benalleng, line 17: I am just grabbing it from the answer posted in the genesis content page, I am going to try and make catch to make sure that only numbers/characters can be pasted in correctly

benalleng:
@ecurrencyhodler I added a regex like function that should remove spaces and line breaks to prevent any invisible characters when pasting

ecurrencyhodler:
tACK! Looks good to me. Great job!
Screenshot 2023-02-01 at 9 21 13 PM

ui/common/Input.tsx
```diff
@@ -1,39 +1,93 @@
...
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLTextAreaElement>) => {
+    if (event.key === ' ') {
+      event.preventDefault()
+    }
+  }
+
+  const blankSpace = () => {
+    let underscores = '_'.repeat(answer.length)
+    underscores = underscores.substring(0, answer.length - textAreaValue.length)
```
afterburn, line 23: Nitpick but you can chain methods on the String class:
```js
const underscores = '_'.repeat(answer.length).substring(0, answer.length - textAreaValue.length)
```
Alternatively you can do:
```js
const underscores = '_'.repeat(Math.max(0, answer.length - textAreaValue.length))
    
```

ui/common/Input.tsx
```diff
@@ -1,39 +1,93 @@
...
+
+interface UserInputProps {
+  onChange: Function
+  answer: string
+  hints?: boolean
+}
+
+export default function Input({ onChange, answer, hints }: UserInputProps) {
+  const [textAreaValue, setTextAreaValue] = useState('')
+  const [correctAns, setCorrectAns] = useState(false)
```
afterburn, line 13: Nitpick but let's try to steer away from abbreviations like ans for answer, these might be confusing for non-english folk and could leave some room for interpretation, which we want to avoid.

ui/common/Input.tsx
```diff
@@ -1,39 +1,93 @@
...
+  }
+
+  const blankSpace = () => {
+    let underscores = '_'.repeat(answer.length)
+    underscores = underscores.substring(0, answer.length - textAreaValue.length)
+    return (
+      <>
+        {Array.from({ length: answer.length }, (_, i) => {
+          if (i >= answer.length - underscores.length)
+            return (
```
afterburn, line 28: Another nitpick but let's always use curly braces, even for switches. Though its not incorrect to do it this way in this specific scenario I think curly braces help prevent issues with scoping, so it's a good practice to have.

ui/common/Input.tsx
```diff
@@ -1,39 +1,93 @@
...
+      </>
+    )
+  }
+
+  return (
+    <form className="relative">
+      <textarea
+        onKeyDown={handleKeyDown}
+        onChange={(e) => {
+          let noSpace = e.target.value.replace(/[\s\u00A0]+/g, '').replace(/\r?\n|\r/g, '').slice(0, answer.length).toLowerCase()
```
afterburn, line 62: Let's try to keep the jsx markup separated from the logic, this helps with readability. So what I mean by that is simply create a handler function below `handleKeyDown`, maybe call it `handleChange` or something akin to that. Also makes things easier to find.

afterburn, line 62: I'm not sure what this code is doing:
```js
e.target.value.replace(/[\s\u00A0]+/g, '').replace(/\r?\n|\r/g, '').slice(0, answer.length).toLowerCase()
```

Its use seems to be to remove linebreaks from the users input, if that is the case I think a better approach is to use the `keypress` event and check what keyCode is being pressed. If the keyCode matches certain characters you wish to block you can `return false`. This way you dont have to modify the value after the fact.

afterburn:
Hey Ben thank you for all the hard work so far! I left some nitpicky comments but I hope they are helpful.

I also noticed that I'm able to input characters other than `a-z0-9` (see screenshot). Not sure if we want that for this specific lesson but @GBKS can confirm.
image

Perhaps out of scope for this PR but I think it makes sense to have the `<LessonPrompt>` component take in a `pattern` prop, which could be a regex (or a function so you we can create custom validators), and in the `<LessonPrompt>` component you can validate the input against that regex (or function) to make sure no invalid characters can be input. Something like this:
```jsx
<LessonPrompt
  className="max-w-[840px] px-4 py-8 md:items-center"
  label="Enter the OP_RETURN type"
  answer={answer}
  pattern={/^[a-z0-9]+$/i}
  onChange={setUserInput}
/>
```

ui/common/Input.tsx
benalleng: This was intended to handle formatting when pasting in text, it is intended to catch linebreaks and whitespace characters, i refactored it to make it more clear what does what

afterburn: Okay I see, let's use [the paste event](https://developer.mozilla.org/en-US/docs/Web/API/Element/paste_event) for that though. Something like this:

```js
function handlePaste(event) {
  event.preventDefault()

  const pasteData = (event.clipboardData || window.clipboardData).getData('text')
    .replace(/\s+/g, '')

  document.execCommand('insertText', false, pasteData)
}
```

```jsx
<textarea
  onPaste={handlePaste}
  ...
/>
```

benalleng:
great idea about the pattern matching on the UI!

ui/common/Input.tsx
benalleng: wow well that does it! thanks for showing me this solution!

benalleng: I tried implementing this paste event but it no longer set the textarea to the clipboard remaining blank after pasting

afterburn: Hey @benalleng, are you referring to this code or have you not pushed your attempt at this to GitHub?

```jsx
  const handlePaste = (event) => {
    const formattedText = event
      .replace(/[\s\u00A0]+/g, '')
      .replace(/\r?\n|\r/g, '')
      .slice(0, answer.length)
      .toLowerCase()
    return formattedText
  }
  
  const handleChange = (event) => {
    let formattedText = handlePaste(event.target.value)
    ...
  })
  ```
  
  ```jsx
  <textarea
    onChange={handleChange}
    ...
  />
  ```

**Comment by afterburn** [comment]
I gave this another test but I'm still able to enter characters other than `a-z0-9`. Let's make sure it doesn't accept any other characters than those and we can leave the pattern prop for another PR unless you want to resolve it in this one.

**Comment by benalleng** [comment]
@afterburn Sorry it wasn't ready for you to take a look at it from the call, I had reverted to a prior commit. This afternoon I took another pass and I got it working! The `<LessonPrompt />` component now accepts an optional pattern to allow the route template code to limit what is input. upon pasting the `onPaste` handler gets rid of anything that doesn't match the regex expression.  The `onChange` handler has also been simplified separating duties somewhat

File: ui/common/Input.tsx (hunk 12 above)
Comment (benalleng) [reply ui/common/Input.tsx]: I resolved this function with a slight adjustment to the onPaste handler

File: ui/common/Input.tsx [hunk 13]
```diff
@@ -1,39 +1,111 @@
...
+  const handlePaste = (event) => {
+    event.preventDefault()
+
+    const pasteData= (event.clipboardData || window.Clipboard).getData('text')
+    .toString()
+    .match(pattern)
+    .join('')
+    .slice(0, answer.length)
+
+  document.execCommand('insertText', false, pasteData)
```
afterburn, line 66: Some strange formatting here, have you ran eslint/prettier?

afterburn (changes requested):
Looks good, couldn't find any bugs. Left a small comments and you have some conflicts but after those have been resolved LGTM.

app/[lang]/chapters/[slug]/page.tsx
```diff
@@ -1,5 +1,5 @@
-import chapters from 'content/chapters'
-import introductions from 'content/introductions'
+import { chapters } from 'content'
+import { introductions } from 'content'
```
Comment (afterburn) [inline app/[lang]/chapters/[slug]/page.tsx:2]: You can combine these into a single import statement
```jsx
import { chapters, introductions } from 'content'
```

Comment (afterburn) [reply app/[lang]/chapters/[slug]/page.tsx:2]: Oh I actually think merge conflicts caused these, nevermind this comment.

benalleng:
There are a few new conflicts introduced by the new template layouts, I am going to convert this to a draft until it is fully merged with master.

benalleng:
@afterburn Should be good to go with the new challenge templates but lets have one final review just to check!

afterburn (approved):

GBKS:
Congrats @benalleng for getting this massive update merged, appreciate your work on this.

Sorry to report a bug right away, I'm finding that when I start entering text in the input field on Safari (desktop and mobile), it instantly switches from multi line to single line, stretching beyond the browser frame. This might just be a simple CSS setting (max-width or so).

PR #113, opened 2023-01-27
Update ADD_ISSUES_TO_PROJECT.yml

ecurrencyhodler:
Updated secret in the file specific to the repo to enable workflow.


satsie (approved):
ACK 0af818c04730f2e528059f7eb0f245eff405b44c - Checked that this is the correct secret variable name

PR #117, opened 2023-01-27
chore: automate code formatting

SamSamskies:
We haven't discussed this, but I think setting up a precommit hook to format files will help make sure the code is formatted consistently in the codebase. Feel free to let me know if any of you don't want this tooling in place for some reason.

This PR also makes sure every dev runs the `husky install` command by running it as a `postinstall` script. I noticed that setting up husky enforces conventional commits. I think that's because the `@commitlint/config-conventional` package is installed.

### DIFF
--- a/.commitlintrc.json
+++ b/.commitlintrc.json
@@ -1,3 +1,0 @@
-{
-  "extends": ["@commitlint/config-conventional"]
-}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,3 +1,3 @@
 {
-  "extends": "next/core-web-vitals"
+  "extends": ["next/core-web-vitals", "prettier"]
 }
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@ yarn-error.log*
 next-env.d.ts
 
 .vscode
+.contentlayer
+.idea
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,0 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-npx commitlint --edit $1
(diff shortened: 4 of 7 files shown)


GBKS:
I a not a pro on this, but sounds like a great idea. There were a few PRs that were tricky to read due to inconsistent tabbing.

One thing that I have a hard time with using Tailwind is that you end up with super long strings of classes that you mentally have break down, decode and order to make sense of ([example](https://github.com/saving-satoshi/saving-satoshi/blob/master/components/Hero.tsx#L6)). So I've started splitting up classes one per line ([example](https://github.com/saving-satoshi/saving-satoshi/blob/5cbe465b0d750691b64986c2c8fc692b851ed42d/components/partials/ChallengeItem.tsx#L33)), which is just much easier to work with. Is that something that would be standardized?

SamSamskies:
Prettier is very opinionated and that's by design. It let's you configure a small set of rules, but I don't think you can configure it to put classes on separate lines. I believe it puts the classes in a specific order though.

Prettier is the industry standard and has been for years though, so developers should be used to the formatting done by Prettier. Prettier tooling was already setup in this project. It just wasn't being enforced. It was just there if people wanted to use it. 

Husky was also already installed, but doesn't seem like anyone was actually using it.

GBKS:
Yeah, we never had a good group discussion about these practices, just the occasional comment exchange here and there. Something we should do to make our lives a little easier.

SamSamskies:
docs for conventional commits https://www.conventionalcommits.org/en/v1.0.0/

SamSamskies:
I'm planning on removing `@commitlint/config-conventional` to remove the requirement for conventional commits. If anyone is in favor of requiring conventional commits, we can discuss here.

afterburn (approved):

SamSamskies:
@ecurrencyhodler for some reason merging is still blocked even though @afterburn approved the PR.

SamSamskies:
@afterburn please test again when you get a chance. I added type checking to the pre-commit hook. That should make it fail if there's any type errors in the files being committed.

afterburn (approved):

ecurrencyhodler (approved):

Issue #114, opened 2023-01-27
Site crashes if you sign in and then click clear saved progress


SamSamskies:
**To Reproduce**

1. Open site in incognito window
2. Click on avatar icon on top right
3. Click `I've copied and backed up my code`
4. Click on avatar icon on top right again
5. Click `Clear saved progress`

**Screenshots and/or logs**

https://user-images.githubusercontent.com/3655410/215148827-2846c821-a101-48cf-b635-72f9732329cb.mp4

SamSamskies:
Cause seems to be a similar issue to https://github.com/saving-satoshi/saving-satoshi/issues/94.

Took a quick glance at the code and based on [the previous discussion ](https://github.com/saving-satoshi/saving-satoshi/pull/96), I think the `LoginModal` should only be rendered in `NavBar` if the user is registered. I also noticed a bunch of redundant code for keeping track of user state. Might be worth it to consolidate the user state management into a provider or hook.

GBKS:
Yes, looks like state management across the signup/login modals is broken. For step 5, you currently see the login state of the login modal, but you are already logged in and should see that state.

I think the problem is that the two modals do not refresh when they become visible. So they are dealing with an outdated state. We may need to introduce some type of global event bus to broadcast log-in/log-out/user-update states, or find a way to introduce a globally accessible reactive user object whose properties update components. I am not super well-versed with React and Next, maybe you have some good ideas on how to achieve this?

SamSamskies:
Maybe use something like `use-local-storage-state` hook https://github.com/astoilkov/use-local-storage-state. That looks like it broadcasts events whenever the values in localstorage are updated. It might not be necessary though. It might be possible to just use React Context or a simple custom `useUserReducer` hook to handle the user state management.

SamSamskies:
If not one else wants to tackle this, I can jump on it.

GBKS:
Please go ahead. Would be amazing if you can solve this.

SamSamskies:
@ecurrencyhodler you can assign this to me and move it to in progress.

PR #119, opened 2023-01-31
fix: user state management

SamSamskies:
resolves https://github.com/saving-satoshi/saving-satoshi/issues/114

I fixed the bug, but can whoever reviews this make sure the signup and login flows are correct? I'm not 100% sure if they are correct just by looking at the mockups.

I used `use-local-storage-state` to access values from localStorage in order to make sure components using localStorage values re-render whenever the values in localStorage changes.

~~A lot of the changes are just formatting that was done by Prettier. I prettified every file that I touched which is a strategy I've used on codebases that are not fully prettified. If anyone does not agree with this strategy and would prefer to review a smaller changeset without all the formatting changes, let me know and I can simplify the PR.~~

I also removed any unused variables in files I touched. I'd really recommend that we set up a linting rule to not allow unused variables.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,11 +13,7 @@
     "postinstall": "husky install"
   },
   "lint-staged": {
-    "**/*.{js,jsx}": "prettier --write",
-    "**/*.{ts,tsx}": [
-      "prettier --write",
-      "tsc-files --noEmit"
-    ]
+    "**/*.{js,jsx,ts,tsx}": "prettier --write"
   },
   "dependencies": {
     "@lionello/secp256k1-js": "^1.1.0",
@@ -32,7 +28,8 @@
     "react-modal": "^3.16.1",
     "react-responsive": "^9.0.2",
     "react-terminal-ui": "^1.0.0",
-    "react-transition-group": "^4.4.5"
+    "react-transition-group": "^4.4.5",
+    "use-local-storage-state": "^18.1.2"
   },
   "devDependencies": {
     "@svgr/webpack": "^6.5.1",
@@ -47,7 +44,6 @@
     "prettier": "^2.8.1",
     "prettier-plugin-tailwindcss": "^0.2.1",
     "tailwindcss": "^3.2.2",
-    "tsc-files": "^1.1.3",
     "typescript": "4.8.4"
   }
 }
--- a/components/Modals/Login.tsx
+++ b/components/Modals/Login.tsx
@@ -1,51 +1,32 @@
 'use client'
 
-import { useEffect, useState } from 'react'
+import { useState } from 'react'
 import Modal from 'react-modal'
 import Secp256k1 from '@lionello/secp256k1-js'
-
-import Avatar from 'components/Avatar'
-import { Button, CopyButton } from 'shared'
 import CloseIcon from 'public/assets/icons/close.svg'
 import WarningIcon from 'public/assets/icons/warning.svg'
+import Avatar from 'components/Avatar'
+import { Button, CopyButton } from 'shared'
+import { loginUser, logoutUser, clearUser } from 'lib/user'
+import { useUser } from 'hooks'
 
-import {
-  getUser,
-  loginUser,
-  isUserLoggedIn,
-  isUserRegistered,
-  clearUser,
-} from 'lib/user'
-
-export default function LoginModal(props) {
-  const [loaded, setLoaded] = useState(false)
+export default function LoginModal({ onClose, onLogin, open }) {
+  const [willShowProgressRestored, setWillShowProgressRestored] =
+    useState(false)
   const [userPrivateKey, setPrivateKey] = useState('')
-  const [user, setUser] = useState<any>({})
-  const [copied, setCopied] = useState(false)
-  const [isLoggedIn, setLoggedIn] = useState(false)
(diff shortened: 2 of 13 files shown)


components/Fonts.tsx
```diff
@@ -24,8 +24,6 @@ const nunito = Nunito({
 export default function Fonts() {
   return (
     <style
-      jsx={true}
-      global={true}
```
SamSamskies, line 28: There were warnings about these invalid attributes so I removed them. I don't noticed anything that breaks by removing them, but let me know if they are supposed to be there for some reason.

ecurrencyhodler:
Tested the PR. The site no longer crashes when I:

1. Open site in incognito window
2. Click on avatar icon on top right
3. Click I've copied and backed up my code
4. Click on avatar icon on top right again
5. Click Clear saved progress

SamSamskies:
This bug still exists on master, so this PR will still need to be reviewed once I rebase. Not sure if I'll get to it today. Probably tomorrow.

SamSamskies:
@ecurrencyhodler you should probably functionally test this again. The rebase was gnarly.

@afterburn or whoever else please review whenever you get a chance. Thanks.

afterburn:
> @afterburn we might want to rethink using default exports. I don't really care about it too much, but I tend to agree with the points made here https://basarat.gitbook.io/typescript/main-1/defaultisbad. On the last couple of large codebases I worked on my team switched to using a convention of never using default exports and it was nice.

Sure, I'm not strongly opinionated on this topic, so if you and the rest of the team feel this improves our codebase and DX then I think we should do it.

package.json
```diff
@@ -13,11 +13,7 @@
     "postinstall": "husky install"
   },
   "lint-staged": {
-    "**/*.{js,jsx}": "prettier --write",
-    "**/*.{ts,tsx}": [
-      "prettier --write",
-      "tsc-files --noEmit"
```
SamSamskies, line 19: I removed the TS compile check from the pre-commit hook. It wasn't working correctly. Spent a little time to try and fix it, but couldn't figure it out and decided to punt on it.

SamSamskies:
oops I broke the build. right after I took out the typecheck from the pre-commit hook 😂. at least the Vercel deployment check caught it.

SamSamskies:
k ready for review now

afterburn:
Maybe its just the deployment that's broken, but when I create a new user, then refresh the page, it gets me into a state I cannot get out of unless I clear my localStorage manually (see video). All seems to work just fine when I check out this branch locally though.

https://user-images.githubusercontent.com/10757768/216434405-ce6e6a9f-65a9-4bd6-88b8-d5816b783195.mp4

Can someone else help verify? Maybe @GBKS or @ecurrencyhodler? If it turns out that just the deployment is broken then LGTM.

afterburn (approved):

SamSamskies:
Wth. I wonder why does that only happen on the deployed at? 🤔

SamSamskies:
I was able to reproduce locally by running the app in prod mode instead of dev mode. Don't merge until I fix the bug.

SamSamskies:
Should be fixed now. Please test all user flows thoroughly before merging. 🙏

SamSamskies:
So prod mode statically builds some of the components which was causing the bug. So far not liking the Next.js app directory magic at all.

afterburn (approved):

ecurrencyhodler:
This is the flow of testing. I'm not sure this is accurate or expected:

1. I go to website and click on empty profile on top right.
2. I'm initially prompted to save my progress.
3. I copy and click "I've copied and backed up my code".
4. I exit the screen and click on the profile (which now has an avatar) on the top right again.
5. I'm prompted to save code or to log out now.
6. After I click "log out" I click on empty profile and am prompted to enter code.
7. I click "load my progress"
8. Then it says progress loaded and I'm able to continue.

If this is all expected user flow then this is a tACK.

ecurrencyhodler:
tACK. Okay I think to me the UX flows makes sense after going through prod and comparing the differences.

The language on the profile screen is confusing though at this state. There is nothing to load. But I'll either submit a PR or Issue to fix it. 
Screenshot 2023-02-02 at 9 13 06 PM


Sam are we g2g with merge?

GBKS:
Thanks for implementing this properly, much appreciated.

@ecurrencyhodler what's confusing about the text? Let's make sure copy change are first addressed in design before doing PRs.

SamSamskies:
@ecurrencyhodler ship it 🚀

ecurrencyhodler:
Okay will do Christoph.

ecurrencyhodler (approved):

ecurrencyhodler:
@SamSamskies looks like there's one conflict that needs to be resolved before getting merged.

SamSamskies:
@ecurrencyhodler rebased. I'd recommend functionally testing again before merging.

ecurrencyhodler:
I'll definitely test. But we should have another review before merge too since commits were made after the last approval.

SamSamskies:
I only made one tiny commit to fix the build after the rebase https://github.com/saving-satoshi/saving-satoshi/pull/119/commits/d3c3139befc636a69a66eb33c863bc74864d5920

ecurrencyhodler:
tACK

Issue #58, opened 2023-01-06
Challenge completion celebration


adamjonas:
As a learner, when I complete a challenge, I should be exhilarated and encouraged by completing a challenge.

Possibilities: 
1. Modal drop on completion with congratulations with some advancement of the narrative
2. An improved version of the completion banner
<img width="1315" alt="Screen Shot 2023-01-06 at 1 09 31 PM" src="https://user-images.githubusercontent.com/755825/211073083-bf82a4f6-4852-4c8c-a0cc-9c666b892ac1.png">

aureleoules:
Is this a duplicate of #53 ?

adamjonas:
@GBKS any input on this one?

GBKS:
Something we should tackle, for sure. Have not thought super deeply about it, but here are some ideas:

- The inputted text turns a celebratory green when it's correct (see [here](https://github.com/saving-satoshi/saving-satoshi/pull/97#issuecomment-1400007697))
- Change "Challenge completed" to be more informational and unique (see [here](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=1056%3A14226&t=c5pqnLqCDgVfvlS4-1))
- Some kind of elaborate animation that's fun and playful (Duolingo does this well, some info [here](https://blog.duolingo.com/streak-milestone-design-animation/)). I have not looked much into animation since there has been more basic stuff to take care of

ecurrencyhodler:
It would be great if we can explore animations. That would 10x the endorphines! If someone investigates how to integrate animations, design can help create something for eng to implement?

GBKS:
Most of the UI animations (really transitions) we need can be handled via the React/Next built-in transition functionality. For more elaborate animations, like drawing out Elliptic Curve Cryptography math, we may need to set up [Lottie](https://airbnb.io/lottie/).

GBKS:
I just got assigned this, so one extra comment. Before worrying about how to do this technically, let's first come up with animations. The tech solution can then follow our intent.

ecurrencyhodler:
Do you think this is worth including into v1?

GBKS:
Depends on how we scope it out. When this issue was posted on Jan 6, the experience looked very different. Maybe this is already addressed?

benalleng:
I tried my hand in a little animation practice with #311. It doesn't work quite like i want it to however but still something to consider as inspiration for this

[...]

PR #124, opened 2023-02-01
Challenge Completion enhancement

benalleng:
This is a further enhancement of #58 .

green is good!
image


ecurrencyhodler:
Tested. This feels like a step in the right direction. However I also wonder what it would like to have the challenge completion (celebration) animated as well.
Screenshot 2023-02-01 at 9 22 58 PM

afterburn (approved):
image

I think we can get this merged and do the animation in a separate pr when we make a decision on how to proceed.

ecurrencyhodler:
LOL that meme is amazing. Okay looks like this PR is still in draft mode.

ui/common/StatusBar.tsx
```diff
@@ -62,7 +62,7 @@ export default function StatusBar({
       className={clsx('border-t border-white/25 max-md:px-4 max-md:py-8', {
         'w-screen': full,
         'w-full': !full,
-        'bg-green-400/25': getStatus() === Status.Success,
+        'bg-[#28b12326]': getStatus() === Status.Success,
```
afterburn, line 65: Are we planning to use this color for other things apart from this element? If so, I think its better to move this to our `tailwind.config.js`. If it's a one off, ignore my comment.

benalleng: I decided to narrow the focus of this PR to just handle the success status bar color, We can continue to consider the custom tailwind config but for now lets just make sure we see green

ecurrencyhodler:
Tested again. Looks good and ran into no bugs for me.

afterburn:
Same here, lgtm

PR #131, opened 2023-02-03
Locking chapter info

### LINKED ISSUE #10: Show dynamic content in tabs on the chapters page
Currently, on the chapter overview we have 2 tabs, Info and Challenges

Screenshot 2022-12-01 at 11 55 44 PM

Those tabs need to load the content dynamically, as shown in the figma, an overview in text, and the list of challenges 

Overview component already exists to load the chapter overview, you can find it in  `components/partials/Overview.tsx`

Overviewinfo, is the class that contains the tab groups (I was looking to link to other views but i think changing the Link inside Tab.tsx to a button that toggles the info would be possible), this class receives a Chapter type in its parameters, a chapter contains its lessons, we could load those in the "challenges" section

Screenshot 2022-12-02 at 2 12 56 AM

vinayaksh42:
Fixes #10 

Utilizing `getUserProgress` for fetching the last completed chapter by user. Added a lock icon and text under the description of chapters which are yet to be unlocked.

**ScreenShots**
Screenshot 2023-02-03 at 2 28 25 PM
Screenshot 2023-02-03 at 2 28 07 PM


ecurrencyhodler:
tACK! Confirmed chapters are locked both on the chapter overview page and when user is in the process of completing Ch. 1. Great work!

afterburn:
Hey @vinayaksh42, thank you for picking this up.

I tested this locally and there is a hydration error on the chapters page. I believe the reason for this error is the `display ?` block you're using in `Chapter.tsx`. I noticed your new `getUserChapterStatus` function is using `getUserProgress` from our `user.ts` library. It's good to remember this function should only be executed on the client-side because if you run it on both client and server-side the results will be different. This means different HTML will be rendered on client- vs. server-side, resulting in the hydration error.

I think a better approach would be to modify the new `useStatus` hook so it can also return `unlocked` status when you only pass it a `chapterId`. The hook uses `useEffect` which ensures the code is only executed on the client-side. You can see how we are using it in `components/Navbar/Tab.tsx`.

vinayaksh42:
> LGTM after conflicts have been resolved.

All merge conflicts resolved

afterburn (approved):

PR #134, opened 2023-02-06
Status bar positioning for mobile

### LINKED ISSUE #128: StatusBar not at the bottom of the page in mobile view
**Describe the bug**
According to the [Figma mockup](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=379%3A10942&t=AbXTb7lnzN43hvk4-0) the `StatusBar` is supposed to be at the bottom of the page for mobile view

**Screenshots**

**To Reproduce**

1. Go to https://savingsatoshi.com/chapters/chapter-1/transacting-1
2. Either open in Mobile or Change to Mobile view using device toolbar
3. See error

benalleng:
Should close #128. Further the correctness of mobile styling for chapter 1.

Make sure to view the preview on mobile!
[Mobile statusBar Preview](https://saving-satoshi-git-fork-benalleng-statusb-40e2ac-saving-satoshi.vercel.app/en/chapters/chapter-1/genesis-1)


benalleng:
Not necessarily blocked by #97 but the keyboard pushing up the content certainly causes issue with this size of input area. in the [figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=377%3A5956&t=wFaOjb2rb1TU80Rj-0) it is shown as moving the status-bar up with the virtual keyboard but worth thinking about the viability of this approach with the space allotted

ecurrencyhodler:
tACK! Thanks for pushing the status bar down to the bottom of mobile view. 

The fix is on the left screen and the current prod version is on the right. This PR still needs a code review.
tack

ecurrencyhodler:
> 

Hey @GBKS Ben found a good catch. The figma (which I believe is the source of truth) shows a keyboard when doing the coding copy/paste challenge. What should be the state we move towards? With or without the keyboard? 

Context: Ben just pushed the status bar to the bottom of the mobile view in this PR. The keyboard could block the status view. Below on the left is the screen with the status bar at the bottom. The right is what we see in figma. 

Screenshot 2023-02-06 at 9 47 37 AM

ui/common/StatusBar.tsx
```diff
@@ -59,7 +59,7 @@ export default function StatusBar({
 
   return (
     <div
-      className={clsx('border-t border-white/25 max-md:px-4 max-md:py-8', {
+      className={clsx('border-t border-white/25 max-md:px-4 max-md:py-8 max-md:bottom-0 max-md:fixed', {
```
benalleng, line 62: for reference the fixed position will keep the statusbar at the bottom so when they keyboard is opened it will dissappear in this PR currently

GBKS:
Yeah, so this is a super common thing for mobile UI. A keyboard will always show up when focusing an input field, that's just how it works. And it always takes 25-50% of the vertical space based on your screen size.

The layout logic needs to account for this. One approach would be to avoid fixed positioning and use flex grow to make the whole content area the full screen height. Then apply [grow](https://css-tricks.com/snippets/css/a-guide-to-flexbox/#aa-flex-grow) to the code input area, so it uses whatever height is available and pushes the bottom bar to the bottom of the screen. With overflow scroll applied users can then scroll within that area. Let me know if you'd like me to mock this up visually.

benalleng:
> Yeah, so this is a super common thing for mobile UI. A keyboard will always show up when focusing an input field, that's just how it works. And it always takes 25-50% of the vertical space based on your screen size.

Just to get some more clarification I am wondering if we want to make sure we can always see the statusBar even when the keyboard is shown like is currently in the figma. Even with the adjustment to flex grow it doesn't quite come within view (note #97 is still not merged and will change the size of this again)

Screenshot_20230214-152444

Screenshot_20230214-152457

afterburn:
Hey @benalleng #97 has been merged, do you need to make changes to this PR or is it ready for review?

benalleng:
@afterburn It should be ready for a review! Before merging I still would like a review from @GBKS to make sure this is going in the direction of what is wanted to best align with the figma

GBKS:
@benalleng ACK, this looks good. There is no 100% perfect solution to this due to how dynamic the viewport height is on mobile Safari with the keyboard sliding in and out, and also the top & bottom chrome getting smaller and larger as you scroll. 

Maybe a useful tweak would be to interpret an "Enter" key press as an "I am done with this input" and remove input focus (and therefore causing the keyboard to disappear)?

benalleng:
> Maybe a useful tweak would be to interpret an "Enter" key press as an "I am done with this input" and remove input focus (and therefore causing the keyboard to disappear)?

As of #97 I have the input remain in focus until the user gets the correct answer at which point the keyboard disappears bringing the statusBar back into view

afterburn (approved):

Issue #121, opened 2023-02-01
Add translations


afterburn:
Right now Saving Satoshi only supports English but by adding translations we can increase accessibility and improve the user experience.

The idea behind the new architecture (#118) is to define the translations in the same file where they will be utilized. I'm a bit on the fence about whether that is a good idea or if maybe we should have a single file containing all translations (or group them based on locale) to prevent collisions between translation keys. 

Anyway, the solution I propose for now is to create a `useTranslations` hook that we can use from anywhere within our codebase. It returns a function `t` you can use like this:

```jsx
export default function Genesis() {
  const t = useTranslations()
  
  return <>
    <span>{t('genesis.paragraph_one')}</span>
  </>
}

export const translations = {
  en: {
    genesis: {
      paragraph_one: `Let's find the very first block in the Bitcoin blockchain...`
    }
  }
}
```

The `useTranslations` hook should take care of the getting the right `locale` from the users' browser settings, or we could create a language selector in the Topbar. It then finds the correct translation key-value pair using the key that was passed to the function and returns the corresponding value.

Like I mentioned during our call on `January 31, 2023` I think it might prove to be difficult to combine this with the Tooltip functionality that we have planned. It would be ideal if we could define our translations using `jsx`, but I'm not sure that's the way to go if we want to outsource the translation work at some point in the future.

So what I propose instead is that we create some logic that will find and replace parts of the strings that we mark as being a tooltip. My first thought was to use markdown syntax:
```jsx
export const translations = {
  en: {
    genesis: {
      paragraph_one: `Let's find the very first [block](genesis.tooltip) in the Bitcoin blockchain...`,
      tooltip: `A block refers to ...`
    }
  }
}

```
But since the text we target could be a clickable link at the same time this doesn't work very well since we cannot pass a href (unless we come up with our own syntax). Perhaps instead we can do something like this:
```jsx
export const translations = {
  en: {
    genesis: {
      paragraph_one: `Let's find the very first <Tooltip content="genesis.tooltip" href="...">block</Tooltip> in the Bitcoin blockchain...`,
      tooltip: `A block refers to ...`
    }
  }
}
```
This approach feels like you're writing jsx, which is nice. While at the same time it still allows us to store our translations as strings, which is good for third party translation tools. On the flip side this might be a little verbose. I'm open to alternative solutions that allow for both translations and tooltips to interface nicely.

It might be a good idea to establish some requirements first though, so let's use this issue to discuss.

[...]

PR #138, opened 2023-02-07
React intl translations spike

SamSamskies:
Did a spike to demonstrate how react-intl could be used for translations. See discussion on issue https://github.com/saving-satoshi/saving-satoshi/issues/121.

You can view the partially translated genesis lesson [here](http://localhost:3000/chapters/chapter-1/genesis). I only translated the phrase `block explorer` as an example. You'll need to update your browser language settings to Spanish in order to see the translated text. Another way we could go is to have the locale in the URL. For example, https://savingsatoshi.com/es would set the locale to `es`. I think Next.js has a way to auto detect locale and add it to the URL if the locale is missing from the URL.

Translators would only ever have to look at one specific file. For example, for Spanish translators, they check the `es.json` file and if any translations are missing they can easily tell because the values are blank strings.

### DIFF
--- a/app/chapters/[slug]/[lesson]/page.tsx
+++ b/app/chapters/[slug]/[lesson]/page.tsx
@@ -1,6 +1,9 @@
+'use client'
+
 import chapters from 'content/chapters'
 import lessons from 'content/lessons'
 import { Button } from 'shared'
+import { IntlProvider } from 'components'
 
 export default function Page({ params }) {
   if (!(params.lesson in lessons) || !(params.slug in chapters)) {
@@ -18,5 +21,9 @@ export default function Page({ params }) {
 
   const Lesson = lessons[params.lesson].default
 
-  return <Lesson />
+  return (
+    <IntlProvider>
+      <Lesson />
+    </IntlProvider>
+  )
 }
--- a/components/IntlProvider.tsx
+++ b/components/IntlProvider.tsx
@@ -2,15 +2,10 @@
 
 import { IntlProvider as ReactIntlProvider } from 'react-intl'
 import { useHasMounted } from 'hooks'
+import enMessages from 'translations/en.json'
+import esMessages from 'translations/es.json'
 
 const DEFAULT_LOCALE = 'en-US'
-const enMessages = {
-  hello: 'hello',
-}
-
-const esMessages = {
-  hello: 'hola',
-}
 
 const getMessages = (locale) => {
   const truncatedLocale = locale.toLowerCase().split(/[_-]+/)[0]
--- a/components/IntlProvider.tsx
+++ b/components/IntlProvider.tsx
@@ -1,0 +1,48 @@
+'use client'
+
+import { IntlProvider as ReactIntlProvider } from 'react-intl'
+import { useHasMounted } from 'hooks'
+
+const DEFAULT_LOCALE = 'en-US'
+const enMessages = {
+  hello: 'hello',
+}
+
+const esMessages = {
+  hello: 'hola',
+}
+
+const getMessages = (locale) => {
+  const truncatedLocale = locale.toLowerCase().split(/[_-]+/)[0]
+
+  switch (truncatedLocale) {
+    case 'en':
+      return enMessages
+    case 'es':
+      return esMessages
+    default:
+      return enMessages
+  }
+}
+
+export const IntlProvider = ({ children }) => {
+  const hasMounted = useHasMounted()
+  const getLocale = () => {
+    if (typeof navigator === 'undefined' || !hasMounted) {
+      return DEFAULT_LOCALE
+    }
+
+    return (
+      (navigator.languages && navigator.languages[0]) ||
+      navigator.language ||
+      DEFAULT_LOCALE
+    )
+  }
+  const locale = getLocale()
+
+  return (
+    <ReactIntlProvider locale={locale} messages={getMessages(locale)}>
+      {children}
+    </ReactIntlProvider>
+  )
+}
--- a/content/lessons/genesis/genesis.tsx
+++ b/content/lessons/genesis/genesis.tsx
@@ -12,6 +12,8 @@ import {
   Tooltip,
   LessonPrompt,
 } from 'ui'
+import { FormattedMessage, useIntl } from 'react-intl'
+import { formattedMessages } from './messages'
 
 const inputAmount = 154
 const answer =
@@ -30,6 +32,7 @@ const tabData = [
 
 export default function Genesis1() {
   const [userInput, setUserInput] = useState('')
+  const { formatMessage } = useIntl()
 
   return (
     <Lesson>
@@ -39,15 +42,21 @@ export default function Genesis1() {
         <Title>Find block 0</Title>
 
         <Text className="text-lg">
-          Let’s find the very first block in the Bitcoin blockchain. Go to a
-          <Tooltip
-            content="We recommend https://blockstream.info/"
-            href="https://blockstream.info/"
-          >
-            block explorer
-          </Tooltip>
-          and search for ’0’. This will take you to the block at height 0. As
-          you can see, this block only contains one transaction.
+          <FormattedMessage
+            {...formattedMessages.firstParagraph}
+            values={{
+              blockExplorerTooltip: (
+                <Tooltip
+                  content={formatMessage(
+                    formattedMessages.blockExplorerTooltipContent
+                  )}
+                  href="https://blockstream.info/"
+                >
+                  <FormattedMessage {...formattedMessages.blockExplorer} />
+                </Tooltip>
+              ),
+            }}
+          />
         </Text>
 
         <Text className="text-lg">
--- a/content/lessons/genesis/messages.js
+++ b/content/lessons/genesis/messages.js
@@ -1,0 +1,16 @@
+import { defineMessages } from 'react-intl'
+import { formatMessages } from 'lib/messages'
+
+const messages = defineMessages({
+  blockExplorer: 'block explorer',
+  blockExplorerTooltipContent: 'We recommend https://blockstream.info/',
+  firstParagraph:
+    'Let’s find the very first block in the Bitcoin blockchain. Go to a {blockExplorerTooltip} and search for ’0’. This will take you to the block at height 0 you can see, this block only contains one transaction.',
+})
+
+export default messages
+
+export const formattedMessages = formatMessages(
+  messages,
+  'content.lessons.genesis'
+)
--- a/lib/messages.ts
+++ b/lib/messages.ts
@@ -1,0 +1,17 @@
+import { MessageDescriptor } from 'react-intl'
+
+export const formatMessages = (
+  messages: Record<string, string>,
+  messageIdPrefix: string
+) => {
+  const result: Record<string, MessageDescriptor> = {}
+
+  for (const key in messages) {
+    result[key] = {
+      id: `${messageIdPrefix}.${key}`,
+      defaultMessage: messages[key],
+    }
+  }
+
+  return result
+}
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "react-individual-character-input-boxes": "^1.2.1",
+    "react-intl": "^6.2.7",
     "react-jsx-parser": "^1.29.0",
     "react-modal": "^3.16.1",
     "react-responsive": "^9.0.2",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "prettier": "prettier --write --ignore-unknown .",
     "prettier:check": "prettier --check --ignore-unknown .",
     "test": "yarn prettier:check && yarn lint",
-    "postinstall": "husky install"
+    "postinstall": "husky install",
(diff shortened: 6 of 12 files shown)


package.json
```diff
@@ -10,7 +10,8 @@
     "prettier": "prettier --write --ignore-unknown .",
     "prettier:check": "prettier --check --ignore-unknown .",
     "test": "yarn prettier:check && yarn lint",
-    "postinstall": "husky install"
+    "postinstall": "husky install",
+    "generate-translations": "extract-messages -l=en,es -o translations -d en --flat true **/messages.js"
```
SamSamskies, line 14: This command updates the translations files and should be run whenever a `messages.js` file it created, updated, or deleted.

content/lessons/genesis/messages.js
```diff
@@ -0,0 +1,16 @@
...
+  blockExplorerTooltipContent: 'We recommend https://blockstream.info/',
+  firstParagraph:
+    'Let’s find the very first block in the Bitcoin blockchain. Go to a {blockExplorerTooltip} and search for ’0’. This will take you to the block at height 0 you can see, this block only contains one transaction.',
+})
+
+export default messages
+
+export const formattedMessages = formatMessages(
+  messages,
+  'content.lessons.genesis'
```
SamSamskies, line 15: This value needs to match the path to the folder of the messages.js file.

content/lessons/genesis/messages.js
```diff
@@ -0,0 +1,16 @@
...
+const messages = defineMessages({
+  blockExplorer: 'block explorer',
+  blockExplorerTooltipContent: 'We recommend https://blockstream.info/',
+  firstParagraph:
+    'Let’s find the very first block in the Bitcoin blockchain. Go to a {blockExplorerTooltip} and search for ’0’. This will take you to the block at height 0 you can see, this block only contains one transaction.',
+})
+
+export default messages
+
+export const formattedMessages = formatMessages(
```
SamSamskies, line 13: Every messages.js file should follow a pattern like this with a default export from the `defineMessages` results and a an export called `formattedMessages` which is the result from `formatMessages`.

I made the file a JS file cause I couldn't figure out how to make the TS gods happy. These files are pretty basic, so I don't think it's a big deal to leave them as JS.

SamSamskies:
@afterburn let me know what you think. I know you already wrote a bunch of code for translations, but just wanted to demo an example of what it would look like to use `react-intl`.

app/chapters/[slug]/[lesson]/page.tsx
```diff
@@ -18,5 +21,9 @@ export default function Page({ params }) {
 
   const Lesson = lessons[params.lesson].default
 
-  return <Lesson />
+  return (
+    <IntlProvider>
```
Comment (SamSamskies) [inline app/chapters/[slug]/[lesson]/page.tsx:25]: The provider probably should just go in the top level layout file, but I couldn't figure out how to get that to work with Next.js app directory.

translations/es.json
```diff
@@ -0,0 +1,5 @@
+{
+  "content.lessons.genesis.blockExplorer": "explorador de bloques",
+  "content.lessons.genesis.blockExplorerTooltipContent": "",
+  "content.lessons.genesis.firstParagraph": ""
```
SamSamskies, line 4: Translators can easily scan the file for blank strings to see where translations are needed.

SamSamskies: Tools like [Crowdin](https://crowdin.com/) could also be used with this setup.

SamSamskies:
react-intl also have a bunch of other components that can localize numbers, dates, times, etc. https://formatjs.io/docs/react-intl/#the-react-intl-module

content/lessons/genesis/messages.js
```diff
@@ -0,0 +1,11 @@
...
+import { formatMessages } from 'lib/messages'
+
+const definedMessages = defineMessages({
+  blockExplorer: 'block explorer',
+  blockExplorerTooltipContent: 'We recommend https://blockstream.info/',
+  firstParagraph:
+    'Let’s find the very first block in the Bitcoin blockchain. Go to a {blockExplorerTooltip} and search for ’0’. This will take you to the block at height 0 you can see, this block only contains one transaction.',
+})
+
+export default formatMessages(definedMessages, 'content.lessons.genesis')
```
SamSamskies, line 11: The 2nd param  needs to match the path to the folder of the messages.js file.

SamSamskies:
The [Next.js i18n routing](https://youtu.be/SJDYCw5KC7s) actually looks really nice and simple to setup. I can try adding that in tomorrow.

content/lessons/genesis/messages.js
afterburn: I feel there is a little too much boilerplate in this file. It would be great if we could have it be JSON instead and handle the JS elsewhere. This would make it easier for contributors who are not familiar with `react-intl` to add translations and gets rid of the .js/.ts issue. I am not familiar with `react-intl` though, so please let me know if this is possible.

components/IntlProvider.tsx
```diff
@@ -0,0 +1,43 @@
+'use client'
```
afterburn, line 1: Let's create a new folder called `providers` either in the root folder or under `components/providers` so we can more easily differentiate between `ui` and `functionality`.

lib/messages.ts
```diff
@@ -0,0 +1,17 @@
+import { MessageDescriptor } from 'react-intl'
```
afterburn, line 1: I think it makes sense to rename `lib/messages.ts` to `lib/translations.ts`.

content/lessons/genesis/messages.js
```diff
@@ -0,0 +1,11 @@
+import { defineMessages } from 'react-intl'
```
afterburn, line 1: I think it makes sense to rename this file from `messages` to `translations`.

afterburn:
Hey @SamSamskies, thank you for creating this spike. While I would love to be able to keep our content and translations inside a single file, I think for translations it makes sense to separate them to make sure they can interface well with third-party solutions (like Crowdin), so its great that you did.

I left some comments but I also have a question about the babel stuff. I'm not sure why we need to add `babel-plugin-react-intl-auto` but from a quick glance it will prevent us from having to assign unique identifiers to our translations, while at the same time it prevents collisions between translation pairs. Is there a different solution that but doesn't require babel?

content/lessons/genesis/messages.js
SamSamskies: Usually you don't need this extra util. It should be possible to do things with existing message extraction tooling, but I couldn't figure out how to get any of them to work with Next.js. Not sure if it's worth the time to continue trying to figure out how to get it to work. I spent hours trying to figure it out yesterday and ended up just writing this hack instead. Another option is to ditch existing message extraction tooling and write our own message extraction script, but I'm not sure if that's worth our time either. Maybe you or someone else want to try and take a quick stab at trying to figure out how we could remove this hack.

If we keep my hack, we could split the message definitions and formatting of the messages up into different files, but I actually think it's better to couple them like this. It's more convenient and the contributors writing this code will most likely be the same people writing the components that consume this data, so they'd have to become familiar with react-intl. Translators shouldn't ever be touching these files. They will only work with the JSON files that get created by the message extraction tooling or they would work with the platform that we upload the JSON files to. 

I'd suggest trying out this pattern and if it doesn't feel right after creating a few messages.js files, then we pivot.

content/lessons/genesis/messages.js
SamSamskies: All the react-intl docs, tooling, and examples all refer to them as `messages` though. Are we sure we want to use `translations` rather than `messages`?

SamSamskies:
> While I would love to be able to keep our content and translations inside a single file, I think for translations it makes sense to separate them to make sure they can interface well with third-party solutions (like Crowdin), so it's great that you did that already.

Keeping everything in one file with [a solution like yours ](https://github.com/saving-satoshi/saving-satoshi/pull/122) may be good enough for us. Would be good to hear other peoples opinions. Not sure if we need to use a platform such as Crowdin. That will just be another added expense and not sure if we need it.

The big benefit of using something like react-intl is all the existing tooling, but again not sure if we need it.

> I'm not sure why we need to add babel-plugin-react-intl-auto but from a quick glance it prevents us from having to assign unique identifiers to our translations, while at the same time it prevents collisions between translation pairs. Is there a different solution that does that, but doesn't require babel?

The babel stuff is required for the message extraction tooling I used that puts all the messages into the JSON files. I haven't found any other tooling that can extract the messages into one JSON file per locale. We could probably write our own script to do it without using using babel. Not sure if it's worth it to spend time doing that though.

content/lessons/genesis/messages.js
SamSamskies: Also, these files do not include any translations. The translations are what ends up in the locale JSON files.

afterburn: Yes, I think for people who are not familiar with `react-intl` the word `translations` makes more sense than `messages` but this might just be my personal preference. I feel it communicates the intent more clearly. If users need to visit react-intl docs I think our translation system is too complicated anyway, but curious to learn what others think about this.

content/lessons/genesis/messages.js
afterburn: > the contributors writing this code will most likely be the same people writing the components that consume this data, so they'd have to become familiar with react-intl

This is kind of what I'm trying to prevent though. I feel contributors shouldn't have to learn how `react-intl` works just to add some text to their challenge. I'm fine with keeping this as-is for now, but I do think we need to simplify this a little further later on. Ideally contributors would just create a `messages.json/translations.json` file, add some strings and would then be able to use them in their components by referencing the key, like in my solution.

content/lessons/genesis/messages.js
SamSamskies: react-intl might be the wrong choice then. One of the main benefits of using react-intl is to use all the components they provide. You'll have to read the docs in order to take advantage of all the benefits their components provide.

content/lessons/genesis/genesis.tsx
```diff
@@ -39,15 +42,19 @@ export default function Genesis1() {
...
-          Let’s find the very first block in the Bitcoin blockchain. Go to a
-          <Tooltip
-            content="We recommend https://blockstream.info/"
-            href="https://blockstream.info/"
-          >
-            block explorer
-          </Tooltip>
-          and search for ’0’. This will take you to the block at height 0. As
-          you can see, this block only contains one transaction.
+          <FormattedMessage
```
afterburn, line 45: Now that I'm revisiting this PR I think there is a lot of complexity here just to render out some strings. I touched upon this in another comment but it would be great if we can abstract away the `react-intl` specific stuff somehow so people don't have to get familiar with the library.

If we could get this to more closely resemble what I have in [my solution](https://github.com/saving-satoshi/saving-satoshi/pull/122/files#diff-adb4d3d69d8ccad9528f9d28420f1e3a7bc4558ae1baf51b980924a6e707c19eR42), that'd be great.

SamSamskies: Might be worth spiking on something like https://www.i18next.com/ instead of adding abstractions to react-intl. IMO the main benefit of using something like react-intl is using all the components they provide for rendering the strings. i18next more resembles what you wrote and I think it also provides easy use of rendering children like the Tooltip components.

SamSamskies: Or we could just use try out what you made. Might be good enough.

SamSamskies: But also keep in mind that most of the time the translation rendering code doesn't look like this. This is a more complex example of rendering a nested component that also has a translated prop and children.

SamSamskies: Another benefit of i18next is you can easily plug into their platform http://locize.com.

afterburn:
Okay, based on what you explained in your other comments I think we might want to give `i18next` a shot then, assuming it has support for embedding components into the text (like our tooltip). Is this something you want to create a spike for @SamSamskies?

SamSamskies:
> Okay, based on what you explained in your other comments I think we might want to give `i18next` a shot then, assuming it has support for embedding components into the text (like our tooltip). Is this something you want to create a spike for @SamSamskies?

I took a quick look at a [i18next Next.js example](https://locize.com/blog/next-13-app-dir-i18n/#4-Language-switcher) and it actually is also not really that straight forward for nested components. People would still need to look at the docs in order to write this code. I actually find their syntax harder to grok than react-intl. I think that's just something we'd have to do deal with if we want to easily render translated content that has a nested component.

Also, setting this up for Next.js app directory looks pretty crazy. At least they have docs though. I couldn't find any examples of using react-intl with Next.js app directory.

SamSamskies:
> The [Next.js i18n routing](https://youtu.be/SJDYCw5KC7s) actually looks really nice and simple to setup. I can try adding that in tomorrow.

Next.js 13 app directory doesn't support i18n routing 🙁

afterburn:
Okay, thanks for investigating @SamSamskies. In that case let's use my solution for now. If we run into issues with it down the line, we can always refactor it to either use `react-intl` or some other solution. I think we should prioritize developer experience at this stage over potential integrations that may not be necessary in the future.

I'll resolve merge conflicts on my branch and move the translations to their own file, then I'll request a review from you.

PR #136, opened 2023-02-06
CodeExample optional copy

benalleng:
Enhances the `CodeExample.tsx` component to allow for optional copy/paste as well as wraps the code better for mobile. without the copy button present this could help the user interactions on code examples where there has been confusion on what "should" be directly copied and what should be adjusted to fit the code challenge. Will generally help with UX

## Pull Request checklist

Please review the below PR requirements:

- [x] Build was run locally and any changes were pushed
- [x] Ensure that the title clearly describes the changes
- [x] Issue to be referenced in the PR description rather than in the commits or PR title
- [x] Clean commit history
TODO: decide what each code example should look like in their respective routes. 
As an example `echo ScriptSigHex | xxd -r -p` probably should not have the copy button enabled as the user should not copy this code verbatim

### Terminal Challenge 2 (no copy)
image

### Transaction Challenge 1 (copy)
image

PR #143, opened 2023-02-08
Terminal Challenge Variables

benalleng:
This builds on #136 and #75 to more closely match the expectations of @GBKS's user journey in this [youtube video](https://www.youtube.com/watch?v=94R-XbNRrS8) and increasing difficulty with the progression in the lessons. This adds both the ability for custom terminal variables as well as custom error codes and in-line tips for the user to see more clearly.

Custom user variables
image
[User Variable Preview](https://saving-satoshi-git-fork-benalleng-termina-aaed06-saving-satoshi.vercel.app/chapters/chapter-1/genesis-2)

Custom Error messages
image
[Custom Error Preview](https://saving-satoshi-git-fork-benalleng-termina-aaed06-saving-satoshi.vercel.app/chapters/chapter-1/transacting-2)


GBKS:
This is great. The interaction becomes much more rewarding when we have a few unique contextual responses sprinkled in.

The code is getting quite complex, so I will not comment on the implementation. Others are better suited for that.

GBKS:
Also thank you for including the preview links.

ui/common/CodeExample.tsx
```diff
@@ -1,14 +1,23 @@
 import { CopyButton } from 'ui'
 
-export default function CodeExample({ code, language }) {
+export default function CodeExample({ code, language, copy }: {code: string; language: string; copy?: boolean | string}) {
```
vinayaksh42, line 3: Do we require the prop - `copy` to have string value?
I think in most cases the copy button will be required to copy the content within the `CodeExample` box.

ui/common/CodeExample.tsx
```diff
@@ -1,14 +1,23 @@
...
+  const copyElement = () => {
+    if(copy === true) {
+      return <CopyButton compact content={code} />
+    } else if(typeof(copy) === 'string') {
```
vinayaksh42, line 7: This will also assist us in simplifying the code within this function, by limiting the copyable content to only that within the code prop.

ui/lesson/TerminalChallenge.tsx
```diff
@@ -68,6 +90,29 @@ export default function TerminalChallenge({
         }, 1000)
         setAnswer(answerValue)
       }
+    } else if ((input === `echo ${expectedInput.userVariable} | xxd -r -p`) || (input === `echo ${expectedInput.value} | xxd -r -p`)) {
+      const varInput = `echo ${expectedInput.value} | xxd -r -p`
+      const givenInput = varInput.split(' ')[1]
+      const answerValue = Buffer.from(givenInput, 'hex').toString('utf8')
+      setLines((lines) => [...lines, { value: answerValue, type: 'output' }])
+
+      if (givenInput === expectedInput.value) {
```
vinayaksh42, line 99: You can implement something like this:
```js
const onChange = (input) => {
  setLines((lines) => [...lines, { value: input, type: 'input' }])

  const inputs = [
    `echo ${expectedInput} | xxd -r -p`,
    `echo ${expectedInput.userVariable} | xxd -r -p`,
    `echo ${expectedInput.value} | xxd -r -p`,
  ]

  if (inputs.includes(input)) {
    const varInput = input === `echo ${expectedInput} | xxd -r -p`
      ? `echo ${expectedInput} | xxd -r -p`
      : `echo ${expectedInput.value} | xxd -r -p`

    const givenInput = varInput.split(' ')[1]
    const answerValue = Buffer.from(givenInput, 'hex').toString('utf8')
    setLines((lines) => [...lines, { value: answerValue, type: 'output' }])

    if (givenInput === expectedInput.value) {
      setTimeout(() => {
        saveProgress()
        setSuccess(true)
      }, 1000)
      setAnswer(answerValue)
    }
  } else if (commonError && input.includes(commonError.error)) {
    setLines((lines) => [...lines, { value: commonError.message, type: 'output' }])
  } else {
    setLines((lines) => [...lines, { value: 'Hmm... Sorry that’s not quite right.', type: 'output' }])
  }
}
```
It can help in making the code a bit cleaner by combining the common logic into a single block.

vinayaksh42 (commented):
I have left a few very nitpicky suggestions that can help in making the code a bit cleaner.

ui/common/CodeExample.tsx
benalleng: Good call, I had added this just to check if this would be a good solution instead of adding the additional logic for `expectedInput.userVariable` and `.value` but with this being added this feature isn't really necessary

ui/lesson/TerminalChallenge.tsx
benalleng: I refactored the code to contain the logic all within a single block but i did add the functionality allowing the user to run the function on 'any' hex value they like to have some fun with what they've learned

ui/lesson/Terminal.tsx
```diff
@@ -39,40 +30,20 @@ export default function Terminal({
...
                 className={`${
-                  line.type === 'input'
-                    ? 'react-terminal-previous-input text-white/80'
-                    : 'react-terminal-output text-[var(--terminal-output)]'
+                  (line.type === 'answer' &&
+                    'react-terminal-answer rounded bg-green/25 text-green') ||
+                  (line.type === 'output' &&
+                    'react-terminal-output text-[var(--terminal-output)]') ||
+                  (line.type === 'input' &&
+                    'react-terminal-previous-input text-white/80')
```
benalleng, line 38: I recognize this code may be a bit convoluted for what it does, and i think it could be simplified somewhat but I hit a blocker on just trying to get it to work without using these operands.

vinayaksh42:
great work @benalleng, I have noticed 2 issues:
1. Upon submitting the code the previous terminal lines are erased (not sure if this is an intentional change as the figma mockup suggests having the previous lines as well upon submitting the code - https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=1604%3A21834&t=Hq8czLd2MYK4HZFW-0)
2. When pressing the left arrow key the `insertion pointer` of the terminal remains at the end of the line. (This bug is present on the live website as well, I think it will be great if we can fix this)

I have created a video summarizing the above 2 bugs:
https://drive.google.com/file/d/12_yp5wGq5rXJPcszUyZ-z1mMZc_tdjkC/view?usp=sharing

benalleng:
>     2. When pressing the left arrow key the `insertion pointer` of the terminal remains at the end of the line. (This bug is present on the live website as well, I think it will be great if we can fix this)

Unfortunately, looking into the code responsible for the terminal caret 'insertion pointer' It looks like it is handled by css and actually would not be able to be moved without substantially redesigning the current editor emulator we are using. There exist some terminal emulators that work in react but it seems as though we would have to essentially redesign our current challenges to work with them.

benalleng:
I know we are talking in the slack about some possible code-sandbox programs to work with as well

vinayaksh42 (approved):
Great work @benalleng and thanks a lot for addressing all the feedback. 
I guess it will be best to address this bug in a separate PR, If the code sandbox is integrated we can maybe change the logic being used for this challenge to that of the code sandbox.

ecurrencyhodler (approved):
Thanks for the review Vinayak!

Can you or ben submit an issue for the insertion pointer bug?

PR #144, opened 2023-02-11
Improved readability for signup modal

### LINKED ISSUE #111: Improve readability of the signup modal
I might be wrong so please close this issue if preferred.

I think the profile page font is not clear enough to read:
For example this text on the profile page:

Copy and store a simple code to save and load your progress on any device and browser. If you already have a code, load your progress here.
 
It's hard to read. Letters are mixed up

I believe either changing the background color or font size and text style may resolve this problem.

vinayaksh42:
Fixes #111 

- Fixed font and font size
- Avatar selection color and size

Addressed the issue mentioned on figma - https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=1194%3A14198&t=Hq8czLd2MYK4HZFW-0

**ScreenShot**
Screenshot 2023-02-11 at 7 32 11 PM


ecurrencyhodler:
tack. Confirmed that profile image doesn't change size and selection color is white.

afterburn:
Hey @GBKS do you think we'll have to omplement more modals in the future, if the answer is yes, do you think they will have different sizes or should we try to keep them roughly the same size and maybe introduce some variants (e.g. small, medium large)?

I'm asking because I'm seeing a lot of arbitrary values in our styling which will become hard to maintain if we're not careful, it would be better to define some sizes in our Tailwind config and only use those.

GBKS:
@afterburn by sizes, do you mean just the width?

Overall, it's better to avoid having so many modals. I don't foresee us having very many more.

Standardizing the modal frame sounds like a good idea. There is a bit of complex logic around accessibility and preventing scrolling of the page in the background on mobile that would be good to just do once.

afterburn:
@GBKS No, I was talking about both width and height, but I'm basically just advocating for standardization of the modal frames like you suggested. Not relevant for this PR perse, we can address it in a separate issue.

components/Modals/SignUp.tsx
```diff
@@ -38,59 +39,71 @@ export default function SignUpModal({ open, onClose, onConfirm }) {
...
+              className="flex h-[84px] w-[84px] flex-col items-center"
+              key={i}
+            >
+              <Avatar
+                avatar={i}
+                size={80}
+                onClick={() => setAvatar(i)}
+                classes={clsx(
+                  'h-20 w-20 rounded-full',
+                  {
```
afterburn, line 73: This can become a single object:
```jsx
{
  'border-2 border-white': avatar === i,
  'border-2 border-transparent': avatar !== i
}
```
And you can move `border-2` to `h-20 w-20 rounded-full` since it is applied in both cases.

afterburn (changes requested):
Left a comment, other than that it looks good.

GBKS:
@afterburn standardizing height in responsive design is a bit futile in my experience. Content size varies based on screen size, people change font sizes and weight via accessibility settings or website zoom, etc. It works in some specific cases, like grid items with highly consistent content (but even then you'd use flex box stretch so the shorter items grow to the height of the tallest one).

afterburn:
> @afterburn standardizing height in responsive design is a bit futile in my experience. Content size varies based on screen size, people change font sizes and weight via accessibility settings or website zoom, etc. It works in some specific cases, like grid items with highly consistent content (but even then you'd use flex box stretch so the shorter items grow to the height of the tallest one).

I'm mostly referring to giving the modal frame itself a standard height, not it's content. We'd render a scrollbar in the modal frame element when its content would overflow. On mobile we could make the modal span the entire screen to prevent any weird issues. I don't know the particulars with regard to accessibility, maybe when you have time you could do some designs for this, if you haven't already?

GBKS:
@afterburn sorry, but what is a reason for artificially limiting the height and making users scroll? 

On mobile, it's often a good idea to simply take over the whole screen with a modal. Lots of apps work that way as well when a temporary screen is shown (to briefly view/edit something before going back to the main flow).

image

afterburn:
> @afterburn sorry, but what is a reason for artificially limiting the height and making users scroll?

I'm simply saying when a modal exceeds the viewport height (which I'm not sure there currently are situations where this happens), we should make sure that instead of the modal frame exceeding the viewport height we should limit the modal frame to be contained within the viewport, and handle overflowing of the modal content in the modal frame itself. This doesn't conflict with your designs.

> On mobile, it's often a good idea to simply take over the whole screen with a modal. Lots of apps work that way as well when a temporary screen is shown (to briefly view/edit something before going back to the main flow).

Yes that is actually what I meant, when I said "On mobile we could make the modal span the entire screen to prevent any weird issues." sorry if it came across differently.

afterburn (approved):

PR #146, opened 2023-02-12
Navbar update 2

shaavan:
Introducing the new Navbar for Mobile!

**To test:**
- Decrease the screen size to the width of the mobile screen and see the magic happen!

**Sneak Peak:**

https://user-images.githubusercontent.com/85434418/218328662-349c82bb-b893-4ce7-96ae-1c79bdcc6808.mov


ecurrencyhodler:
tACK! 

Great work! Thanks for tackling the mobile nav bar. Tested and this is what I see when I click on the navbar.
Screenshot 2023-02-12 at 10 15 42 AM

Only thing that's missing from the figma design is the `?` but not sure what exactly that's supposed to point to. But we can prob address that in a separate PR.
Screenshot 2023-02-12 at 10 18 16 AM

@GBKS do you know what the `?` is supposed to point to?

shaavan:
AFAIK, the `?` probably represents the `Help` link. But since we didn't have yet implemented the Help page. The `?` shall come later in the picture.

ecurrencyhodler:
Sounds good! Let me follow up with christoph to see if we have something designed for the `?` button on mobile.

GBKS:
Awesome work, @shaavan. This looks great. I found a few things to tweak.

A tricky one is the top positioning of the menu. You have it fixed to 67px. However, users may use accessibility settings to scale text, in which the header height changes. Ideally, this position can be dynamic based on the current header bar height.

This is maybe not code you touched, but the top line in the header should never be break to 2 lines. It should just read "Chapter 1", not sure how the chapter title also snuck in there. For this line, it might also be good to apply an ellipsis if it's too long ([see CSS here](https://caniuse.com/?search=text-overflow%3A%20ellipsis)).

Font sizes need updating to match the design. They are a bit small right now (for visual tweaks, the easiest way to see if it's right is to take a screenshot and overlay it over the design in Figma).

Let's make the background fully opaque. The transparency makes it hard to read the nav menu text when the background copy also "shines through".

I should not be able to click on locked challenges.

The hamburger button needs some accessibility tweaks. It should either be a `button` tag, or have the `role="button"` attribute. It also needs a title that reads "Show menu" or "Hide menu" based on its state, as well as an `aria-pressed`
 attribute. Without those, a screen reader (and therefore people with low vision) does not know what this element does. It also needs [`aria-controls`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls) set to the ID of the menu.

The menu itself also needs some accessibility tweaks.
- [`aria-modal`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-modal)
- [`aria-hidden`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden) when it's closed so screen readers ignore it then
- [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) to state what it is  (maybe "Chapter navigation"?)

Modal accessibility is a it of a mess. Might be good if we mark it up as we think is correct, and then simple test it (we have some resources [here](https://bitcoin.design/guide/designing-products/accessibility/#usability-testing-resources)).

And a small animation thing, could the individual challenges also slide in in a staggered fashion? Might not be too tricky using `transition-delay` based on the index. I think you can also slightly slow down the animation. It's blazing fast right now.

GBKS:
And a note on the help icon, the idea was to give people a help page in case they get completely lost. It was in my initial mock-ups, but so far we have not discovered a need for a help page, and if we design the whole experience well, maybe people don't need it? I'd say, let's not implement a help icon/page, and see if the need arises in the future.

app/chapters/[slug]/layout.tsx
```diff
@@ -1,13 +1,37 @@
...
+function getWindowWidth() {
+  const { innerWidth: width } = window
+  return width
+}
 
-export default async function Layout({ children, params }) {
+export default function Layout({ children, params }) {
   const isValidChapter = chapters[params.slug]
 
+  const desktopSize = 800
```
Comment (vinayaksh42) [inline app/chapters/[slug]/layout.tsx:15]: Instead of `800` we should use `767` as the breaking point for mobile UI, we are already utilizing 767 for several components and pages

Comment (GBKS) [reply app/chapters/[slug]/layout.tsx:15]: Probably ideal if this number is whatever Tailwind is configured to for responsive sizes.

Comment (vinayaksh42) [reply app/chapters/[slug]/layout.tsx:15]: Yes, we should use the default Tailwind screen size breakpoints - https://tailwindcss.com/docs/screens#fixed-range-breakpoints

Comment (shaavan) [reply app/chapters/[slug]/layout.tsx:15]: Sure @vinayaksh42. Let me fix this up!

shaavan:
Thanks a lot, @GBKS, for your detailed review.

I read through all the points you mentioned, and all of them seem like a great suggestions. However, there's one I think I didn't properly understand.

> This is maybe not code you touched, but the top line in the header should never be break to 2 lines. It should just read "Chapter 1", not sure how the chapter title also snuck in there. For this line, it might also be good to apply an ellipsis if it's too long ([see CSS here](https://caniuse.com/?search=text-overflow%3A%20ellipsis)).

So would you please explain this point in detail? Thanks!

app/chapters/[slug]/layout.tsx
```diff
@@ -1,13 +1,37 @@
...
+
+  const windowWidth = getWindowWidth()
+  const [isDesktopSize, setIsDesktopSize] = useState(windowWidth > desktopSize)
```
Comment (afterburn) [inline app/chapters/[slug]/layout.tsx:18]: Let's just use `window.innerWidth` directly here instead of creating a function for it (`getWindowWidth`).

components/Navbar/NavbarMobile/HamburgerMenu.tsx
```diff
@@ -0,0 +1,37 @@
...
+      width: props.isOpen ? '20px' : '15px',
+    },
+  }
+
+  return (
+    <div
+      className="mr flex h-8 w-10 cursor-pointer flex-col items-start justify-center overflow-hidden border-r border-white/25 p-1"
+      onClick={props.clicked}
+    >
+      <div style={{ ...styles.line, ...styles.lineTop }} />
```
afterburn, line 33: Hey @shaavan, please use Tailwind, we don't use inline styles in this project.

components/Navbar/NavbarMobile/Menu.tsx
```diff
@@ -0,0 +1,26 @@
...
+    container: {
+      height: props.isOpen ? 'calc(100% - 67px)' : 0,
+      transition: 'height 0.3s ease',
+    },
+  }
+
+  return (
+    <div
+      className="absolute top-[67px] z-10 flex w-full flex-col blur"
+      style={styles.container}
```
afterburn, line 15: Same here, instead of using inline styles please use Tailwind, you can use [clsx ](https://www.npmjs.com/package/clsx) if necessary.

package.json
```diff
@@ -16,6 +16,7 @@
     "**/*.{js,jsx,ts,tsx}": "prettier --write"
   },
   "dependencies": {
+    "@headlessui/react": "^1.7.10",
```
afterburn, line 19: I don't have a lot of experience with transitions in react but let's try to use [react-transition-group](https://github.com/reactjs/react-transition-group/tree/v1-stable) which seems to be the package [suggested by the React team](https://reactjs.org/docs/animation.html), instead of adding a new ui library just so we can use their `Transition` component.

afterburn (changes requested):
Thanks @shaavan, looks great from the gif! I did leave some comments for you to resolve  before we can merge this.

app/chapters/[slug]/layout.tsx
Comment (afterburn) [reply app/chapters/[slug]/layout.tsx:18]: Also now that I think about it, please use the `useMediaQuery` hook, you can find an example in `TerminalChallenge.tsx`

shaavan:
Updated from aaf7152ae71277ea33e029d737887833297b4004 to b5210a1c3c0612450de52ddea543f2984248d6bd ([pr146.01](https://github.com/shaavan/saving-satoshi/commits/pr146.01) -> [pr146.02](https://github.com/shaavan/saving-satoshi/commits/pr146.02), [diff](https://github.com/shaavan/saving-satoshi/compare/pr146.01..pr146.02))

Addressed @GBKS, @vinayaksh42, @afterburn comments. Thank you very much, everyone, for your valuable reviews.


Differences:
1. Made DropDown position responsive to Navbar size.
2. Introduced a new global class and used it to make the DropDown menu background opaque.
3. Match font size with the designs for the dropdown menu.
4. Add accessibility fixes to the hamburger menu and dropdown menu.
5. Removed dependency on `@headlessui/react` for animation and used Tailwind for the animation.
6. Made animation staggered so that each element comes with a delay from each other.
7. Lowered the animation speed to make them more fluid.

8. Changed the breakpoint from `800` -> `767`.

9. Introduced a new useMediaQuery hook. This hook replaces the use of react’s default useMediaQuery which can cause a hydration failure error on initial loading.
10. Removed inline styles and replaced them with classes wherever applicable.

**Note:**

- In a few places to achieve the desired result, some dynamically named tailwind classes were needed. Since tailwind [advice against](https://tailwindcss.com/docs/content-configuration#dynamic-class-names) generating class names dynamically, I have used inline styles there.


**Demo:**

https://user-images.githubusercontent.com/85434418/219070353-f63af167-13a2-41fb-9f95-039836a6bd9b.mov

shaavan:
Updated from d23807f1d866006fabd5a90a16f045a5cdf3c08d to 174b570fb3c054a601117a45bcac0415c2d8a132 ([pr146.02](https://github.com/shaavan/saving-satoshi/commits/pr146.02) -> [pr146.03](https://github.com/shaavan/saving-satoshi/commits/pr146.03), [diff](https://github.com/shaavan/saving-satoshi/compare/pr146.02..pr146.03))


Changes:
1. Add the class responsible for the dropdown menu's opaque background in globals.css.

ecurrencyhodler:
Hmmm the build seems to be failing. Ping me again when it's resolved and I'll test again! But thanks for working on all the requested changes!

shaavan:
> Hmmm the build seems to be failing.

I am not able to pinpoint the reason for the failure, and I can't access the build logs as well by clicking the Vercel preview link.

So is there a way I can access the build docs? Thanks!

afterburn:
I see an unresolved conflict, maybe that's why the build is failing?

shaavan:
Updated from 174b570fb3c054a601117a45bcac0415c2d8a132 to cb96ca7d2b938dbb6cf83a736c6bfdaa69934306 ([pr146.03](https://github.com/shaavan/saving-satoshi/commits/pr146.03) -> [pr146.04](https://github.com/shaavan/saving-satoshi/commits/pr146.04), [diff](https://github.com/shaavan/saving-satoshi/compare/pr146.03..pr146.04))

Changes:

- Fixed the build failure issue caused by an iterated div not having a key value.
- Also addressed the warning that came up along with the preceding error in the console.

afterburn:
Hey @shaavan not sure if this is Github being weird or not but I still see a conflict here:

image

shaavan:
Updated from cb96ca7d2b938dbb6cf83a736c6bfdaa69934306 to 7d5c5d449dc255a1d72892d9fcf953a03ad6f7fb ([pr146.04](https://github.com/shaavan/saving-satoshi/commits/pr146.04) -> [pr146.05](https://github.com/shaavan/saving-satoshi/commits/pr146.05))

Changes:

- Rebased over master. 
- Resolved conflicting files.
- Made appropriate changes using `useLang` and `useTranslation` to correctly display the title and lesson name in the DropDown menu.

ecurrencyhodler:
Tested it and it works great! Looks like pr is blocked by a change request still. 

@GBKS How do you feel about the animated sliding menu?

app/[lang]/chapters/[slug]/layout.tsx
```diff
@@ -1,13 +1,21 @@
...
+  const isSmallScreen = useMediaQuery('(max-width: 767px)')
 
   return (
     <div className="ch1-background flex flex-col">
       <div className="fix-grow-issue flex min-h-screen flex-col blur">
-        {isValidChapter && <Navbar params={params} />}
+        {isValidChapter && isSmallScreen ? (
+          <NavbarMobile params={params} />
+        ) : (
+          <Navbar params={params} />
```
Comment (vinayaksh42) [inline app/[lang]/chapters/[slug]/layout.tsx:17]: the `Navbar` component should not be displayed to the user if the chapter is not valid. for example this page - https://saving-satoshi-git-fork-shaavan-navbar-update-2-savingsatoshi.vercel.app/en/chapters/chapter2d should only display the text `Coming Soon...`. https://savingsatoshi.com/en/chapters/chapter2d (what it looks like on saving satoshi)

styles/globals.css
```diff
@@ -41,6 +41,12 @@ html {
   background: rgba(46, 64, 90, 0.75);
 }
 
+.blur-opaque {
+  filter: drop-shadow(0px 4px 4px rgba(0, 0, 0, 0.25));
+  backdrop-filter: blur(75px);
+  background: rgba(46, 64, 90);
```
vinayaksh42, line 47: The `chapter-1` has an overall blue theme and having the RGB(46,64,90) works for this challenge but what about the other challenges that will be added in future, for example - [figma link](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=65%3A1012&t=yi7Typ86UjVL20qs-0), it has a red theme.

vinayaksh42 (changes requested):

app/[lang]/chapters/[slug]/layout.tsx
Comment (afterburn) [reply app/[lang]/chapters/[slug]/layout.tsx:17]: Good catch @vinayaksh42! I think we should also make sure the height of the content doesn't exceed that of the viewport on that page @shaavan.

shaavan:
Updated from 7d5c5d449dc255a1d72892d9fcf953a03ad6f7fb to d884f7a5c4caca0536fa42a21125dbe7739111fe ([pr146.05](https://github.com/shaavan/saving-satoshi/commits/pr146.05) -> [pr146.06](https://github.com/shaavan/saving-satoshi/commits/pr146.06), [diff](https://github.com/shaavan/saving-satoshi/compare/pr146.05..pr146.06))
Addressed @vinayaksh42 and @afterburn's suggestions.

- Fixed the layout.tsx file so that it will not display the navbar when the chapter is not valid.
- Fixed the "Coming Soon" page to take the viewport size and not display the scrollbar.

app/[lang]/chapters/[slug]/layout.tsx
Comment (shaavan) [reply app/[lang]/chapters/[slug]/layout.tsx]: Thanks for catching that, @vinayaksh42.

I have corrected the code to display the navbar only when the chapter is valid.

Addressed in d884f7a5c4caca0536fa42a21125dbe7739111fe

Comment (shaavan) [reply app/[lang]/chapters/[slug]/layout.tsx]: @afterburn
I have also addressed your suggestion in d884f7a5c4caca0536fa42a21125dbe7739111fe

The container div on the "Coming Soon" page doesn't overflow from the viewport and triggers the scroll bar.

However, it is better to split this commit into a separate PR since this PR focuses on the Navbar update.

styles/globals.css
shaavan: The `blur-opaque` class is the "opaque" version of the `blur` class which is used by **layout.tsx** to display its background.
And similar to the `blur-opaque` class, the `blur` class also uses static color.

Once new chapters get introduced in the codebase, we can make both these classes have dynamic color properties.

app/chapters/[slug]/layout.tsx
Comment (shaavan) [reply app/chapters/[slug]/layout.tsx]: Addressed in b5210a1c3c0612450de52ddea543f2984248d6bd

package.json
shaavan: Addressed in b5210a1c3c0612450de52ddea543f2984248d6bd by using plain tailwind.

components/Navbar/NavbarMobile/HamburgerMenu.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <div
+      className="mr flex h-full w-10 cursor-pointer flex-col items-start justify-center overflow-hidden border-r border-white/25 p-1"
+      onClick={props.clicked}
+      role="button"
+      aria-label={aria_label}
+      aria-expanded={props.isOpen}
+      aria-controls="lesson-navigation-menu"
+    >
+      <div
+        className={clsx(className, 'mb-[5px]', {
```
afterburn, line 20: Hey @shaavan, no need to define a className variable, just inline the css classes here.

components/Navbar/NavbarMobile/HamburgerMenu.tsx
```diff
@@ -0,0 +1,37 @@
...
+          'transform-none': !props.isOpen,
+          'translate-x-0 -translate-y-[1.8px] -rotate-45': props.isOpen,
+        })}
+        aria-hidden="true"
+      />
+
+      <div
+        className={clsx(className, 'mt-[2px]', {
+          'w-[15px] transform-none': !props.isOpen,
+          'w-[20px] translate-x-[1.2px] translate-y-[3px] rotate-45':
```
afterburn, line 30: Do we really need subpixel values like `1.2px`? I'm not a big fan of arbitrary values in tailwind, I think they become hard to maintain, let alone subpixel ones.

components/Navbar/NavbarMobile/TabMobile.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const isActive = challenge.lessonId === challengeId
+  const isLast = index == count - 1
+
+  return (
+    <Link
+      href={`/chapters/${slug}/${challenge.lessonId}`}
+      title={t(challenge.title)}
+      onClick={() => clicked()}
+      className={clsx(
+        'justify-left flex items-center border-t border-white/25 py-[9.5px] text-center text-[21px] text-white/50 transition duration-100 ease-in-out',
```
afterburn, line 46: Here a subpixel value (and arbitrary values) again, would be better if we can avoid those.

app/[lang]/chapters/[slug]/layout.tsx
```diff
@@ -1,13 +1,23 @@
-import Navbar from 'components/Navbar'
+'use client'
+
+import Navbar, { NavbarMobile } from 'components/Navbar'
 import { chapters } from 'content'
+import useMediaQuery from 'hooks/useMediaQuery'
```
Comment (afterburn) [inline app/[lang]/chapters/[slug]/layout.tsx:5]: Please use the following, would better match how we import other hooks:
```
import { useMediaQuery } from 'hooks'
```

styles/globals.css
afterburn: I'm just wondering why we're doing this via our `globals.css` file. Why not define a new color in our `tailwind.config.js` file and use the resulting tailwind class instead? If we then need to support a new color we just add that too and use its resulting tailwind class later on when we implement new challenges that have a different design.

afterburn:
I also noticed some issue with the styling when you close the menu. There is this weird blurry element that stays visible until the animation has completed, it doesn't change in height:

image

On Windows Chrome version 110.0.5481.96

app/[lang]/chapters/[slug]/layout.tsx
```diff
@@ -1,13 +1,23 @@
...
 
-export default async function Layout({ children, params }) {
+export default function Layout({ children, params }) {
   const isValidChapter = chapters[params.slug]
+  const isSmallScreen = useMediaQuery('(max-width: 767px)')
```
Comment (gadabra) [inline app/[lang]/chapters/[slug]/layout.tsx:9]: https://user-images.githubusercontent.com/100441043/220091481-1b4e2d30-3abc-46dc-9c76-d168bdaa937d.mov

After reloading a page,` <Navbar />` appears within a few milliseconds, followed by `<NavbarMobile />` flipping. To fix this, you can use conditions in Tailwind (CSS) instead of React (JS). Here's a simple example of the code I wrote to give an idea. Class name most likely should be inside of Navbars.

```jsx
<div className='md:hidden'>
  <NavbarMobile params={params} />
</div>
<div className='hidden md:block'>
  <Navbar params={params} />
</div>
```

shaavan:
Updated from d884f7a5c4caca0536fa42a21125dbe7739111fe to ea822132cc3d13887c40e033b9b5d1605f6ed579 ([pr146.06](https://github.com/shaavan/saving-satoshi/commits/pr146.06) -> [pr146.07](https://github.com/shaavan/saving-satoshi/commits/pr146.07), [diff](https://github.com/shaavan/saving-satoshi/compare/pr146.06..pr146.07))

Addressed @afterburn and @gadabra's suggestions.

Changes:

1. Removed variable `className` and used inline classes in **Hamburger.tsx**
2. Replaced arbitrary values and subpixel values with tailwind classes wherever needed. The subpixel values in **Hamburger.tsx** are not changed because that will make the hamburger menu "cross state" slightly off.
3. Changed the method of importing `useMediaQuery` to match the convention.
4. Used tailwind breakpoint instead of `useMediaQuery` to select which navbar to display in the **layout.tsx**.

app/[lang]/chapters/[slug]/layout.tsx
Comment (shaavan) [reply app/[lang]/chapters/[slug]/layout.tsx]: Addressed in ea822132cc3d13887c40e033b9b5d1605f6ed579

Thanks a lot for the suggestion!

It made the code respond better to screen size while simplifying the code.
I went with the approach that you have suggested in the example instead of adding className `md` in individual components to keep better track of what components will be displayed and when in the same place.

components/Navbar/NavbarMobile/HamburgerMenu.tsx
shaavan: Addressed in ea822132cc3d13887c40e033b9b5d1605f6ed579
Thanks for the suggestion!

I went with extracting class names into variables to decrease repetition. But I understand your point in keeping the code complexity minimal.

components/Navbar/NavbarMobile/HamburgerMenu.tsx
shaavan: I experimented with changing the values to the nearest integer pixel here, but it made the "cross state" of the hamburger menu slightly off when displayed.

On a smaller screen, it would be more evident, so I didn't touch the subpixel value here for now. But I would love to hear your take on this point.

components/Navbar/NavbarMobile/TabMobile.tsx
shaavan: Addressed in ea822132cc3d13887c40e033b9b5d1605f6ed579
Thanks for the suggestion!

app/[lang]/chapters/[slug]/layout.tsx
Comment (shaavan) [reply app/[lang]/chapters/[slug]/layout.tsx]: Addressed in ea822132cc3d13887c40e033b9b5d1605f6ed579
Thanks!

styles/globals.css
shaavan: Addressed in ea822132cc3d13887c40e033b9b5d1605f6ed579

Defined new custom theme elements for `opacity` `backdropBlur` and `dropShadow` in **tailwind.config.js** and used them in place of the `blur` and `blur-opaque` classes.

shaavan:
@afterburn 

> I also noticed some issue with the styling when you close the menu. There is this weird blurry element that stays visible until the animation has completed, it doesn't change in height

I tried running this PR on **Windows 11 Firefox** but could not replicate this problem.
Also, it runs perfectly well on **Mac OS Chrome**. So my guess is that this issue is unrelated to the additions in PR.

It's best if others could check and try replicating this issue.

vinayaksh42:
While attempting to reproduce the issue reported in Safari browser (iPhone), I encountered the following problems:

1. The border on the right side of the back button appears to be slightly shorter, making it unable to stretch across the entire length. (See image below.)

Screenshot 2023-02-21 at 8 25 54 PM

2. When the menu is activated and I scroll down past its height, I can view all the content below the menu. (Watch this video for a better understanding: https://drive.google.com/file/d/1Boc-xZOP9IWaufAntH6TF2RtUN6MpMoh/view?usp=sharing)

gadabra:
@shaavan  I think a little bit refactoring makes it is more readable. I would try to describe it like this.

1. in [layout](https://github.com/shaavan/saving-satoshi/blob/ea822132cc3d13887c40e033b9b5d1605f6ed579/app/%5Blang%5D/chapters/%5Bslug%5D/layout.tsx#L12) div's are not necessary. You can write just these code
```
 <Navbar params={params} />
 {children}
```

2. relocate `md:hidden` and `hidden md:block` classes on outer div element in `NavbarMobile` and `Navbar`. 

3. Rename `Navbar.tsx` as `NavbarDesktop.tsx`

4. Create new component in `Navbar.tsx` to bring  `NavbarMobile` and `NavbarDesktop` together.

```
export default function Navbar({ params }) {
  return (
    <>
      <NavbarMobile params={params} />
      <NavbarDesktop params={params} />
    </>
  )
}
```

In this way `Navbar` and `Layout` more intact, they are doing only their jobs. Layout component does not decide which Navbar should be displayed.
 
If it does not work we can make quick huddle call in slack.

components/Navbar/NavbarMobile/HamburgerMenu.tsx
GBKS: This is a unique situation where we do need super fine-grained control to get the bars to cross visually in just the right way, so I think it's warranted here. Of course not something you want to do all over the place.

GBKS:
Awesome progress on this one, I think we're very close. A few things I noticed in addition to the comments above:

- Try reducing the transition delays more. Right now it looks like each item is sliding in on its own. It should feel like they are coming in like a wave, not one by one.
- I am seeing a slow performance on my Android. This might be related to having a drop shadow and backdrop blur on `lesson-navigation-menu`. You can remove both of them. Especially the backdrop blur does not do anything since the element also has a flat color background.
- When loading the page on a wider screen, resizing it to mobile, and then expanding the menu, the menu will sit at the top of the screen (covering the top bar). Is it because the top bar height calculation happens only on initial load? Could this also be done right before the menu is being opened to prevent this problem?

shaavan:
Updated from ea822132cc3d13887c40e033b9b5d1605f6ed579 to 9f6aa1dd8e78f726bf242d2c7664781da500a7ee ([pr146.07](https://github.com/shaavan/saving-satoshi/commits/pr146.07) -> [pr146.08](https://github.com/shaavan/saving-satoshi/commits/pr146.08), [diff](https://github.com/shaavan/saving-satoshi/compare/pr146.07..pr146.08))
Addressed, @vinayaksh42 @gadabra @GBKS comments.

<h2>
Changes
</h2>

1. Fixed the classes for div for the arrow icon to take full navbar length for the border.
2. Refactored the code according to @gadabra's suggestions.
3. Reduced transition to give a smoother feeling to transition.
4. Removed dropShadow and blur from the dropdown menu, to fix performance issues on Android.
5. Fix the code to fix the issue pointed out by @GBKS here.

> When loading the page on a wider screen, resizing it to mobile, and then expanding the menu, the menu will sit at the top of the screen (covering the top bar). Is it because the top bar height calculation happens only on initial load? Could this also be done right before the menu is being opened to prevent this problem?

<h2>
Discussion:
</h2>

@vinayaksh42 
> When the menu is activated and I scroll down past its height, I can view all the content below the menu.

- I was not able to find a solution to this issue.
AFAIK, safari browser does not react correctly with a mixture of flex, absolute positioning, and calculation ([reference](https://stackoverflow.com/a/33644245))
So I tried replacing calculations with the tailwind `h-full` class, but Safari did not react correctly to this change. ([reference](https://github.com/tailwindlabs/discuss/issues/226))

@GBKS 
> When loading the page on a wider screen, resizing it to mobile, and then expanding the menu, the menu will sit at the top of the screen (covering the top bar). Is it because the top bar height calculation happens only on initial load? Could this also be done right before the menu is being opened to prevent this problem?

- I was experimenting to find a solution for this, and then I discovered not specifying the position for the dropdown menu automatically positions it correctly below the navbar. I don't know the cause of this behavior. But I checked, and it works perfectly on Chrome and Safari on Mac. And also on mobile and tablet devices (checked using Chrome devTool).
- This is an _experimental_ change that simplifies the code by removing the usage of `useEffect`, `useRef`, and `position` props.
- I urge reviewers to confirm the correct positioning of the dropDown menu on their system in this version of PR.

shaavan:
Updated from 9f6aa1dd8e78f726bf242d2c7664781da500a7ee to 99951980a68fdfcefa0feb103c21dd90832ec26f ([pr146.08](https://github.com/shaavan/saving-satoshi/commits/pr146.08) -> [pr146.09](https://github.com/shaavan/saving-satoshi/commits/pr146.09), [diff](https://github.com/shaavan/saving-satoshi/compare/pr146.08..pr146.09))

Correction:

- Value of transitionDelay changed from `120 (...) ms` -> `60 (...) ms`.

components/Navbar/NavbarMobile/HamburgerMenu.tsx
shaavan: I agree with @GBKS! Also, as we discussed on the call last Tuesday, to correctly render specific cases like hamburger cross, we can work with arbitrary subpixel values.

shaavan:
A gentle ping to my fellow contributors.

@afterburn, @GBKS, @vinayaksh42, @gadabra, @ecurrencyhodler

- All conversations have been resolved, and the PR is ready for another (and likely the last round) of reviews

app/[lang]/chapters/[slug]/layout.tsx
```diff
@@ -1,13 +1,15 @@
+'use client'
```
Comment (afterburn) [inline app/[lang]/chapters/[slug]/layout.tsx:1]: Any particular reason why this needed to become a Client component?

components/Navbar/NavbarMobile/HamburgerMenu.tsx
```diff
@@ -0,0 +1,41 @@
+import clsx from 'clsx'
+import React, { useState } from 'react'
+
+export default function HamburgerMenu(props) {
+  const aria_label = props.isOpen ? 'Close Menu' : 'Open Menu'
```
afterburn, line 5: Please use camelCase for variable names

components/Navbar/NavbarMobile/Menu.tsx
```diff
@@ -0,0 +1,27 @@
...
+      className={clsx('absolute z-10 w-full bg-back duration-[400ms]', {
+        'h-full': props.isOpen,
+        'h-0 delay-200': !props.isOpen,
+      })}
+      style={{}}
+      aria-modal="true"
+      aria-hidden={!props.isOpen}
+      aria-label="Lesson Navigation Menu"
+    >
+      {
```
afterburn, line 18: Why is `TabGroupMobile` inside an curly braces here? Did our formatter do that?

components/Navbar/NavbarMobile/TabGroupMobile.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const chapter = chapters[slug]
+
+  if (!chapter) {
+    return null
+  }
+
+  const chapterMeta = chapter?.metadata
+
+  const challenges = chapter.metadata.challenges.map((lessonId) => {
+    let { title } = lessons[lessonId].metadata
```
afterburn, line 22: Can be `const`

hooks/useMediaQuery.ts
```diff
@@ -0,0 +1,18 @@
+'use client'
+import { useState, useEffect } from 'react'
+
+export const useMediaQuery = (query) => {
+  const [matches, setMatches] = useState(false)
+
+  useEffect(() => {
+    const media = window.matchMedia(query)
```
afterburn, line 8: Nitpick but I think some blank lines here would help improve readability a little, something like:

```ts
useEffect(() => {
    const media = window.matchMedia(query)
    
    if (media.matches !== matches) {
      setMatches(media.matches)
    }
    
    const listener = () => setMatches(media.matches)
    
    window.addEventListener('resize', listener)
    return () => window.removeEventListener('resize', listener)
  }, [matches, query])
``

**Review by afterburn — CHANGES_REQUESTED** [review]
Thanks @shaavan, made a request for a couple small changes but otherwise code looks good to me.

**Comment by shaavan** [comment]
Updated from 99951980a68fdfcefa0feb103c21dd90832ec26f to 9b4c5743cb1375d2eae011ab956848b2f92682c6 ([pr146.09](https://github.com/shaavan/saving-satoshi/commits/pr146.09) -> [pr146.10](https://github.com/shaavan/saving-satoshi/commits/pr146.10))
Addressed @afterburn comments.

1. Rebased over the master to resolve conflict.
2. Addressed @afterburn's suggestion commented above.

File: app/[lang]/chapters/[slug]/layout.tsx (hunk 13 above)
Comment (shaavan) [reply app/[lang]/chapters/[slug]/layout.tsx]: I had used it earlier when I was calling hooks in this file. But after refactoring the code, I had forgotten to remove this.

Thanks for catching this.

Addressed in 9b4c5743cb1375d2eae011ab956848b2f92682c6

File: components/Navbar/NavbarMobile/HamburgerMenu.tsx (hunk 14 above)
Comment (shaavan) [reply components/Navbar/NavbarMobile/HamburgerMenu.tsx]: Renamed it in 9b4c5743cb1375d2eae011ab956848b2f92682c6

File: components/Navbar/NavbarMobile/TabGroupMobile.tsx (hunk 16 above)
Comment (shaavan) [reply components/Navbar/NavbarMobile/TabGroupMobile.tsx]: Thanks for catching this.

Addressed in 9b4c5743cb1375d2eae011ab956848b2f92682c6

File: hooks/useMediaQuery.ts (hunk 17 above)
Comment (shaavan) [reply hooks/useMediaQuery.ts:8]: Addressed in 9b4c5743cb1375d2eae011ab956848b2f92682c6

It does improve readability!

File: components/Navbar/NavbarMobile/Menu.tsx (hunk 15 above)
Comment (shaavan) [reply components/Navbar/NavbarMobile/Menu.tsx]: I also noticed them earlier but didn't touch them because I thought they might be here for some reason.

Turns out they weren't :)

Addressed in 9b4c5743cb1375d2eae011ab956848b2f92682c6

**Comment by ecurrencyhodler** [comment]
@vinayaksh42 , Shashwat said the following for a request you made:

"I was not able to find a solution to this issue (When the menu is activated and I scroll down past its height, I can view all the content below the menu.).

AFAIK, safari browser does not react correctly with a mixture of flex, absolute positioning, and calculation ([reference](https://stackoverflow.com/a/33644245)).
So I tried replacing calculations with the tailwind h-full class, but Safari did not react correctly to this change. (https://github.com/tailwindlabs/discuss/issues/226)."

I think we should create a separate issue for this. If you agree, can you resolve your request and open a separate issue for it? WDYT?

@shaavan after you address the last request from Kevin, I will test one more time and we can merge. Also I have to apologize. This issue was most definitely not a "good first issue" and I wrongly categorized it. But you've done a great job addressing and incorporating everyone's feedback. So thank you for your persistence.

**Comment by vinayaksh42** [comment]
> @vinayaksh42 , Shashwat said the following for a request you made:
> 
> "I was not able to find a solution to this issue (When the menu is activated and I scroll down past its height, I can view all the content below the menu.).
> 
> AFAIK, safari browser does not react correctly with a mixture of flex, absolute positioning, and calculation ([reference](https://stackoverflow.com/a/33644245)). So I tried replacing calculations with the tailwind h-full class, but Safari did not react correctly to this change. ([tailwindlabs/discuss#226](https://github.com/tailwindlabs/discuss/issues/226))."
> 
> I think we should create a separate issue for this. If you agree, can you resolve your request and open a separate issue for it? WDYT?

Sure, I think we can address this issue in a separate Pull Request. I'll quickly open an issue for the same.

**Comment by vinayaksh42** [comment]
@ecurrencyhodler I have created issue #213 for the safari bug. @shaavan, I suggest that we tackle this bug separately by creating a new PR. Alternatively, if you prefer, you can include it in the current PR.

**Comment by benalleng** [comment]
I noticed on the gnesis-2 challenge there is an odd behavior with the code and info buttons were growing more than expected.
image

For the life of me I can't see any reason why your code could be causing these changes though

**Comment by shaavan** [comment]
Updated from 9b4c5743cb1375d2eae011ab956848b2f92682c6 to db5bb4659fe2dfcadabe91d614b0482b5f6a051c ([pr146.10](https://github.com/shaavan/saving-satoshi/commits/pr146.10) -> [pr146.11](https://github.com/shaavan/saving-satoshi/commits/pr146.11))
Addressed @afterburn comment.

- Removed the redundant style property.
- Rebased over master.

**Comment by shaavan** [comment]
@ecurrencyhodler 
> @shaavan after you address the last request from Kevin, I will test one more time and we can merge. Also I have to apologize. This issue was most definitely not a "good first issue" and I wrongly categorized it. But you've done a great job addressing and incorporating everyone's feedback. So thank you for your persistence.

I have addressed @afterburn's suggestion, and the PR is ready to be tested. And no need to apologize. It was a fun challenge. I learned a lot of things while building this, and it was great seeing this PR slowly evolve and become more refined!

@vinayaksh42 

>  @shaavan, I suggest that we tackle this bug separately by creating a new PR.

Sure, it's best to tackle this issue in a separate PR.


@benalleng 

> I noticed on the gnesis-2 challenge there is an odd behavior with the code and info buttons were growing more than expected.

- I can reproduce this issue. 
- This behavior is present only with the second challenge, not with the first one.
- Also, I noticed one more thing. When I click the "Code" tab, the textual description does not "hide".

**Screenshot:**
image

- The property responsible for showing and hiding the text description (i.e., `isActive` in **ui/lesson/Info.tsx** file) is not triggering for the second challenge when switching tabs between Info and Code.

<details>
<summary> Patch to test </summary>

Step 1:

```diff --git a/ui/lesson/Info.tsx b/ui/lesson/Info.tsx
index 824c4c5..23e669e 100644
--- a/ui/lesson/Info.tsx
+++ b/ui/lesson/Info.tsx
@@ -16,6 +16,8 @@ export default function LessonInfo({ children }) {
     )
   }
 
+  console.log(isActive)
+
   return (
     <div
       className={clsx(
```
Step 2:

1. Observe that there is no output in the console when shifting tabs (between <strong>info</strong> and <strong>code</strong>) for challenge 2, but it shows for challenge 1.

</details>


- I think the bug you mentioned and this are somehow getting triggered by this PR. So let's address them in a separate PR.

ecurrencyhodler:
@afterburn I believe shaavan has addressed your change request. Can you take a quick look?

I'm going to test this PR as well.

ecurrencyhodler:
So testing this pr, the menu bar on mobile looks great. However in chrome, what I see is that the info from Ch1. Challenge 1 part 2 is now being carried over to the code tab. 

Below you see the tested PR version and on the right is prod.
bug

afterburn (approved):

afterburn:
I merged this, let's address remaining issues in a separate pr

ui/lesson/TerminalChallenge.tsx
```diff
@@ -3,7 +3,7 @@
 import { useState, useEffect } from 'react'
 import { LessonDirection } from 'types'
 import { Lesson, LessonTabs, LessonTerminal } from 'ui'
-import { useMediaQuery } from 'react-responsive'
+import { useMediaQuery } from 'hooks'
```
vinayaksh42, line 6: changing it back to `react-responsive` is fixing the error

benalleng: Is there something wrong with our useMediaQuerey fn? also is there a particular reason we have a media query hook and are not using the imported one instead?

GBKS:
> I noticed on the gnesis-2 challenge there is an odd behavior with the code and info buttons were growing more than expected. image
> 
> For the life of me I can't see any reason why your code could be causing these changes though

This is due to the grid layout. I switched this to flex box in #236 to fix it. Safari has a few "particularities" with grid due to following an older spec version, so flex box is usually a better way to go (also, we don't really have grids with multiple rows/columns).

PR #151, opened 2023-02-14
Fix for missing translation on Done page

afterburn:
## Summary
This PR fixes #150.

Please review the below PR requirements:

- [x] Build was run locally and any changes were pushed
- [x] Ensure that the title clearly describes the changes
- [x] Issue to be referenced in the PR description rather than in the commits or PR title
- [x] Clean commit history

PR #153, opened 2023-02-16
Fix for title leads to 404 page

### LINKED ISSUE #149: Saving Satoshi title leads to 404 page
**Describe the bug**
Click on the `Saving Satoshi` title leads to a 404 page

**To Reproduce**

1. Go to https://savingsatoshi.com/en/chapters
2. Click on the `Saving Satoshi` title
3. See the error

**Expected behavior**
Clicking on the title should lead to the homepage

vinayaksh42:
Fixes #149

Adding `/en` to the link in `Topbar.tsx` leads to the home page.

### DIFF
--- a/components/Topbar.tsx
+++ b/components/Topbar.tsx
@@ -11,7 +11,7 @@ export default function Topbar({ items }: { items: NavItem[] }) {
     <div className="absolute left-0 top-0 w-full">
       <div className="m-auto flex items-center justify-between px-6 py-4 text-white">
         <Link
-          href="/"
+          href="/en"
           className="transition duration-150 ease-in-out hover:opacity-75"
         >
           <h1 className="text-xl md:text-3xl">{siteConfig.name}</h1>
--- a/components/Topbar.tsx
+++ b/components/Topbar.tsx
@@ -5,13 +5,15 @@ import Link from 'next/link'
 import { siteConfig } from 'config/site'
 import { NavItem } from 'types'
 import UserButton from './Navbar/UserButton'
+import { useLang } from 'hooks'
 
 export default function Topbar({ items }: { items: NavItem[] }) {
+  const lang = useLang()
   return (
     <div className="absolute left-0 top-0 w-full">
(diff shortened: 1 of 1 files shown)


components/Topbar.tsx
```diff
@@ -11,7 +11,7 @@ export default function Topbar({ items }: { items: NavItem[] }) {
     <div className="absolute left-0 top-0 w-full">
       <div className="m-auto flex items-center justify-between px-6 py-4 text-white">
         <Link
-          href="/"
+          href="/en"
```
afterburn, line 14: This should be dynamic though, if a user is using another language we don't want to force the `en` version, so this is not the way we should fix this. In the `middleware.ts` file we redirect the user to their preferred locale, but something is wrong if the target page is the index file, needs some further investigation there.

afterburn (changes requested):

components/Topbar.tsx
vinayaksh42: Yes, I saw the changes you did here - https://github.com/saving-satoshi/saving-satoshi/pull/151/commits/239f7715ee29064a26637b1c73d3826b899d02cf in order to solve this issue and understood that it requires much more investigation than what I did before creating this PR. One solution that came to my mind is to redirect the user using the selected `locale` or we can create an expectation when we are redirecting to the homepage as this issue is only occurring when redirecting to the homepage (This only occurs in production server and not in local development enviroment)

afterburn: Yeah, I also tried reproducing this locally and wasn't able to. I think we can use your first suggestion, so we pass the `lang` as a prop to the `<Topbar lang={lang} />` component and redirect users to `/${lang}` when they click the title. I think things will have to change anyway once the Next 13 app directory approach matures a little bit more, so this is fine as a temporary fix.

afterburn (approved):
Ah good, I hadn't noticed Topbar was a Client component so you can indeed use `useLang` instead of passing lang as a prop from the pages.

PR #159, opened 2023-02-17
Input Area Max-width

### LINKED ISSUE #154: Input styling on Safari
>Sorry to report a bug right away, I'm finding that when I start entering text in the input field on Safari (desktop and mobile), it instantly switches from multi line to single line, stretching beyond the browser frame. This might just be a simple CSS setting (max-width or so).

On Safari there might be some browser specific styling that is not supported and is causing some strange rendering

**To Reproduce**

1. Go to [Input screen](https://savingsatoshi.com/en/chapters/chapter-1/genesis-1)
3. Apply mobile view by adjusting screen size or open with iphone
4. Begin typing in input area to reproduce error

**Expected behavior**
The styling should behave like other browsers

- Browser safari

I don't have easy access to test on safari so any more probing of this error would be appreciated! Any advice on how to test safari browser support with a Linux laptop?

benalleng:
Fixes #154 

Added max width of the screen to prevent any input from extending beyond the screen size.
[iPhone input screenshot](https://user-images.githubusercontent.com/108441023/219769902-41700422-ce19-4a81-8913-f1794cef8149.PNG)

[Input Preview](https://saving-satoshi-git-fork-benalleng-safariwidth-savingsatoshi.vercel.app/en/chapters/chapter-1/genesis-1)

With just an iPhone some more testing may be needed on a Mac but this particular safari bug seems to be solved

Tested on
Safari (iPhone)
Chrome (Windows)
firefox (Android)

### DIFF
--- a/ui/common/Input.tsx
+++ b/ui/common/Input.tsx
@@ -100,7 +100,7 @@ export default function Input({
         onChange={handleChange}
         value={textAreaValue}
         spellCheck="false"
-        className={`absolute top-0 left-0 h-full w-full resize-none overflow-hidden bg-transparent font-space-mono text-[18px] leading-[180%] tracking-[1px] text-transparent outline-none md:text-[30px] md:tracking-[5px]`}
+        className={`absolute top-0 left-0 h-full w-full max-w-[95vw] resize-none overflow-hidden bg-transparent font-space-mono text-[18px] leading-[180%] tracking-[1px] text-transparent outline-none md:text-[30px] md:tracking-[5px]`}
         style={{
           caretColor: '#6e7d92',
         }}
@@ -108,7 +108,7 @@ export default function Input({
       <p
         className={`${
           correctAnswer ? 'overlay-complete' : 'overlay-incomplete'
-        } pointer-events-none h-full w-full font-space-mono text-[18px] leading-[180%] tracking-[1px] text-inherit md:text-[30px] md:tracking-[5px]`}
+        } pointer-events-none h-full w-full max-w-[95vw] font-space-mono text-[18px] leading-[180%] tracking-[1px] text-inherit md:text-[30px] md:tracking-[5px]`}
         style={{
           lineBreak: 'anywhere',
         }}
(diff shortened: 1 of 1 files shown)


ui/common/Input.tsx
```diff
@@ -100,15 +100,15 @@ export default function Input({
         onChange={handleChange}
         value={textAreaValue}
         spellCheck="false"
-        className={`absolute top-0 left-0 h-full w-full resize-none overflow-hidden bg-transparent font-space-mono text-[18px] leading-[180%] tracking-[1px] text-transparent outline-none md:text-[30px] md:tracking-[5px]`}
+        className={`absolute top-0 left-0 h-full w-full max-w-[95vw] resize-none overflow-hidden bg-transparent font-space-mono text-[18px] leading-[180%] tracking-[1px] text-transparent outline-none md:text-[30px] md:tracking-[5px]`}
```
afterburn, line 103: Hey @benalleng, any particular reason why you've chosen `95vw` here?

ecurrencyhodler:
Thanks ben for taking this on!

I tested it and it looks like the width expands to screen wide if any action is taken on the coding challenge in desktop. 

Here are two scenarios where I type and refresh the page:
type and refresh

Here is one where I click on the link:
click explorer

The same is true for mobile but the difference isn't as noticeable. Is there a way where the width that the user sees the moment they land on the page stays the same?

ui/common/Input.tsx
benalleng: So at first, I set the max width to the screen width but I noticed that the input wasn't respecting any padding from its parent elements and was filling the screen completely. Now that you've pointed it out I realize that may mean there is an overflow error I just now tried `word-break: break-all` and that also fixes the problem. In addition to the input now respecting CSS padding from its parent elements. Thanks for pushing me to check twice on this one!

benalleng:
@afterburn @ecurrencyhodler I got a little hasty with the solution. Now that it is simply a word-break there 'shouldn't' be any more odd width changes or arbitrary values. Sorry!

ecurrencyhodler:
tACK! No width changes on mobile and desktop. 🚀

afterburn (approved):

afterburn:
Unable to test this myself (no machine capable of running safari) but code is fine so feel free to merge.

Issue #152, opened 2023-02-14
Use new translation system throughout codebase


afterburn:
**Current situation**
Only some of our components and content are using the new translation system.

**Desired situation**
The entire app should use the new translations system.

**Notes**
Let's first complete #148 before we pick this up, should be easier to test this that way.

I've provided some examples in `components/translations.ts` for the `Hero.tsx` component.
For content you can take a look at `content/genesis/genesis-1/translations.ts`

benalleng:
The translation system is almost there! with #245 and #279 now merged into master, many of the translation bugs we encounter have now been solved.
We are still dealing with two things moving forward.
- [x] #256 flashing white screen following a language switch after loading the page
- [x] #250 both languages visible in the same render 
- [ ] There is no easily accessible language selection, addressed by #148
- [ ] #313 Upon clicking a link the locale will always switch back to the default `/en` url and load the english language
There are two ways to solve this IMO

ecurrencyhodler:
@afterburn Do you 2 pr's close this issue out?

PR #169, opened 2023-02-21
New translations system

benalleng:
Works towards closing #152.

I have some questions on how to best plan for the translations moving forward
1. How should reused translations like a copy or continue button be handled?
2. Should there be a set template to ensure easy translation changes so that each translation page can be quickly replicated.
I think making a simple translation key template would help with 2 and I went ahead and listed them below.
```
const translations = {
    ${route}: {
        ${lang}: {
            title: ‘Lorem ipsum’,
            heading: ‘Lorem ipsum’,
            subheading_one: ‘Lorem ipsum’,
            paragraph_one: ‘Lorem ipsum dolor sit amet, ex ea commodo consequat.’,
            link: ‘Click Me!’,
            paragraph_two: ‘Lorem ipsum dolor sit amet, ex ea commodo consequat.’,
            ${component_object_param}: ‘Error Message’,
            next: ‘Continue’,
            ...
            paragraph_nine: ‘Lorem ipsum dolor sit amet, ex ea commodo consequat.’,,
            paragraph_(n): ‘Lorem ipsum dolor sit amet, ex ea commodo consequat.’,
        },
    },
}
```
Most notably I think having some syntax-ordering i.e. not having all paragraphs together but ordering them how they appear on the page can help with quickly showing what work needs to be completed in the translations


afterburn:
1. I suggest we create a translations file at the root of our project that we can use to define translations that are commonly reused. Something along the lines of: 
```ts
export const translations = {
  common: {
    en: {
      continue: 'Continue'
    }
  }
}
```
This file would then have to be imported in `lib/translate.tsx` as well, at least until we implement an auto-import translations solution.

2. I mean yeah that is how all existing translation files have been set up so far. To me it makes sense to continue with that approach. Also means we don't have to support other translation structures in `lib/translate.tsx` . I'm referring to these top-level ones:
```ts
export const translations = {
  [namespace]: {
    [lang]: {
      ...
    }
  }
}
```
As for how we name the translations keys that go into `...`, I don't really care as long as it's `snake_case` and makes at least some sense 😁.

afterburn (changes requested):
Code looks good but noticed this issue in the Topbar/Navbar:

image

It should say: "Chapter 1 - Challenge 2"

lib/translate.tsx
```diff
@@ -75,7 +71,9 @@ export function t(key: string, lang: string) {
   let translation = get(translations, `${lang}.${key}`)
 
   if (!translation) {
-    return `{${key}}`
+    // console.log(`No translation for {${lang}.${key}}`)
+    let fallbackTranslation = get(translations, `en.${key}`)
```
afterburn, line 75:
```js
// Fallback translation
return get(translations, `en.${key}`)
```

afterburn:
@ecurrencyhodler feel free to merge this once it has your tACK

ecurrencyhodler:
tACK!

PR #190, opened 2023-02-24
Chapter Introduction scroll

### LINKED ISSUE #166: Too much height on Chapter Intro page
**Describe the bug**
Not sure if this is a bug or if its intended (@GBKS) but on the intro page for a chapter there is this extra height which causes a scrollbar to appear. It seems to be the exact height of the Topbar, so we should probably just subtract that from the content height.

**To Reproduce**

1. Go to Chapter 1 Intro page.
2. Observe the scrollbar caused by extra height.

**Expected behavior**
I expect the scrollbar to only be visible if the content exceeds the viewport height.

**Screenshots and/or logs**
image

benalleng:
Closes #166

This problem was a little more involved than i first anticipated but only because we are using the <Image /> component from next.js and not a standard  html tag. Once i realized this there is clear documentation on how to utilize the parameters to get it to behave how we expect it to.


benalleng:
Wrong code. Closed

PR #191, opened 2023-02-24
Chapter intro page Scroll

benalleng:
Closes https://github.com/saving-satoshi/saving-satoshi/issues/166

This problem was a little more involved than i first anticipated but only because we are using the component from next.js and not a standard html tag. Once i realized this there is clear documentation on how to utilize the parameters to get it to behave how we expect it to.

[Preview](https://saving-satoshi-git-fork-benalleng-chapter-67b81e-saving-satoshi.vercel.app/en/chapters/chapter-1)

### DIFF
--- a/ui/introduction/Introduction.tsx
+++ b/ui/introduction/Introduction.tsx
@@ -21,13 +21,24 @@ export default function Introduction({ children, lang }) {
   return (
     <div className="flex grow">
       <div className="lg:flex lg:grow">
-        <div className="lg:order-last lg:flex lg:shrink lg:basis-1/2 lg:border-l lg:border-white/25">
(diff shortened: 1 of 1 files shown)


benalleng:
As an aside is this the wrong picture for the chapter?

ecurrencyhodler:
tACK! Also tested on mobile. 
tACK

afterburn (approved):

GBKS:
Tested and works. Although not a super clean solution, as the code includes the image twice with slightly different styling. The second image is still visible and rendered on mobile, it's just that the container height is 0 so you can't see it. There's probably a cleaner way to do this, but this works great for now, so let's not worry about it (IMHO).

PR #194, opened 2023-02-24
Improve success message for Genesis-2 Terminal challenge

### LINKED ISSUE #182: Add a more helpful success message to chapter 1, challenge 1, page 2
It just reads "Great job!". We can give some input on why this message means and its importance on bitcoin culture. See a proposed change in [Figma here](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=1604%3A21834&t=wtEGAyNRp9cwYerM-1).

localhost_3000_en_chapters_chapter-1_genesis-1 (1)

afterburn:
## Summary
This PR should close #182. It also adds support for `<Link>`'s in translations.

## Checklist
- [x] Clean commit history


afterburn:
According to the [Figma file](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=1604%3A21834&t=Tblr6nkhyTgZNBKe-0) `"here"` should be a link. I couldn't find a satisfying website where users can read the article, nor a pdf or something similar. So instead I found a picture of the Times article and added it as an image to our project. Perhaps someone else has a better alternative?

lib/translate.tsx
```diff
@@ -66,44 +72,83 @@ export function t(key: string, lang: string) {
...
-
-  while ((match = tooltipRegex.exec(result))) {
-    const tooltipHtml = match[0]
-    const tkey = tooltipHtml.match(contentRegex)[1]
+  let result = []
 
-    const hrefMatch = tooltipHtml.match(hrefRegex)
-    const href = hrefMatch?.length > 0 ? hrefMatch[1] : null
+  result = injectComponent([translation], ComponentType.Link, lang)
+  result = injectComponent(result, ComponentType.Tooltip, lang)
```
shaavan, line 91: I have a code architecture-related question. 

Since you are checking for both the Link regex and Tooltip regex in the string for every string. Could you move this check inside the function?

Since the function automatically stops calculation if it does not find a regex match,
```js
while ((match = regex.exec(part))) { ...
```
why not just let it do both the checks inside the `injectComponent` function?

shaavan (commented):
The changes look good to me.

This PR also introduces a regex check of `Link` for the success message in the string, and moves check for Link Regex and ToolTip regex in the same function.

The link is displayed correctly in the success string in the Terminal output, and points to the right article (which for now is the screenshot of the original times article).

lib/translate.tsx
afterburn: Hi @shaavan, I'm assuming you're referring to why I moved this logic to a new function called `injectComponent`?
There is a lot of specifics involved but the general explanation is that I did it this way because the alternative is something along the lines of this:
```
function t() {
  const result = []
  
  while ((match = linkRegex.exec(part))) {
    // do link related stuff, add to result while preserving original structure of the translation
  })

  while ((match = tooltipRegex.exec(part))) {
    // do tooltip related stuff, add to result while preserving original structure of the translation
  })
  
  // trickery to get all of it into a single array, which React knows how to render.

  return result
}
```

In this approach a lot of code will have to be repeated inside the while loop. I'm assuming we'll add more types besides the `Tooltip` and `Link` down the road, so instead of having to define a while loop for every new type, I chose to create a helper function (`injectComponent`) which should make it easier to add new ones.

I'll admit this is not the cleanest code but I plan on improving this a bit more later on as the requirements become clearer.

ecurrencyhodler:
I kinda like this article: https://bitcoinmagazine.com/culture/ten-years-later-reflection-bitcoins-genesis-and-satoshis-timing

Tried to test but for some reason I'm failing the challenge. Am i inputting the value incorrectly? I tried several ways. 
Screenshot 2023-02-24 at 9 44 29 AM

afterburn:
> I kinda like this article: https://bitcoinmagazine.com/culture/ten-years-later-reflection-bitcoins-genesis-and-satoshis-timing
> 
> Tried to test but for some reason I'm failing the challenge. Am i inputting the value incorrectly? I tried several ways. 
> Screenshot 2023-02-24 at 9 44 29 AM
> 
> 

Your input should be `echo $scriptSigHex | xxd -r -p` now, but I feel both should pass the check, feel free to create an issue for it.

As for the link you sent, I'll swap out the image for the link later today. 😁

ecurrencyhodler:
Hmm that isn't working either. I'll create an issue.
Screenshot 2023-02-24 at 10 05 02 AM

ecurrencyhodler:
I figured it out. When I copied the op_return from blockstream.info it also copied over an extra space or character(?) at the front. However it's not visible when I paste it into another text tool. But when I copy it from the text tool to challenge it works.

Also the pt. 1 of the challenge accepts it but it fails in pt. 2. Will include this in issue report.

ecurrencyhodler:
Okay just kidding I was testing the wrong challenge but still found a bug. 

Your PR works great. Thanks kevin.

Screenshot 2023-02-24 at 10 13 29 AM

ecurrencyhodler:
Thanks for the pr and including the link! One small requested change.

Can we swap the text to say:

Interesting! The message references a news article on the same day Satoshi mined the genesis block. This gives us an important clue about Satoshi's motivation for creating bitcoin. You can read more about it [here](https://bitcoinmagazine.com/culture/ten-years-later-reflection-bitcoins-genesis-and-satoshis-timing).

afterburn:
> Thanks for the pr and including the link! One small requested change.
> 
> Can we swap the text to say:
> 
> Interesting! The message references a news article on the same day Satoshi mined the genesis block. This gives us an important clue about Satoshi's motivation for creating bitcoin. You can read more about it [here](https://bitcoinmagazine.com/culture/ten-years-later-reflection-bitcoins-genesis-and-satoshis-timing).

Hey @ecurrencyhodler, made the requested change. :grin:

ecurrencyhodler:
Amazing ty! Tested and it looks good.

ecurrencyhodler:
@shaavan seems to be 1 unresolved convo on your end. Can you do a quick code review too? Then we can merge.

lib/translate.tsx
shaavan: Thanks for the explanation, @afterburn.

I was thinking that calling the `injectComponent` function might become repetitive as the number of `Component.Type` increases. But then we can move repetitive calling under a simple _for loop_ to save us some steps.

PR #198, opened 2023-02-24
Sanitize input before checking if valid for Terminal challenges

### LINKED ISSUE #197: Ch.1 Challenge 2 Pt. 2 copy/paste bug
**Describe the bug**
When I copy and paste the value from the explorer to the challenge, it passes Ch.1 Challenge 2 Pt 1 but when I paste the same value into pt. 2 it fails. 

I think what's happening is that when I copy the op_return from the explorer, it brings over an extra character/space. The thing is it passes pt. 1 but not pt. 2. But when I then paste the op_return into a text tool and copy it over to the challenge, it works.

This might require some deeper investigation but I'm going to attach a screen recording of what I did. I think the ideal solution is that pt. 2 would accept the value just like it does in pt. 1.

afterburn:
## Summary
This PR should close #197. Added some logic that removes `\t`, `\r` and `\n` from the input before it compares it with the `expectedInput`.

## Checklist
- [x] Clean commit history

### DIFF
--- a/ui/lesson/TerminalChallenge.tsx
+++ b/ui/lesson/TerminalChallenge.tsx
@@ -73,29 +73,31 @@ export default function TerminalChallenge({
   }
 
   const onChange = (input) => {
-    setLines((lines) => [...lines, { value: input, type: 'input' }])
+    const sanitizedInput = input.replace(/[\t\r\n]/g, '')
+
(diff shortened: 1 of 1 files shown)


ecurrencyhodler:
You're amazing. Thanks for getting on this so quickly! Tested and it works. 💪

PR #199, opened 2023-02-25
Homepage layout

### LINKED ISSUE #174: Polish home page layout
This is primarily on issue on mobile, where the minor inconsistencies end up making the layout pretty messy. It's just a few details on padding, margins, line-height, etc. Below you see the live site on the left, and design on the right.

image

image

benalleng:
Closes #174, #222

Brings the homepage closer to the figma for both desktop and mobile
I tried to match the [figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=161%3A1652&t=vcFdzcmWXtp6aJ7K-0) as close as i could, please point out anything i missed!

I have a few optional styles at the moment for the different homepage looks!
[Preview 1](https://saving-satoshi-r3v1op6fb-saving-satoshi.vercel.app/en)
[Preview 2](https://saving-satoshi-ahxvlc6x3-saving-satoshi.vercel.app)
[Preview 3](https://saving-satoshi-t7737q076-savingsatoshi.vercel.app)

[Up-to-date Preview](https://saving-satoshi-git-fork-benalleng-homepagemobile-savingsatoshi.vercel.app/)

I encourage you to watch my progress and use this up to date preview to test on [PageSpeed Insights](https://pagespeed.web.dev/report?url=https%3A%2F%2Fsaving-satoshi-git-fork-benalleng-homepagemobile-savingsatoshi.vercel.app%2Fen&form_factor=mobile). 
Compared to [savingsatoshi.com](https://pagespeed.web.dev/report?url=https%3A%2F%2Fsavingsatoshi.com%2Fen&form_factor=mobile)


components/Hero.tsx
```diff
@@ -9,19 +9,21 @@ export default function Hero() {
 
   return (
     <div className="homepage-img -mt-20 flex grow items-end justify-center bg-cover bg-fixed bg-[-10em] md:bg-[0]">
-      <div className="flex w-screen flex-col justify-center bg-gradient-to-b from-back/0 to-back/100 p-10 font-cbrush text-white">
-        <h1 className="text-center text-6xl sm:text-7xl lg:text-8xl">
+      <div className="flex w-screen flex-col justify-center bg-gradient-to-b from-transparent via-[#2a3b61d5] to-[#2A3B61] px-5 pt-5 font-cbrush text-white sm:p-10">
```
benalleng, line 12: This is the color matched to the figma but not the same as the background used on the rest of the site

afterburn: @GBKS is this color a one off or do we need to use this throughout the app? If its a one-off you can resolve this comment @benalleng! 😁

adamjonas:
Preview 3's hero image and the header text blend making it unreadable. The other two look good!


Screen Shot 2023-02-27 at 8 31 45 PM

GBKS:
Awesome! It's going to look so good.

A couple of thoughts on how to improve this overall:
- Let's ensure we use JPGs for these to get better compression (PNG is good for images with lots of flat color)
- Let's use different images for desktop and mobile (should be doable via `srcset`, [example](https://github.com/BitcoinDesign/Guide/blob/master/_includes/picture.html))
- Let's have regular and retina images for both desktop and mobile (so a total of 4 images, also via `srcset`)
- This will allow us to optimize image layout/cropping and download size
- We may need to introduce a light gradient at the top and bottom to slightly dim the background to provide better contrast for the text content. I can mock something up. Usually something subtle works well enough, so subtle you don't even notice it's there

afterburn:
> Awesome! It's going to look so good.
> 
> A couple of thoughts on how to improve this overall:
> 
> * Let's ensure we use JPGs for these to get better compression (PNG is good for images with lots of flat color)
> * Let's use different images for desktop and mobile (should be doable via `srcset`, [example](https://github.com/BitcoinDesign/Guide/blob/master/_includes/picture.html))
> * Let's have regular and retina images for both desktop and mobile (so a total of 4 images, also via `srcset`)
> * This will allow us to optimize image layout/cropping and download size
> * We may need to introduce a light gradient at the top and bottom to slightly dim the background to provide better contrast for the text content. I can mock something up. Usually something subtle works well enough, so subtle you don't even notice it's there

Let's actually use [WebP](https://developers.google.com/speed/webp) to save on bytes sent over the network. We're mostly using Next's [Image](https://nextjs.org/docs/api-reference/next/image) component for our images. When using this component you can just pass it a jpg/png image and it will automatically convert it to WebP AFAIK (you can see this in action for the Chapter images on `/chapters`).

Currently the only non-webp images we have are the ones that we define via CSS directly, the best solution would be to avoid doing that, but if we cannot, let's at least make sure to manually convert them to webp and use those in the css definition.

components/Hero.tsx
ecurrencyhodler: @GBKS Just want to make sure this comment doesn't fall through the cracks! Mind taking a look when you get a chance?

benalleng: This pertains specifically to [Preview 1](https://saving-satoshi-r3v1op6fb-saving-satoshi.vercel.app/en)

vinayaksh42:
the `tell me more` button is getting hidden behind the `footer` - `An open-source...` in mobile view.

**Screenshot**
Screenshot 2023-03-03 at 10 06 02 PM

benalleng:
image

image

Bam!!

@shaavan @vinayaksh42 I went ahead and reverted the changes that broke the footer if you want to re-review.

@adamjonas readability of the topbar has been fixed by adding a bg-gradient. It's pretty weak at the moment. Let me know if you think it's sufficient.

@GBKS next/image actually natively has a srcset of ~8 different images.
```
srcset="/_next/image?url=%2Fassets%2Fimages%2Fmain-image.png&w=640&q=100 640w, 
/_next/image?url=%2Fassets%2Fimages%2Fmain-image.png&w=750&q=100 750w, 
/_next/image?url=%2Fassets%2Fimages%2Fmain- image.png&w=828&q=100 828w, 
/_next/image?url=%2Fassets%2Fimages%2Fmain-image.png&w=1080&q=100 1080w, 
/_next/image?url=%2Fassets%2Fimages%2Fmain-image.png&w=1200&q=100 1200w, 
/_next/image?url=%2Fassets%2Fimages%2Fmain-image.png&w=1920&q=100 1920w, 
/_next/image?url=%2Fassets%2Fimages%2Fmain-image.png&w=2048&q=100 2048w, 
/_next/image?url=%2Fassets%2Fimages%2Fmain-image.png&w=3840&q=100 3840w" 
```
@afterburn next/image also automatically turns the image into .webp format as can be seen here.

image

@ecurrencyhodler working to enhance our PageSpeed Insights also got me to thinking about our google search results and I found this!

image

We obviously need this to be more exciting! I'll go ahead and open an issue for us all to have some open feedback for!

ecurrencyhodler:
Wow great work! Our stats look awesome. 

Fun fact, I actually have a little bit of a background in SEO. But let's push all SEO-related tasks till after the demo.

afterburn (approved):
LGTM, @ecurrencyhodler feel free to merge if everything is resolved.

components/Hero.tsx
ecurrencyhodler: Hey Ben, kevin reviewed and approved. If you resolve this conversation I can merge.

adamjonas:
ACK on the mobile nav bar

styles/globals.css
```diff
@@ -25,14 +25,10 @@ html {
...
 
-.homepage-img {
-  background-image: url('/assets/images/main-image.png');
-}
-
-.ch1-background {
-  background-image: url('/assets/images/chapter-2-cover.jpg');
-  background-repeat: no-repeat;
-  background-size: cover;
+.blur {
```
GBKS, line 28: I don't see this `blur` class used anywhere in the code (did a search). Is this a leftover or did I miss it?

components/Footer.tsx
```diff
@@ -7,9 +7,9 @@ export default function Footer() {
   const lang = useLang()
   const t = useTranslations(lang)
   return (
-    <div className="h-auto w-full">
-      <div className="flex flex-col items-center justify-center px-6 py-4 text-white sm:flex-row">
-        <p className="p-1">{t('footer.paragraph_one')}</p>
+    <div className="bottom-0 h-auto w-full bg-back">
```
GBKS, line 10: I don't think you need `bottom-0`, `h-auto`, and `w-full` here. `bottom-0` has no effect since the element position is relative. `h-auto` and `w-full` are determined by the flex layout of the parent container.

Unnecessary layout classes is something I generally see a good amount in this codebase overall. Not a big issue if things look good, but also results a bit in spaghetti code. Not sure if I should point these out in general or only if they cause issues... what do you think?

components/Navbar/UserButton.tsx
```diff
@@ -51,7 +51,7 @@ export default function UserButton() {
 
   return (
     <div>
-      <div>{renderIconButton()}</div>
+      <div className="h-[43px] w-[56px]">{renderIconButton()}</div>
```
GBKS, line 54: Not a super clean solution to hardcode these pixel values here. Is it possible to make this dynamic?

content/introductions/reaping-rewards/index.tsx
```diff
@@ -15,5 +15,5 @@ export default function ReapingRewards({ lang }) {
 export const metadata = {
   title: 'reaping_rewards.title',
   subtitle: `reaping_rewards.subtitle`,
-  image: 'https://saving-satoshi.s3.amazonaws.com/ch2.png',
+  image: '/assets/images/chapter-1-cover.jpg',
```
GBKS, line 18: Preview images should be 1200x630. It's a standard across social networks.

GBKS (commented):
Awesome work here. Just had a few comments:

- The original home page image should be 2x the current size, for retina screens. Cover images for chapters should also be 2x. I'll prep those in Figma for easy export
- Preview images should be 1200x630 (a standard size across social networks). I'll also set those up
- I see various unnecessary layout classes not just in this PR but across the codebase and open PRs. Would be nice to clean those up to make the logic easier to follow and avoid potential unwanted side effects. This is more of a general comment
- You may also want to switch from grid to flex box. Avoids Safari bugs (really just following an old spec) and is more appropriate since we have just one column or row usually
- For mobile, we want to switch to using `-webkit-fill-available` instead of `h-screen` when we want the page to fill the screen vertically. This takes the mobile browser bars into consideration (independent of whether they are expanded or minimized). Might need to test this one in a separate PR

Really exciting to see this update.

GBKS:
I created a new page in Figma called "[Images for export](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=2323%3A25471&t=B0b3qToHvkRzFrti-1)". I'll set it up so you can just select an image in there and hit the export button in the sidebar and should have everything you need. Let me know if you have any preferences for file naming or otherwise. I'll keep adding to this page.

shaavan (commented):
The footer is fixed! 🚀

styles/globals.css
shaavan: Yes, this is a leftover class. It was used in `layout.tsx` to set the appropriate background properties. But this was replaced with the tailwind theme config in #146.

image

This class can be safely removed.

components/Navbar/UserButton.tsx
shaavan: I experimented with removing these classes. But I didn't observe changes in the **UserButton** size on the Desktop and Mobile view.

@benalleng, are these classes required for some other function?

components/Footer.tsx
afterburn: Feel free to point these out @GBKS, you have keener eye for these than I do. Fewer css classes ultimately means fewer bytes we have to send over the network to our clients, resulting in better pagespeed.

components/Navbar/UserButton.tsx
benalleng: Thanks for pointing this out, I had changed it to be hard-coded because there was some content shifting when loading the svg-avatar but I didn't take the time I should have to realize that was because the button itself had a `ml-4` class attached so when it loaded in it caused shifting. I went ahead and moved this margin to the parent element, replacing the hard coded pixel heights

components/Hero.tsx
GBKS: It's a one-off (sorry). In certain places it looks just tons better when colors match the images (like using a dark tone from the image instead of just black).

GBKS:
Looks awesome, I think it's good to merge?

shaavan (approved):
Looks awesome! Thanks for this excellent work, @benalleng!

benalleng:
Should be RFM!

ecurrencyhodler:
I noticed the image for the home page is .png. I believe we are predominantly working with jpegs because we have a tool that converts jpegs->webp. Do we need to swap the image out with a jpeg?

PR #210, opened 2023-03-01
New i18n Translation filesystem

### LINKED ISSUE #209: i18n folder to condense translation system
The creation of a separate i18n folder within the repository will allow easier contribution and general searching for content and copy within the project. As well it will help to minimize the amount of reused localized copy as it will be easier to find and document what already exists if it is all in one place. This will act as the one stop shop for all localization and copy for contributors to find and add to moving forward.

benalleng:
Closes #209

Adds new filesystem for all internationalization copy moving forward and a README for future translation contributors

Inspired by the work that [Jam](https://github.com/joinmarket-webui/jam) has done with their translation system. much of our README is built off of theirs. Still wondering what localization manager we should use. <del>Is [Transifex](https://www.transifex.com/) a good option or should we just stick with github for all translation updates? </del>
Github should be a fine solution for our size (of project), scope (of words), and scale (of contributors)

#### TODO:
- refactor `translations.ts` to remove unneeded locale keys


benalleng:
I was having more trouble than I'd like to admit removing the locale key in the translation objects so each translation.ts file still has the `en:{}` and `nl:{}` key for each namespace. Any reviewer that could give me some advice on refactoring the `parseTranslations()` function to not need these keys would be greatly appreciated :pray:

afterburn:
> I was having more trouble than I'd like to admit removing the locale key in the translation objects so currently each translation.ts file still has the `en:{}` and `nl:{}` key for each namespace even though they are organized into separate localized files. Any reviewer that could give me some advice on refactoring the `parseTranslations()` function to not need these keys would be greatly appreciated :pray: 

LGTM, I'll modify the parseTranslations file so we can get rid of `en` and `nl` in a later pr.

ecurrencyhodler:
Nothing visually for me to test. We need 1 code review and we can merge.

benalleng:
> I posted this comment earlier but maybe it didn't go through properly, is [Transifex](https://www.transifex.com/) a free service, could you explain a little why you think this solution is a good fit for us?

I am not necessarily sure that it is, I took inspiration from Jam and how they handle their internationalization and figured we might be able to use a similar solution but seeing that it really is an Enterprise solution with enterprise pricing I think that github is a fine solution for us.

afterburn (changes requested):
@benalleng okay makes sense, I also think using translation files + github is a good enough solution for now. I left a small comment, but afterwards we can merge.

benalleng:
The locale files themselves still need the `locale: {}` portion removed but now `translate.ts` has this logic:
```jsx
export function loadTranslations(lang) {
  const {
    translations: localeTranslations,
  } = require(`../i18n/locales/${lang}`)

  const Translations = [localeTranslations]
```

which allows us to actually add all of our locales separately

afterburn (approved):
Looks good, thanks @benalleng! Let's merge after the conflicts have been resolved.

PR #221, opened 2023-03-02
OP Codes preceding Input

benalleng:
This is a change to the input field to match the proposed [figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=1965%3A21604&t=RvOeVB4M7mvfkPfz-0) where Op_Codes have been added to begin to show the user how a Bitcoin transaction is separated.

This change only applies to the second lesson Transacting, Genesis will retain the OP_Codes in the Input for the user to more easily copy and complete.

Much of the copy remains the same, this is just a change to the input and the status bar

[Preview](https://saving-satoshi-git-fork-benalleng-inputopcodes-savingsatoshi.vercel.app/en/chapters/chapter-1/transacting-2) (Input starts with 5, 3)

### Screenshots
image

image

image


ecurrencyhodler:
I kinda dig the color scheme.

content/lessons/transacting/transacting-3/index.tsx
```diff
@@ -6,7 +6,7 @@ export default function Transacting3({ lang }) {
   const t = useTranslations(lang)
   return (
     <TerminalChallenge
-      expectedInput={'6a127461636f7320666f722065766572796f6e65'}
+      expectedInput={'7461636f7320666f722065766572796f6e65'}
```
satsie, line 9:
```suggestion
      expectedInput={'5361746f7368693433335061726b53744368696e6f43413933323433'}
```

Is there any way to take this value out and make it a constant somewhere? Needing to update it in two places doesn't leave me with a high level of confidence that I haven't missed another occurrence of it.

styles/globals.css
```diff
@@ -6,6 +6,7 @@
   --back: 50, 71, 96; /* Theme background color (#324760) */
   --green: 40, 177, 35; /* Success color (#28B123) */
   --terminal-output: #3DCFEF;
+  --incorrect-answer: #EF960B;
```
satsie, line 9: Yay thanks for adding this! @GBKS if you have any opinions on the color to use for incorrect answers (at least for the demo), speak now or forever hold your peace (until v1)

Or @sabakhilji! Are there any good accessibility choices here? Right now we're using orange, a color I arbitrarily picked because red felt a little too scary, harsh, and punishing.

satsie (commented):
Thank you for tackling this!!! <3 I can't comment on the correctness of the React code, but the screenshots look awesome and appreciate the details you've incorporated like the incorrect error coloring and the in progress status message.

We have a new set of screens that Andrew is working on to make the story more cohesive. This impacts the transaction used for challenge 2, so I have have left suggestions for places that will need updating.

Also left a comment about if we want to pre-fill the opcodes in for the user in challenge 1. Aside from what I wrote, I'm on the fence because I have frequently seen the entire scriptSig decoded, like on t-shirts, mugs, and images on the internet. That garbage you get at the beginning (from converting the opcodes to ASCII) is almost culturally significant at this point. It also let's the user have a quick win (albeit a slightly messy one) before challenge 2 when we try to gently point out we don't actually want the opcodes. 

Let me know your thoughts on this, and hopefully we can get this merged upstream within the first half of the week. Thanks so much Ben!

satsie:
Ok I just tested the branch and LOVE what you have done to highlight just the wrong characters, like in this screeshot. That is really cool!

Screenshot from 2023-03-06 10-13-45

I think you've also been working on the input box and I gotta say, the experience is SO much smoother. It's nice that I can click anywhere on the empty challenge, and it moves my cursor to the beginning, and how I can select all with a hotkey and change my answer really quickly. Well done!

styles/globals.css
sabakhilji: Love the idea of keeping correct things in white color and I love this orange color. My only concern is about the contrast ratio which might create an accessibility issue. Will dig more into the color scheme to make it more accessible.

content/lessons/genesis/genesis-1/index.tsx
```diff
@@ -8,11 +8,17 @@ export default function Genesis1({ lang }) {
...
+        '5468652054696d65732030332f4a616e2f32303039204368616e63656c6c6f72206f6e206272696e6b206f66207365636f6e64206261696c6f757420666f722062616e6b73'
       }
       next={'/chapters/chapter-1/genesis-2'}
       label={t('genesis_one.placeholder')}
       pattern={/[a-z0-9]+/gi}
+      opcode={
+        <>
+          OP_PUSHBYTES_4 ffff001d <br className="hidden md:inline" />{' '}
+          OP_PUSHBYTES_1 04 OP_PUSHBYTES_69
+        </>
```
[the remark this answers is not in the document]
benalleng: Shouldn't be hard at all to have one the standard way and the second challenge how this PR lays it out

content/lessons/transacting/transacting-3/index.tsx
benalleng: I could use the first content page of the first challenge to be passed as a prop to `/content/lessons/transacting/transacting-3/index.tsx`

styles/globals.css
benalleng: We can change it to whatever we decide is best, I think a combination of accessibility and user feedback is the way to go here

GBKS: If we're looking into accessibility and contrast, let's please use WCAG 3 standards and not the outdated WCAG 2.1. Most color contrast checking tools use an old model from 2008 based on simple color value math that does not take into consideration how humans perceive color or that screens and accessibility features have improved since then. Thanks.

satsie: Everyone okay with moving forward with this orange then following up with a later PR once more research is done?

GBKS: Yep.

content/lessons/transacting/transacting-3/index.tsx
benalleng: Looking more into this, This is a one off because we are having both challenges in Transacting-2 and Transacting-3 use the same answer. But I do not think this will occur enough that we will need to create a template for this particular case

benalleng:
Marking as draft until I can get it to deploy properly

ui/common/Input.tsx
```diff
@@ -7,17 +7,34 @@ interface UserInputProps {
...
   answer,
   pattern,
   hints,
+  opcode,
 }: UserInputProps) {
   const [textAreaValue, setTextAreaValue] = useState('')
   const [correctAnswer, setCorrectAnswer] = useState(false)
 
+  const opCodeOverlay = (opcode) => {
+    let OPCodeColor =
```
afterburn, line 24: This can be a const

ui/common/StatusBar.tsx
```diff
@@ -14,25 +15,34 @@ export default function StatusBar({
...
   errorMsg?: string
   full?: boolean
+  hints?: boolean
 }) {
   const { activeView } = useLessonContext()
   const isActive = activeView === LessonView.Code
 
   const getStatus = () => {
-    if (!input) return Status.InProgress
+    if (!input) return Status.Begin
```
afterburn, line 39: Let's not use naked if's, they make the code less readable imo. I prefer something like this:
```ts
const getStatus = () => {
  if (!input) {
    return Status.Begin;
  }

  if (input === expected) {
    return Status.Success;
  }

  if (
    hints &&
    (!input || (expected.startsWith(input) && expected[0] === input[0]))
  ) {
    return Status.InProgress;
  }

  return Status.Error;
};

```

ui/common/StatusBar.tsx
```diff
@@ -14,25 +15,34 @@ export default function StatusBar({
...
   successMsg,
   inProgressMsg,
   errorMsg,
   full,
+  hints,
 }: {
   next: string
   input: string
   expected: string
+  beginMsg?: string
```
afterburn, line 28: Can we change these to `beginMessage`, `successMessage`, etc? Let try to stay away from abbreviations at least for prop names.

ui/common/Input.tsx
```diff
@@ -93,28 +110,38 @@ export default function Input({
...
+          onChange={handleChange}
+          value={textAreaValue}
+          spellCheck="false"
+          className={`absolute top-0 left-0 h-full w-full resize-none overflow-hidden break-all bg-transparent font-space-mono text-[18px] leading-[180%] tracking-[1px] text-transparent outline-none md:text-[30px] md:tracking-[5px]`}
+          style={{
+            caretColor: '#6e7d92',
+          }}
+        />
+        <p
+          className={`${
```
afterburn, line 138: Could you use `clsx` instead of string interpolation?

ui/common/Input.tsx
```diff
@@ -93,28 +110,38 @@ export default function Input({
...
-      />
+    <>
       <p
         className={`${
           correctAnswer ? 'overlay-complete' : 'overlay-incomplete'
-        } pointer-events-none h-full w-full break-all font-space-mono text-[18px] leading-[180%] tracking-[1px] text-inherit md:text-[30px] md:tracking-[5px]`}
-        style={{
-          lineBreak: 'anywhere',
-        }}
+        } pointer-events-none h-full w-full break-all text-left font-space-mono text-[18px] leading-[180%] tracking-[1px] md:text-center md:text-[30px] md:tracking-[5px]`}
```
afterburn, line 117: Use `clsx` please.

content/lessons/transacting/transacting-3/index.tsx
```diff
@@ -6,9 +6,7 @@ export default function Transacting3({ lang }) {
   const t = useTranslations(lang)
   return (
     <TerminalChallenge
-      expectedInput={
-        '6a1c5361746f7368693433335061726b53744368696e6f43413933323433'
-      }
+      expectedInput={'5361746f7368693433335061726b53744368696e6f43413933323433'}
```
afterburn, line 9: No curly braces needed here either.

ui/common/StatusBar.tsx
```diff
@@ -77,7 +89,10 @@ export default function StatusBar({
             className={clsx(
               'font-nunito text-[21px] text-white transition duration-150 ease-in-out',
               {
-                'opacity-50': getStatus() === Status.InProgress,
+                'opacity-50': getStatus() === Status.Begin,
```
afterburn, line 92: You can just use a single object here:
```jsx
clsx("font-nunito text-[21px] text-white transition duration-150 ease-in-out", {
  "opacity-50": getStatus() === Status.Begin,
  "text-[#EF960B]": getStatus() === Status.Error,
})
```

ui/lesson/InputChallenge.tsx
```diff
@@ -1,6 +1,6 @@
 'use client'
 
-import { useState } from 'react'
+import { ReactElement, useState } from 'react'
```
afterburn, line 3: You import `ReactElement` here but I don't see it used anywhere.

ui/lesson/Prompt.tsx
```diff
@@ -1,4 +1,5 @@
 import clsx from 'clsx'
+import { ReactElement } from 'react'
```
afterburn, line 2: Same here

afterburn (changes requested):
Hey @benalleng, thanks. I left some comments.

ecurrencyhodler:
Tested. Ben you might need to rebase. Your PR in challenge for pt. 2 accepts I think the old one "Taco's are for everyone". Other than that it looks great.

ecurrencyhodler:
Oh I see what the issue is. 

To me, pt.2(decoding address) of this challenge makes it seem like the op_return from pt.1 should be re-used in pt.2 but that's not the case. That's why I assumed the op_return from pt.2 was the answer to pt.1

I pinged @satsie on this in figma. 

This is a tACK for me and I'll see if we need to do a fast follow based on what stacie says and create another issue.

benalleng:
I think it should be, 
the answer in Challenge-2 pt-1
answer
```
5361746f7368693433335061726b53744368696e6f43413933323433
```
and Challenge-2 pt-2
answer
```
echo 5361746f7368693433335061726b53744368696e6f43413933323433 | xxd -r -p
```

satsie:
I just ran through the flow in the Vercel preview and everything works for me (at least the happy path, I didn't try to break anything!)

Challenge 1 and Challenge 2 use two different transaction IDs and are unrelated. @ecurrencyhodler can you provide a screenshot of where you're seeing the "tacos are for everyone" message/where that answer is being accepted?

ecurrencyhodler:
@satsie the issue was pt1 and pt.2 use different op_returns but pt.2 makes it seem the user is supposed to use the same one. I left a comment in figma. I think we need to change copy for pt2 to clearly look for a diff op_return.

satsie:
@ecurrencyhodler moving this convo to Figma :)

ecurrencyhodler:
Okay the copy isn't a blocker then. We can follow up later with other issues. Let's keep moving forward.

satsie:
aw shoot! can't merge until @afterburn approves

afterburn (approved):

PR #232, opened 2023-03-03
Error fix for terminal based challenges

### LINKED ISSUE #225: Styling Error in Terminal based challenges in Mobile view
**Describe the bug**
The `info` and `code` buttons are larger than required. Upon switching to the `code` tab the content from the `info` tab continues to display. Error on [chapter1, Challenge2 (Transacting)](https://savingsatoshi.com/en/chapters/chapter-1/transacting-3) and [chapter1, Challenge1 (Genesis)](https://savingsatoshi.com/en/chapters/chapter-1/genesis-2)

**To Reproduce**

1. Go to https://savingsatoshi.com/en/chapters/chapter-1/transacting-3 or https://savingsatoshi.com/en/chapters/chapter-1/genesis-2
2. Click on the `code` tab
3. See the error

**Expected behavior**
The height of the tab should be `32px` ([figma link](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=2019%3A30306&t=I3d3TBGp2xKaD4A7-0)) and the content from the `info` tab should not display in `code` tab.

Tasks:
- [ ] The info and code tabs are larger than they should be
- [ ] In the code section, the text from the info tab persists inside the code tab when it shouldn’t.

**Screenshots**
Screenshot 2023-03-03 at 10 08 28 PM
Screenshot 2023-03-03 at 10 08 36 PM

**Additional context**
Upon a quick look at the deployment history, the bug seems to have been introduced after merging PR #146 - [deployment link](https://saving-satoshi-np4s3cskq-savingsatoshi.vercel.app/en/chapters/chapter-1/genesis-2)

conversation related to this bug - https://github.com/saving-satoshi/saving-satoshi/pull/146#issuecomment-1451900563

vinayaksh42:
Fixes #225 

Removing the use of the `useMediaQuery` hook from the `terminalChallenge` component.


ecurrencyhodler:
tACK!
image
image

vinayaksh42:
I have removed the https://github.com/saving-satoshi/saving-satoshi/blob/master/hooks/useMediaQuery.ts file as it is not being utilized and `react-responsive` dependency.

ui/lesson/TerminalChallenge.tsx
```diff
@@ -62,8 +62,25 @@ export default function TerminalChallenge({
           },
         ]
   )
+  const [windowWidth, setWindowWidth] = useState(window.innerWidth)
 
-  const isSmallScreen = useMediaQuery({ query: '(max-width: 767px)' })
+  useEffect(() => {
```
afterburn, line 67: I think I prefer keeping the `useMediaQuery` hook rather than moving the resizing logic into a specific component like `TerminalChallenge`. If, in the future, we have to do anything related to resizing the window, we would have to reimplement the same logic again using your approach. However, if it were just a hook, we could simply import and use it.

afterburn (changes requested):
Hey @vinayaksh42, I left a comment.

ui/lesson/TerminalChallenge.tsx
vinayaksh42: I agree we should implement it as a reusable hook. The current `useMediaQuery` hook is not working as we intend it to work, I will try to make a new implementation for it.

ui/lesson/TerminalChallenge.tsx
```diff
@@ -3,9 +3,9 @@
 import { useState, useEffect } from 'react'
 import { LessonDirection } from 'types'
 import { Lesson, LessonTabs, LessonTerminal } from 'ui'
-import { useMediaQuery } from 'hooks'
 import { setUserProgress } from 'lib/user'
 import { useStatus } from 'hooks'
```
afterburn, line 7: Nitpick but here you should be able to do:
```jsx
import { useStatus, useMediaQuery } from 'hooks'
```

afterburn (approved):
@vinayaksh42 thanks for making the change. I left a small nitpicky comment, but I think its okay to merge without resolving it @ecurrencyhodler.

ui/lesson/TerminalChallenge.tsx
vinayaksh42: Thanks for the suggestion, I have changed it

Issue #214, opened 2023-03-01
Website size on Safari is not correct. Creates scroll bar.


ecurrencyhodler:
**Describe the bug**
@vinayaksh42 raised this issue on a call with me as we were troubleshooting a PR. 

In Safari, the website size does not match the initial frame of the website when it initially loads which creates a scroll bar on the right.

<img width="1257" alt="Screenshot 2023-03-01 at 3 01 44 PM" src="https://user-images.githubusercontent.com/32662508/222287358-ab2247a0-3204-48be-9bc6-2efd7fd65b66.png">

This combined with Shashwat's issue in this PR https://github.com/saving-satoshi/saving-satoshi/pull/146#issuecomment-1441809114 with safari as well, I'm going to open up a discussion for us to see if we want to continue to support safari as a browser since it has a tendency to have lots of bugs.

ecurrencyhodler:
Closing this issue based on this discussion. https://github.com/saving-satoshi/saving-satoshi/discussions/216

ecurrencyhodler:
Closed by #236

PR #236, opened 2023-03-06
Fixes a few layout issues in lesson screens

GBKS:
Replacing grid with flex layout to fix a few issues. Also removing some unnecessary Tailwind tags.

Should fix issue #214 and the problem pointed out in this comment: https://github.com/saving-satoshi/saving-satoshi/pull/146#issuecomment-1450874989

I'd recommend avoiding grid in favor of flex box. I only saw it in a few places in the codebase and on those a grid wasn't even used (meaning multiple rows and columns), but it's either a row a column. Safari follows an older spec in regards to height calculation, which caused one of the layout issue mentioned above.

Overall I also see a lot of unnecessary layout tags that have no effect. I wonder if everyone is clear on exactly how flexbox works. If not, I recommend [this guide](https://css-tricks.com/snippets/css/a-guide-to-flexbox/).

**[Live site](https://savingsatoshi.com/en/chapters/chapter-1/genesis-2):**
image

**[Preview](https://saving-satoshi-git-feature-fix-lesson-layo-b18562-savingsatoshi.vercel.app/en/chapters/chapter-1/genesis-2):**
image

**Live site:**
image

**Preview:**
image

### DIFF
--- a/app/[lang]/chapters/[slug]/[lesson]/layout.tsx
+++ b/app/[lang]/chapters/[slug]/[lesson]/layout.tsx
@@ -1,4 +1,4 @@
 import { chapters } from 'content'
 export default function Layout({ children }) {
-  return <div className="flex h-full grow flex-col">{children}</div>
+  return <div className="flex grow flex-col">{children}</div>
 }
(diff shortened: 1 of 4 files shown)


afterburn (approved):
@GBKS looks good. +1 for always using flexbox unless we specifically need some grid features.

ecurrencyhodler:
tACK!

Issue #189, opened 2023-02-23
Remove /en ending for website


ecurrencyhodler:
**Describe the enhancement**
We should remove the /en in our url: https://savingsatoshi.com/en

**Additional information**
We don't currently support another language at this time.

afterburn:
This is not very easy to do. We could spend time implementing our own fix for this, but I suggest we keep using `/en` until the official i18n solution for Next 13 is out of beta. I definitely don't think this is important for the demo.

ecurrencyhodler:
@adamjonas How do you feel about punting this to a later date?

afterburn:
I think the first mistake was using Next 13 (which is still in beta) with the app directory approach. This has caused some headaches with the translation system in particular, but I feel we've gotten the system to a point now where its workable for a while.

I actually spoke with someone from the Next.js team and they haven't let anything on about whether or not they are even considering addressing the i18n issues some have mentioned. I'm assuming they will though because it doesnt work very well in its current state. So to answer your question about when it's coming out of beta, I have no clue.

I think I can take a stab at making 'en' the default locale so that if its not present in the url it defaults to 'en' anyway. I think that could work.

benalleng:
@afterburn I can go ahead and add that default localization to my PR #169

afterburn:
> @afterburn I can go ahead and add that default localization to my PR #169 

Hey @benalleng , sure thanks. You will have to make some changes in `middleware.ts`, that's where we redirect the user to `/<locale>`. Presumably you'll also have to modify `lib/translate.ts` so it defaults to `en` if no lang param is passed.

afterburn:
Sorry @benalleng I forgot to mention this in my previous comment, since we're using Next 13's app directory approach you might have some difficulty getting the correct page to show up when navigating to `/` since the index page is now under `app/[lang]/page.tsx` as opposed to `/app/page.tsx`. I found the Next 13 docs very lacking on how this should be solved idiomatically within the framework so feel free to do something you think makes sense (and is easiest).

adamjonas:
@ecurrencyhodler 
It really bothers me but shouldn't stand in the way of forward progress or be a blocker of the demo milestone.

ecurrencyhodler:
Briefly sync'd with Ben. It seems like the path to fix this is a bit unclear. I think it's best for us to go through the "to do" list first on our project board and then circle back. So I'm marking this for v1.

ecurrencyhodler:
[...]

PR #245, opened 2023-03-07
Fix translation system

### LINKED ISSUE #243: Translation Fallbacks
At the moment the current localization system `/lib/translate.ts` does not properly grab fallback translations when a specific translation is missing. It seems t gather some translations but not all of them. This also happens when the localization switches without updating the language loaded for the page which happens whenever a link is clicked.

### LINKED ISSUE #244: Fix translation system
According to @benalleng and me there are currently some issues with the translation system:
1. Hydration issue when the user switches to a locale other than `en`. I was able to reproduce this issue, but not consistently.
2. `i18n/locales/<locale>/translations.ts` requires you to specify a locale object, since we already handle this in the folder structure this shouldn't be needed.
3. There is some issue with fallbacks (see #243), TLDR I think we should just return the translation key if a translation is missing for a specific locale for now, that way they'll be easy to identify and fix. Later on, we can look into proper error logging and providing fallbacks.
4. `lib/translate.ts` is overly complex, after @benalleng changes this file can be simplified/cleaned up.

afterburn:
This PR will at some point close #243 and #244, it:

- Simplifies how `lib/translate.ts` works.
- Removes the need for specifying the locale in `i18n/locales/<locale>/translations.ts` files.
- Makes sure a translation key is returned if a translation is missing, so we can easily identify what needs work.
- Fixes hydration issues.

I'm drafting a PR so I can test on the Vercel deployment.


benalleng:
Wow! I'll make sure @vinayaksh42 also sees this as he and I spoke some about this.

afterburn:
Cool! It'd be great if @vinayaksh42 can focus on removing the need for the `/en` if the user locale is set to english, so #189. I will be handling other translation issues, updated my PR description to reflect this.

PR #251, opened 2023-03-08
Fix missing translations on Homepage for english locale

### LINKED ISSUE #249: Fix missing translation for Homepage with english locale active
When users have an english locale active and they navigate to savingsatoshi.com, the system will not redirect users to `/en` as per #189. This however means that the translation system does not know what translations to load resulting in the following:

image

For users with english locales we should make sure the translation system will fall back to english (at least for the homepage) for now.

afterburn:
This PR fixes #249.

### DIFF
--- a/hooks/useLang.ts
+++ b/hooks/useLang.ts
@@ -1,8 +1,17 @@
 import { usePathname } from 'next/navigation'
 
+import { i18n } from 'i18n/config'
+
 export const useLang = () => {
   const pathName = usePathname()
+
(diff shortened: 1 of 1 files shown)


benalleng:
tACK

vinayaksh42:
The homepage looks perfect now - https://saving-satoshi-git-fix-249-savingsatoshi.vercel.app/
Screenshot 2023-03-09 at 2 57 33 AM

PR #254, opened 2023-03-08
Fix: 404 for favicon and android-icon

### LINKED ISSUE #222: favicon 404 on firefox desktop
I am getting a 404 error on firefox when it attempts to load the icon.

I think this may have to do with the order that it loads the website, it first does a redirect to `/en` then attempts to load the icon from the URL `/en/favicon.png` causing a 404.
image

image

image

I think this could be fixed by making sure the favicon.ico loads which at the moment isn't present in the networking tab.
I can go ahead and add this to #199

vinayaksh42:
Fixes #222 (testing)

In firefox favicon and android-icon are being redirected to `/en

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -45,6 +45,6 @@ export function middleware(request: NextRequest) {
 export const config = {
   // TODO: Find a way to handle these dynamically
   matcher: [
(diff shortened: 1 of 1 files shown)


benalleng (approved):
LGTM, from my previous comment about not consistently reproducing it. I think my console just had some old errors when I tested it the first time

PR #263, opened 2023-03-09
Copy updates on various pages

GBKS:
These are updates to all existing chapter 1 pages to match the latest Figma content, as well as some style changes for consistency. Leaving this as a draft to resolve a few issues:

- Hard-coded intro and subtitle on the chapter 1 intro that we don't have copy for. Adjust the template to remove, or leave it in and hide elements when no translation value is proivded in case other pages need it? (See an example in the [chapters overview page](https://saving-satoshi-git-feature-copy-revision-savingsatoshi.vercel.app/en/chapters), and [chapter 1 intro page](https://saving-satoshi-git-feature-copy-revision-savingsatoshi.vercel.app/en/chapters/chapter-1).
- I don't think the link I added to the Times articles ([link](https://www.thetimes03jan2009.com/) in the success message [here](https://saving-satoshi-git-feature-copy-revision-savingsatoshi.vercel.app/en/chapters/chapter-1/genesis-2)) is the best, just did a quick search. Let's find a better one.
- Will tweak the images a little further (put an actual countdown on the big screen in the chapter 1 cover, try to make others more crisp and some other minor tweaks)

And a Github issue I am getting. When I try to create a branch and commit some code, I get an unhelpful error message about something related to husky. From a search, this looks like something about my code is not well formatted. No idea what though. Running the lint/prettier/test scripts does not get rid of this. So I can only commit my branch by doing a "--no-verify" commit to establish the branch. Anyone know what this might be about?


i18n/locales/en/translations.ts
```diff
@@ -67,69 +67,71 @@ export const translations = {
   },
 
   reaping_rewards: {
-    title: 'Of blocks and miners',
-    subtitle:
-      '"To understand the mystery and magic, you must start at the beginning…"',
-    paragraph_one: `Satoshi, the pseudonymous inventor of bitcoin, also mined the first bitcoin block. He left you a hidden clue in the very first bitcoin transaction that was ever made.`,
-    paragraph_two: `Each block in the bitcoin blockchain contains a set of transactions. A nice way to view these blocks and the things they contain is by using a block explorer.`,
+    title: null,
```
afterburn, line 70: Not sure if this is final since this is a Draft pr, but let's remove these both from this translation file and the actual component if they are not being used, otherwise this happens:
image

i18n/locales/en/translations.ts
```diff
@@ -5,15 +5,15 @@ export const translations = {
...
   },
 
   chapter_one: {
     title: 'Secrets in plain sight',
     paragraph_one:
-      'The year is 2139. The very last bitcoin is slated to be mined in two weeks. For months, a final countdown has been taking place in Times Square for the last block when there will be officially no more bitcoin issued.',
-    paragraph_two: `And then suddenly, the network comes to a crawl. Instead of the normal ten-minute block interval, blocks are coming in just a few times a day, if not longer. Your mom's bitcoin node stops syncing. The media starts reporting on the situation. It's everywhere on social media.`,
-    paragraph_three: `The price of bitcoin is falling. There is a knock at your door, and when you open it, a yellowed, age-weathered envelope has been left for you. There's no name or return address. What could be inside?`,
+      'The year is 2139. The last bitcoin is two weeks from being mined. For months, a clock has ticked down in Satoshi Square. The world awaits the last block. Then, suddenly, the network grinds to a halt.',
+      paragraph_two: null,
```
afterburn, line 15: Same here:

image

i18n/locales/en/translations.ts
GBKS: Yes, that was my question in the top comment. We may need some flexibility here when adding title and subtitle. Some pages might need it, others don't. Is there a clean way to make this work? Like setting those translation values to null or so?

content/introductions/reaping-rewards/index.tsx
```diff
@@ -6,14 +6,15 @@ export default function ReapingRewards({ lang }) {
 
   return (
     <Introduction lang={lang}>
-      <Text className="pt-2">{t('reaping_rewards.paragraph_one')}</Text>
-      <Text className="pt-2">{t('reaping_rewards.paragraph_two')}</Text>
+      <Text className="">{t('reaping_rewards.paragraph_one')}</Text>
```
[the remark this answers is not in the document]
GBKS: Fixed, good catch.

ecurrencyhodler:
Hey Christoph. For the times news article, this is the link Jonas wanted: https://en.bitcoin.it/wiki/Genesis_block

i18n/locales/en/translations.ts
benalleng: I noticed that the file `Introduction.tsx` while being a template was quite limiting so I went ahead and created a draft PR to generalize the template for these copy PR's it should make it easier to reuse it

GBKS: Thank you @benalleng, I implemented a slightly modified version of your change in this PR, just so I can remove the draft status and it can be reviewed quickly.

i18n/locales/en/translations.ts
GBKS: Resolved this in this PR by adding strings with just a space character for the two translation keys (so it renders two `p` tags with just a space in each). We can find a solution to make this template more flexible later.

GBKS:
Alright, I think this one is good for review. I would still like to update some images over the next few days, but this can be a separate PR and I don't want to block any other ongoing work.

There's also a bit of a hack in the chapters overview page chapter description. Ideally we can make the template more flexible so it's not fixed to three paragraphs of text. I implemented a work-around so it looks good for now, and we can follow-up with a proper fix.

ecurrencyhodler:
Can we hyperlink "feedback" with [this link](https://docs.google.com/forms/d/e/1FAIpQLSf1xpNqUYJyvYL5IZDnxy78273pkqzfYW2Hf91H4Do4KHgy9g/viewform?usp=sf_link)?
Screenshot 2023-03-09 at 10 42 38 PM

Also do we want to include the short summary of challenges as well?
Screenshot 2023-03-09 at 10 47 05 PM

Challenge 1 
Title: Genesis
Summary: Time to practice! Decode Satoshi Nakamoto’s secret message that he left in the first bitcoin block. What was he trying to say?

Challenge 2 
Title: Transacting
Summary: It’s up to you to save Satoshi. Decode the next clue to proceed.

GBKS:
I added the link. No, to your second point about including challenge summaries. We should add the extra chapter lesson content summary right above the challenges, but I'd put that in a separate issue.

i18n/locales/en/translations.ts
```diff
@@ -5,15 +5,15 @@ export const translations = {
...
 
   chapter_one: {
     title: 'Secrets in plain sight',
     paragraph_one:
-      'The year is 2139. The very last bitcoin is slated to be mined in two weeks. For months, a final countdown has been taking place in Times Square for the last block when there will be officially no more bitcoin issued.',
-    paragraph_two: `And then suddenly, the network comes to a crawl. Instead of the normal ten-minute block interval, blocks are coming in just a few times a day, if not longer. Your mom's bitcoin node stops syncing. The media starts reporting on the situation. It's everywhere on social media.`,
-    paragraph_three: `The price of bitcoin is falling. There is a knock at your door, and when you open it, a yellowed, age-weathered envelope has been left for you. There's no name or return address. What could be inside?`,
+      'The year is 2139. The last bitcoin is two weeks from being mined. For months, a clock has ticked down in Satoshi Square. The world awaits the last block. Then, suddenly, the network grinds to a halt.',
+      paragraph_two: ' ',
+      paragraph_three: ' '
```
benalleng, line 16: Instead of these translations being blank, we should be able to remove the content in `/content/chapters/chapter-1/index.ts` to match this format
```tsx
<Chapter metadata={metadata} lang={lang}>
      <Text>{t('chapter_one.paragraph_one')}</Text>
</Chapter>
```
This will not have any implications for limiting future chapters as this is inside the chapter-1 content

i18n/locales/en/translations.ts
benalleng, line 70: With the new changes to `Introduction.ts` We should no longer need to have this translation key

content/lessons/transacting/transacting-3/index.tsx
```diff
@@ -24,19 +24,28 @@ export default function Transacting3({ lang }) {
...
-
-        <Text className="text-lg">
-          {t('transacting_three.paragraph_three')}
-        </Text>
+        <Text className="text-lg md:text-xl mt-2">{t('transacting_three.paragraph_one')}</Text>
+        <Text className="text-lg md:text-xl mt-2">{t('transacting_three.paragraph_two')}</Text>
 
         <CodeExample
+          className="mt-4"
           code={'echo $scriptPubKeyHex | xxd -r -p'}
```
benalleng, line 32: code={'echo $scriptPubKeyBytes | xxd -r -p'}

benalleng (changes requested):

benalleng:
A comment for any QA, transacting-2 will not make sense until #221 is implemented!

content/introductions/reaping-rewards/index.tsx
```diff
@@ -6,14 +6,15 @@ export default function ReapingRewards({ lang }) {
...
+      <Text className="text-lg md:text-xl">{t('reaping_rewards.paragraph_one')}</Text>
+      <Text className="text-lg md:text-xl mt-4">{t('reaping_rewards.paragraph_two')}</Text>
+      <Text className="text-lg md:text-xl mt-4">{t('reaping_rewards.paragraph_three')}</Text>
     </Introduction>
   )
 }
 
 export const metadata = {
   title: 'reaping_rewards.title',
   subtitle: `reaping_rewards.subtitle`,
```
benalleng, line 18: With these values now being not necessary to render the template they should no longer be needed in the metadata. I'm fairly sure we could always add these values through each individual index for future content

i18n/locales/en/translations.ts
```diff
@@ -67,69 +67,71 @@ export const translations = {
   },
 
   reaping_rewards: {
-    title: 'Of blocks and miners',
-    subtitle:
-      '"To understand the mystery and magic, you must start at the beginning…"',
-    paragraph_one: `Satoshi, the pseudonymous inventor of bitcoin, also mined the first bitcoin block. He left you a hidden clue in the very first bitcoin transaction that was ever made.`,
-    paragraph_two: `Each block in the bitcoin blockchain contains a set of transactions. A nice way to view these blocks and the things they contain is by using a block explorer.`,
+    title: null,
+    subtitle: null,
```
benalleng, line 71: With the title and subtitle removed and ultimately removing them from the reaping_rewards metadata, these should no longer be necessary

i18n/locales/en/translations.ts
GBKS: Fixed, thanks. I could actually not figure out where the template for this was defined until you pointed it out here.

content/lessons/transacting/transacting-3/index.tsx
GBKS: Fixed, thanks.

content/introductions/reaping-rewards/index.tsx
GBKS: Fixed, thanks.

i18n/locales/en/translations.ts
GBKS: Fixed, thanks.

benalleng (approved):
LGTM!

content/introductions/reaping-rewards/index.tsx
```diff
@@ -6,14 +6,13 @@ export default function ReapingRewards({ lang }) {
...
-      <Text className="pt-2">{t('reaping_rewards.paragraph_two')}</Text>
+      <Text className="text-lg md:text-xl">{t('reaping_rewards.paragraph_one')}</Text>
+      <Text className="text-lg md:text-xl mt-4">{t('reaping_rewards.paragraph_two')}</Text>
+      <Text className="text-lg md:text-xl mt-4">{t('reaping_rewards.paragraph_three')}</Text>
     </Introduction>
   )
 }
 
 export const metadata = {
-  title: 'reaping_rewards.title',
```
benalleng, line 16: Sorry! i got a little hasty with removing everything here and the translation, It is still needed in the alt tag for the image.

benalleng (changes requested):

content/introductions/reaping-rewards/index.tsx
benalleng: the subtitle however is not needed

GBKS: Got it, I made the change. Weird that the title is used as the alt text, maybe something we need to revise at some point.

benalleng (approved):

PR #265, opened 2023-03-09
Ensure window is defined for useMediaQuery

### LINKED ISSUE #264: window is undefined when useMediaQuery() hook runs
When loading the Transacting-Challenge-3 page an error occurs on the back-end describing an undefined error of the window.
```bash
ReferenceError: window is not defined
    at useMediaQuery (webpack-internal:///(sc_client)/./hooks/useMediaQuery.ts:9:89)
    at TerminalChallenge (webpack-internal:///(sc_client)/./ui/lesson/TerminalChallenge.tsx:52:79)
    ...
```
I believe this is due to `useMediaQuery()` trying to set the state before the window is defined here as this check is being run on the server-side.
```jsx
const [isWidthMet, setIsWidthMet] = useState<boolean>(
    window.innerWidth <= width
  )
```
One simple fix to this is to make sure this useState() only queries the window.innerWidth after it is defined as refrenced [here](https://stackoverflow.com/questions/55151041/window-is-not-defined-in-next-js-react-app) not 100% on the correctness here but i find that it does work as expected.
```jsx
const [isWidthMet, setIsWidthMet] = useState<boolean>(
    typeof window !== 'undefined' && window.innerWidth <= width
  )
```

benalleng:
Closes #264

when run locally the server no longer issues an error stating that the window is undefined.

### DIFF
--- a/hooks/useMediaQuery.ts
+++ b/hooks/useMediaQuery.ts
@@ -7,7 +7,7 @@ interface UseMediaQueryParams {
 
 export const useMediaQuery = ({ width }: UseMediaQueryParams): boolean => {
   const [isWidthMet, setIsWidthMet] = useState<boolean>(
-    window.innerWidth <= width
+    typeof window !== 'undefined' && window.innerWidth <= width
   )
 
   useEffect(() => {


vinayaksh42 (approved):
LGTM! the challenges in which this hook is being used are working as expected now.

PR #267, opened 2023-03-09
Simplify templates for better reuseability

benalleng:
Some of the most recent copy PR's, #263, #266 are using a template but the template is still too limiting with what can be put in it so i removed some of the content from the template allowing them to be apart of the {children} in the content/index.ts route instead.
TODO:
Scope where else the templates can be simplified to aid in reuseability


benalleng:
Going to close this and just give feedback to the PR's currently addressing these issues

PR #273, opened 2023-03-10
Add intro page for genesis challenge

### LINKED ISSUE #260: Add an intro page for challenge 1 in chapter 1
This should be the next page after the second intro page to be added in #259. It's the first page you see when entering challenge 1. It uses the same template as the [challenge 2 intro page](https://savingsatoshi.com/en/chapters/chapter-1/transacting-1). The design is [in Figma here](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=2374%3A26492&t=1WVTKCVc1kPVBDfc-1).

image

afterburn:
This PR closes #260.

### DIFF
--- a/content/chapters/chapter-1/index.tsx
+++ b/content/chapters/chapter-1/index.tsx
@@ -20,6 +20,7 @@ export const metadata = {
   lessons: [
     'genesis-1',
     'genesis-2',
+    'genesis-3',
     'transacting-1',
     'transacting-2',
     'transacting-3',
--- a/content/lessons/genesis/genesis-1/index.tsx
+++ b/content/lessons/genesis/genesis-1/index.tsx
@@ -1,45 +1,27 @@
 'use client'
-
 import { useTranslations } from 'hooks'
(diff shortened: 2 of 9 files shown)


content/chapters/chapter-1/index.tsx
```diff
@@ -20,6 +20,7 @@ export const metadata = {
   lessons: [
     'genesis-1',
     'genesis-2',
+    'genesis-3',
```
vinayaksh42, line 23: I think you will also have to import `genesis-3` to the `content/index.ts` file. Right now I am getting the following error for genesis-3 page:
Screenshot 2023-03-11 at 2 02 17 AM

vinayaksh42 (changes requested):

content/chapters/chapter-1/index.tsx
afterburn: Thanks @vinayaksh42, I forgot the import there, nice catch.

vinayaksh42 (approved):
LGTM

Issue #259, opened 2023-03-09
Add a second page to the chapter 1 intro


GBKS:
This page goes after [this one](https://savingsatoshi.com/en/chapters/chapter-1) and then leads into the first challenge. The copy in Figma ([here](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=2374%3A25234&t=1WVTKCVc1kPVBDfc-1)) should be up-to-date (just checked it). An export-ready image can be found [here](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=2323%3A25490&t=1WVTKCVc1kPVBDfc-1) (select the image layer and press export in the bottom of the right sidebar).

![image](https://user-images.githubusercontent.com/695901/223947904-b719aaf4-8073-4e4e-bd1f-652d195ab00f.png)

vinayaksh42:
I would like to work on this issue, It might be a little tricky to add this because the introductions are displayed with the URL - `/chapters/chapter-1`, and if we add a dynamic folder named `[intro]` in the `chapters/[slug]` folder it interferes with the `[lesson]` folder as we can't have two dynamic folders under a single folder. One of the possible works around this issue can be modifying the `introduction.tsx` template to sequentially display the two different intros.

vinayaksh42:
> @vinayaksh42 is it possible to treat this page as genesis-1 similarly to how transacting-1 is this intro [page](https://savingsatoshi.com/en/chapters/chapter-1/transacting-1)?

Yes, we can do that but it will require a few modifications to the template for Introduction

benalleng:
Maybe it could be Intro-1 and Intro-2? i guess it all dependes on what we want the Navbar to display 
![image](https://user-images.githubusercontent.com/108441023/224387491-cd97c8c2-e1cd-43c2-a9be-1b06016b0b04.png)
or
![image](https://user-images.githubusercontent.com/108441023/224387551-b9715beb-a544-4c84-98e3-58f6468e583d.png)

PR #274, opened 2023-03-10
Second Intro for Chapter 1

vinayaksh42:
fixes: #259 

Changes:
- Added the second intro for chapter 1
- shifted the styling and layout logic of the introduction template to `introductionLayout.tsx` to avoid code repetition.


afterburn (approved):

afterburn:
There is a small difference between `/en/chapters/chapter-1` and `/en/chapters/chapter-1/holocat` for the first paragraph, one has pt-3 the other has mt-4 which causes a slight jump in the text after you navigate

vinayaksh42:
> There is a small difference between `/en/chapters/chapter-1` and `/en/chapters/chapter-1/holocat` for the first paragraph, one has pt-3 the other has mt-4 which causes a slight jump in the text after you navigate

fixed the requested change

GBKS:
Looking good, thanks for taking this one. Small tip that the image should be twice the size. In the issue, I linked to an export-ready frame with the right settings. Just need to click export in the bottom-right sidebar on that one. I will push some image updates next will and update it, so I'll tackle it.

PR #279, opened 2023-03-11
Further simplifications to translation system

afterburn:
Since all translation files have now been combined into their own large JSON file, we no longer need complex loading functionality in `translate.tsx`. We can also move away from using `require` and use `import` instead.

PR also contains some clean up.


ecurrencyhodler:
tACK! Didn't run into any issues.

ecurrencyhodler:
Let's see if we can get feedback on this by Sunday morning Pt. I'll take another look and merge if there are no objections. 

Kevin also suggested we should prob wait for #266 to merge.

benalleng:
Late review but it looks good to me! Nice simplification!

PR #280, opened 2023-03-11
Refactor lesson structure to allow for multiple intros

### LINKED ISSUE #276: NavBar /[lesson]/ path display
At the moment the filesystem in our app is designed such that `/chapters/[slug]/[lesson]` this limits us to having only a single introduction page, specifically on `/chapters/chapter-1` before it moves on to the lesson content. Adding multiple introduction pages creates problems based on the 2 possible paths. 

1. Create a new dynamic path such that `/chapters/[slug]/[intro] OR [lesson]`. however, this is not possible according to @vinayaksh42 research on it in #259
1. [current method] Simply add any introduction page beyond the second one to the [lesson] path. However, this causes the site to treat the introduction pages following the first one AS a lesson, creating some confusion or at least it leaves us with a poorly optimized NavBar tracking; being incorrect most of the time.

Paths forward:
1. Allow the copy in each page be custom set in the NavBar and thus allow us to make it whatever we want but we would have to make it every time
2.  Find a way to redesign the [lesson] path to become a [page] path in which the layout is organized as follows:

#### Intro
-Will not display Challenge numbers in the NavBar
image

#### Challenges
-Will display Challenge numbers based on the lesson content in the NavBar
image

#### Outro
-Again will not display Challenge info in the NavBar
image

our problem is that the intro and outro can be only one page each

afterburn:
This PR closes #276.

## Explanation
Because we now find ourselves in a place where multiple intros could be needed for a chapter, I think it no longer makes sense to keep `/chapters/chapter-1` responsible for rendering the `<Intro>` components. I propose we don't render anything at all on `/chapters/chapter-1`. Users will no longer be sent to `/chapters/chapter-1` when they click Start chapter on the `/chapters` page.

Instead, users will be sent to `/chapters/chapter-1/intro-1`, if multiple intro's are present in the chapter's metadata, then the next button will redirect the users to `/chapters/chapter-1/intro-2`, etc. Once the final intro has been hit, the next url will point to the first lesson (e.g. `/chapters/chapter-1/genesis-1`).

In other words, introductions no longer exist, everything is a lesson now.

Curious to hear your thoughts.

## Next steps
- Introduce hook that will determine the 'next' url so we no longer manually have to construct urls to other lessons, chapter metadata should be the only place where we move things around.
- Add the same logic for outro pages, I can imagine we at some point might want to add multiple outro pages.
- Add support for dynamic slugs, `/chapters/chapter-1` doesn't mean much, instead I think it would be nice if we could do something like `/chapters/secrets-in-plain-sight` or something, better for SEO too.


ecurrencyhodler:
I think this is a very clean fix. It allows us to be flexible with how many screens we can place in the intro. It looks like the same applies for the outro.

One concern I have is that the refactor comes right before our deadline, but I've tested it Versel and haven't seen any issues yet.

ecurrencyhodler (approved):

ecurrencyhodler:
Let's see if we can get feedback on this by Sunday morning Pt. I'll take another look and merge if there are no objections. 

Kevin also suggested we should prob wait for #266 to merge.

afterburn:
Found one bug where the chapter state doesn't get marked as complete, will resolve soon

ecurrencyhodler:
Tested again. Merged. 

Question: What if we used this standard format for urls and challenges?

/chapter-1/challenge-1

instead of

chapter-1/genesis-1

That way we can avoid using challenge titles (which could change)?

PR #300, opened 2023-03-13
Fix Enter key behavior on Input

### LINKED ISSUE #299: Chapter 1, Challenge 1, Part 1: Repeatedly pressing enter makes input area disappear
**Demo QA**

**Environment**: Linux (Ubuntu), Chrome (private browser, so cache is totally cleared)

**Potentially related issue**: https://github.com/saving-satoshi/saving-satoshi/issues/286

**Description**: If I keep pressing enter instead of doing the challenge properly, I am slowly able to make the input area disappear.

Here's what it looks like after a bunch of presses:

Screenshot from 2023-03-13 15-42-58

Eventually the screen becomes this:

Screenshot from 2023-03-13 15-42-29

To resolve it, I have to refresh to get the input area back so I can properly attempt the challenge (select all with CTRL + A, and backspace also works, but I don't expect that to be the common flow).

**Priority:** Probably low, since no one should be manually typing the scriptSig (they should be copy/pasting). The only somewhat valid user flow I can think of is if the user accidentally presses enter once at the beginning, then pastes in the correct answer. That will not be accepted.

**Possible fix:** Disable usage of the enter key for this input area.

benalleng:
Closes #299 
[Preview](https://saving-satoshi-git-fork-benalleng-issue299-saving-satoshi.vercel.app/en/chapters/chapter-1/genesis-2)

### DIFF
--- a/ui/common/Input.tsx
+++ b/ui/common/Input.tsx
@@ -91,7 +91,7 @@ export default function Input({
   }
 
   const handleKeyDown = (event: React.KeyboardEvent<HTMLTextAreaElement>) => {
-    if (!event.key.match(pattern)) {
+    if (!event.key.match(pattern) || event.key === 'Enter') {
       event.preventDefault()
     }
   }


satsie (approved):
tACK daf8ab6d5c52f0da48bcb4f02679c448e86de7e2 

It works! I pressed enter a ton of times and the input area did not shrink. I also pasted in the correct answer after pressing enter, and my answer was accepted. Thanks for grabbing this ticket so quickly!

ecurrencyhodler:
tACK!

PR #302, opened 2023-03-13
Update copy for Ch.1 Challenge 1 pt.1

### LINKED ISSUE #298: Copy: Chapter 1, Challenge 1 - Stale copy regarding hidden message
**Demo QA**

**Environment:** Linux (Ubuntu), Chrome (private browser, so cache is totally cleared)

**Issue**: Need a sanity check here. I think this bit on the last sentence is stale copy: "and see if we can find the hidden message that was mentioned" 

Screenshot from 2023-03-13 15-29-11

I don't recall a hidden message ever being mentioned in the new version of the story that we have deployed

Opening the issue here so it can be tracked with the project, but it's best to hash out copy changes in Figma.

ecurrencyhodler:
Closes #298 

I also added notes as headings to help navigate the doc better.

## Pull Request checklist

- Clean commit history

### DIFF
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -8,7 +8,7 @@ const translations = {
     coming_soon: 'Coming soon',
     start_chapter: 'Start chapter',
   },
-
+///ABOUT PAGE
   about: {
     title: 'About',
     mobile_title: 'What it’s about',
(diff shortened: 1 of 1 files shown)


satsie (approved):
LGTM! Thanks for taking this! 

Not sure what format we prefer for comments (two or three slashes? I don't see any precedents set in this file) but that certainly shouldn't hold things up. Ship it!

PR #303, opened 2023-03-13
Chapter 1, Challenge 2: Tweak block explorer link

satsie:
In chapter 1, challenge 2, there are explicit instructions to open up the transaction details (see instruction number 2):

Screenshot from 2023-03-13 17-10-57

However, the link we provide already has the details opened up. There are two ways we can smooth this out:

1. Remove instruction number 2 and don't tell the user to expand the details
2. Don't expand the transaction details for the user and have them do it themselves

I like the idea of not expanding the transaction details for the user because it teaches them that you can open transactions up and take a look inside :smile:. But I acknowledge that all block explorers are different and there are pro's to the other option of just having less instructions. Happy to go either way on this!

### DIFF
--- a/app/chapters/[slug]/transacting/transacting-1/page.tsx
+++ b/app/chapters/[slug]/transacting/transacting-1/page.tsx
@@ -55,7 +55,7 @@ export default  function Genesispt2() {
                     ></div>
                     <div className='pb-16 pt-9'>
                         <Link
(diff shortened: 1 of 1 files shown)


satsie:
oof. worked off a super stale copy of master. Re-doing this PR on a new branch.

PR #305, opened 2023-03-13
Accept Double and Single quotes in Terminal Challenge

### LINKED ISSUE #301: Chapter 1, Challenge 1, Part 2: Quotes not accepted on echo command
**Demo QA**

**Environment:** Linux (Ubuntu), Chrome (private browser, so cache is totally cleared)

**Issue**: The following command, with the scriptSig in single quotes, is valid and should be accepted as an answer:

```
echo '04ffff001d0104455468652054696d65732030332f4a616e2f32303039204368616e63656c6c6f72206f6e206272696e6b206f66207365636f6e64206261696c6f757420666f722062616e6b73' | xxd -r -p
```

You can confirm this by running it on your own computer

```
$ echo '04ffff001d0104455468652054696d65732030332f4a616e2f32303039204368616e63656c6c6f72206f6e206272696e6b206f66207365636f6e64206261696c6f757420666f722062616e6b73' | xxd -r -p
��EThe Times 03/Jan/2009 Chancellor on brink of second bailout for banks
```

**Actual result:** It is rejected

Screenshot from 2023-03-13 16-09-31

**Other:** I really don't know if this is worth fixing, and this issue may even be generalized to "I expect the terminal to behave like a real command line interface but it doesn't." I know reworking the terminal has been brought up before, and am not sure where we want to draw the line. Should the terminal behave like a real one? Is it fine to let it slide because this is a game made for learning? 

Another thing to consider is this behavior applies to people that didn't understand `$scriptSigHex` is a variable, or don't want to use it. The former would probably be a beginner coder, and the latter likely an intermediate coder. Either way, I question if it's worth the extra effort.

On another note, to my delight, the command without quotes works! This is awesome!

Screenshot from 2023-03-13 16-26-51

benalleng:
Closes #301 
[Preview](https://saving-satoshi-git-fork-benalleng-issue301-saving-satoshi.vercel.app/en/chapters/chapter-1/genesis-3)

### DIFF
--- a/ui/lesson/TerminalChallenge.tsx
+++ b/ui/lesson/TerminalChallenge.tsx
@@ -89,8 +89,6 @@ export default function TerminalChallenge({
     let answerValue: string
     let newLines: { value: string; type: string }[]
 
-    console.log(sanitizedInput)
-
     if (inputs.includes(sanitizedInput)) {
       if (sanitizedInput === `echo ${expectedInput} | xxd -r -p`) {
         varInput = expectedInput
--- a/ui/lesson/TerminalChallenge.tsx
+++ b/ui/lesson/TerminalChallenge.tsx
@@ -81,11 +81,15 @@ export default function TerminalChallenge({
       `echo ${expectedInput} | xxd -r -p`,
       `echo ${expectedInput.userVariable} | xxd -r -p`,
       `echo ${expectedInput.value} | xxd -r -p`,
-      `echo ${sanitizedInput.split(' ')[1]?.match(/^[\da-f]+$/i)} | xxd -r -p`,
+      `echo ${sanitizedInput
+        .split(' ')[1]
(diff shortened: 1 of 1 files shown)


ecurrencyhodler:
Tested single and double quotes on both challenge 1 and 2 desktop chrome and mobile chrome and it works great.

One question I have is that should this also pertain to the variable?
Screenshot 2023-03-13 at 2 50 02 PM

benalleng:
> One question I have is that should this also pertain to the variable?

IMO no, double and single quotes are what the computer treats as a string type, i.e. the text inside here should not be read as a variable as a string is meant to represent text data. A variable should not be inside a string

ecurrencyhodler:
Then it's a tACK for me.

afterburn (approved):

PR #306, opened 2023-03-13
Update commonError for Transacting-3

benalleng:
Updates the commonError following the update to the logic to accept the scriptPubKeyBytes string
image

### DIFF
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -8,7 +8,7 @@ const translations = {
     coming_soon: 'Coming soon',
     start_chapter: 'Start chapter',
   },
-///ABOUT PAGE
+  ///ABOUT PAGE
   about: {
     title: 'About',
     mobile_title: 'What it’s about',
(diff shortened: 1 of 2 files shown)


ecurrencyhodler:
I really like this addition suggestion. 

It catches the flow where a player tries to do something similar to the challenge 1. 

Tagging @GBKS and @satsie for their thoughts.

GBKS:
Yes, love it. That's exactly the type of custom responses we want.

i18n/locales/en.ts
```diff
@@ -8,7 +8,7 @@ const translations = {
     coming_soon: 'Coming soon',
     start_chapter: 'Start chapter',
   },
-///ABOUT PAGE
+  ///ABOUT PAGE
```
afterburn, line 11: Not sure why we need this comment but we can remove it later

afterburn (approved):

i18n/locales/en.ts
satsie: These were added in https://github.com/saving-satoshi/saving-satoshi/pull/302

ecurrencyhodler: They're just headings so contributors can more easily navigate the doc.

PR #235, opened 2023-03-04
Navigation back button go to chapters page

gadabra:
in order to fix #206. I removed useRouter and replaced it with useContentRouter to improve content route management, and I altered the back button in the Navigation Component to link to the chapters page.

PR #307, opened 2023-03-14
Change back button arrow icon to power off icon

GBKS:
A follow-up to #235. Replaces the arrow left button with a unique button so users don't think it has the same functionality as the browser back button (which looks pretty much the same). This is now a power off icon. Not commonly used, but idea is that this connects to the story in that we as the user are hacking away at our computer in a challenge and press the power off button to disconnect and go back to IRL. Happy to change this later if there are other ideas.


shaavan (approved):
Looks great! 

That's an intriguing idea that we are working on our computers while we solve challenges. The power-off button makes complete sense in this situation!

I have a fun idea if we go forward with "powering off our PC." We can also add a redirect page when the power off is clicked. That says "Powering off your terminal..." with some fancy animation.

PR #314, opened 2023-03-19
Locale persistence + minor bugfixes

### LINKED ISSUE #313: Locale selection persistence
Upon selecting the language in the URL or in the future with the language selector in #148 it properly loads the page without any hydration or render errors. however, clicking any link forces the locale selection and url back to the default `/en` locale.

I think there are two ways to fix this
1. forcing buttons to put in locale specific urls to their links so that the buttons and Links always leads to the locale the user selected or
2. <s>the `middleware.ts` logic no longer forces the user onto a specific locale, only that it defaults to english so that going from `savingsatoshi.com` -> `savingsatoshi.com/en/chapters` is valid but that 
`savingsatoshi.com/nl` -> `savingsatoshi.com/nl/chapters` also works
I am not %100 sure that my first thought in 2. is a solution, but I do believe a middleware or a url based solution is better than a button based one as it will make our jobs much easier moving forward than having to code in locale logic on buttons</s>

benalleng:
Closes #313

### DIFF
--- a/app/[lang]/about/page.tsx
+++ b/app/[lang]/about/page.tsx
@@ -117,7 +117,9 @@ export default function About({ params }) {
           <p className="pt-4 pb-8 text-center font-nunito text-xl">
             {t('about.satoshi_needs_you')}
           </p>
-          <Button href="/chapters">{t('hero.start_journey')}</Button>
+          <Button href={`/${params.lang}/chapters`}>
+            {t('hero.start_journey')}
+          </Button>
         </div>
       </div>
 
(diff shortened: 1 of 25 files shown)


ecurrencyhodler:
On web it works great. Url to nl is preserved. Some mixing text but that's prob cuz there aren't translations for it yet.

Mobile designs seemed to have been affected though. Sizing isn't correct. 

image

vinayaksh42:
> On web it works great. Url to nl is preserved. Some mixing text but that's prob cuz there aren't translations for it yet.
> 
> Mobile designs seemed to have been affected though. Sizing isn't correct.
> 
> image

I think this is unrelated to this PR as the same behavior is visible on the live website - https://savingsatoshi.com/nl/chapters/chapter-1/intro-2 (might be due to the missing translation)

afterburn (approved):

afterburn:
Yes there are some missing translations for `nl` currently.

PR #315, opened 2023-03-20
Language switch button

### LINKED ISSUE #148: Language switch
**Describe the enhancement**
We now have support for translations but currently we automatically detect what translations to load based on the locale set in the user's browser settings. It might be a better experience if users are free to select their preferred language from a dropdown menu.

1. User navigates to savingsatoshi.com, we autodetect locale and choose a language.
2. User might prefer a different language so he/she needs a way to change the language. <-- this is missing

**Additional information**
Users can actually change the translations by changing the locale in their browser settings, but this is not the most pleasant experience.

vinayaksh42:
Fixes #148 

Changes introduced:
- Added a button for switching the language
- ~~storing the selected language in local storage~~

To-do:
- ~~link the local storage with the `middleware.ts` in order to apply the selected language (can't access the local storage from `middleware.ts`, other options might be storing it as cookie or using the [NEXT_LOCALE](https://nextjs.org/docs/advanced-features/i18n-routing#leveraging-the-next_locale-cookie))~~

One way to reroute to the current locale is by using the `useRouter` and `usePathname` from `next/navigation`. The language switch will detect the current locale from the pathname instead of using the local storage and if the user changes the locale the page will re-direct to the selected locale. 

update: currently, if a user selects a locale and then goes to a new page the locale is not persisting but this issue can be solved by the PR #314


components/Navbar/LangSwitch.tsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  return (
+    <div className="relative inline-block text-left">
+      <div className="ml-4 h-10 w-10">
+        <button
+          type="button"
+          className="text-grey-300 h-10 cursor-pointer"
+          id="language-menu"
+          aria-label="profile"
```
GBKS, line 33:
```suggestion
          aria-label="Choose language"
```

components/Navbar/LangSwitch.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+      <div
+        className={`${
+          isOpen ? 'block' : 'hidden'
+        } absolute right-0 mt-2 w-[125px] origin-top-right  border border-white border-opacity-25 bg-back shadow-lg`}
+        role="menu"
+        aria-orientation="vertical"
+        aria-labelledby="language-menu"
+      >
+        <div className="py-1" role="none">
```
GBKS, line 50: What do you need `role="none"` for?

components/Navbar/LangSwitch.tsx
```diff
@@ -0,0 +1,69 @@
...
+          {i18n.languages.map((language, index) => (
+            <button
+              key={index}
+              className={`${
+                currentLanguage == i18n.locales[index]
+                  ? 'text-white'
+                  : 'text-white text-opacity-75'
+              }  px-5 py-2 text-left font-nunito text-lg font-bold `}
+              role="menuitem"
```
GBKS, line 59: Do you need `role="menuitem"`?

components/Navbar/LangSwitch.tsx
```diff
@@ -0,0 +1,69 @@
...
+        className={`${
+          isOpen ? 'block' : 'hidden'
+        } absolute right-0 mt-2 w-[125px] origin-top-right  border border-white border-opacity-25 bg-back shadow-lg`}
+        role="menu"
+        aria-orientation="vertical"
+        aria-labelledby="language-menu"
+      >
+        <div className="py-1" role="none">
+          {i18n.languages.map((language, index) => (
+            <button
```
GBKS, line 52: Might need an extra [aria-selected](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected) attribute to indicate if the button is the currently selected language or not.

GBKS (changes requested):
So great to see this work. Just had some really minor feedback.

vinayaksh42:
> So great to see this work. Just had some really minor feedback.

Thanks a lot for the feedback

hooks/useLocaleUrl.ts
```diff
@@ -0,0 +1,16 @@
+'use client'
+
+export function generateNewUrl(pathname: string, language: string): string {
```
afterburn, line 3: Please use the `use` naming convention for hooks, so `useGenerateNewUrl` and `useCurrentLocale` or something.

I also think `generateNewUrl` does not describe very well what the hook is actually doing, perhaps take a look at `useLocalizedRoutes` and see if you can incorporate this logic into there because we seem to be creating a bunch of hooks that do basically the same but slightly different and this can get confusing for other contributors.

components/Navbar/LangSwitch.tsx
```diff
@@ -0,0 +1,77 @@
...
+        aria-labelledby="language-menu"
+      >
+        <div className="py-1">
+          {i18n.languages.map((language, index) => (
+            <button
+              key={index}
+              className={clsx(
+                'px-5 py-2 text-left font-nunito text-lg font-bold',
+                {
+                  'text-white': currentLanguage == i18n.locales[index],
```
afterburn, line 60: Lets approach this in a slightly different way.

Instead of using the `index` to get check the locale lets modify `i18n/config.ts` so it exports something like this:
```ts
[
  { locale: 'en', label: 'English' },
  { locale: 'nl', label: 'Dutch' }
]
```
This will prevent issues when someone is not aware of this index-based approach your code uses for locales/language, by using the code above we can ensure locales are mapped to the correct labels.

afterburn (changes requested):
Thanks @vinayaksh42, requested some minor changes

hooks/useLocaleUrl.ts
```diff
@@ -0,0 +1,16 @@
...
+  const pathnameWithoutLanguage = pathname.replace(/^\/(en|nl)\b/, '')
```
benalleng, line 4: Does this function only search for `en` or `nl` in the url? if so we might want to do some future proofing for any translations in the future

components/Navbar/LangSwitch.tsx
```diff
@@ -0,0 +1,77 @@
...
+          {
+            block: isOpen,
+            hidden: !isOpen,
+          }
+        )}
+        aria-orientation="vertical"
+        aria-labelledby="language-menu"
+      >
+        <div className="py-1">
+          {i18n.languages.map((language, index) => (
```
benalleng, line 54: > "Want to help translate saving satoshi to more languages?"

We might want to add this option to the end of our locale list

benalleng (changes requested):
Left a few comments!

components/Navbar/LangSwitch.tsx
benalleng: Perhaps this tooltip from the conversation in #148 could be in a "Your Language" option.
I am going to look around on some other websites to see if there is a common theme people use

components/Navbar/LangSwitch.tsx
```diff
@@ -0,0 +1,77 @@
...
+                {
+                  'text-white': currentLanguage == i18n.locales[index],
+                  'text-white text-opacity-75':
+                    currentLanguage != i18n.locales[index],
+                }
+              )}
+              onClick={() => handleLanguageClick(index)}
+              aria-selected={
+                currentLanguage == i18n.locales[index] ? true : false
+              }
```
benalleng, line 68: We should add a focusOut handler to also close the drop down so I do not have to click the button again

benalleng (changes requested):

hooks/useLocaleUrl.ts
vinayaksh42: yes, currently this regular expression is only looking for en and nl.

hooks/useLocaleUrl.ts
vinayaksh42: Yes, I think I will be able to utilize the `usePathData` hook for all the information that I require for re direct (after a modification for providing the entire pathName as well).

vinayaksh42: the function inside the generateNewUrl hook is very specific and I think having it inside the component file itself will make a lot more sense as I doubt that we will require this function for any other component.

components/Navbar/LangSwitch.tsx
vinayaksh42: I have not addressed this feedback as the design/ plan for adding a contribution path for language is not clear yet.

vinayaksh42:
with the changes merged for PR #314 the locale seems to be persisting upon selecting a language different from the default language of my browser.

benalleng:
> with the changes merged for PR #314 the locale seems to be persisting upon selecting a language different from the default language of my browser.

Is this a behavior you didn't expect? I believe that the selector should override your default if choosing to select a new language

hooks/useLocaleUrl.ts
benalleng: `.replace(/^\/..\b/, '')`
Would this replace() work to replace any future language code?

benalleng (approved):
Thanks for resolving my comments! I'll go ahead and open an issue about directing translators via the dropdown

afterburn (approved):
Works like a charm!

@GBKS Currently its only possible to switch language on the home, about and chapters page, we render a different Topbar for lessons which does not include the language switch. Do we want users to be able to switch their language from within lessons?

GBKS:
@afterburn I'd only add that option if people ask for it. Since we lock challenges until people complete previous ones, we won't have a ton of situations where people get deep-linked into content. Pretty much everyone will start with the home and chapters pages. So we might be OK not having the language toggle there?

PR #329, opened 2023-03-28
Block navigation to locked lessons

### LINKED ISSUE #155: Block navigation to locked challenges
**Describe the bug**
Users are currently able to navigate to any challenge, even if they haven't unlocked it yet.

**To Reproduce**
Using a fresh account either go to the first Genesis page, then click on the Transacting challenge in the Navbar or go to [this url](https://savingsatoshi.com/en/chapters/chapter-1/transacting-1) directly.

**Expected behavior**
Users should only be able to navigate to challenges that they have unlocked. If they try to navigate to a locked challenge we should either show a "This challenge is locked" message or redirect them back to either chapter overview or the last challenge that they have unlocked.

vinayaksh42:
Fixes #155 

Changes:
- added hook `useLessonUnlocked` for checking if the challenge is locked or unlocked and returning the URL for the last unlocked lesson if the user visits a locked lesson.
- added `PageLocked` component for conditionally displaying the lesson content or a button for re-directing back to the last unlocked lesson.

To-do:
- A way to avoid the `if-else` nesting (If possible)


components/LessonLocked.tsx
```diff
@@ -0,0 +1,29 @@
...
+    return children
+  } else if (typeof result === 'string') {
+    return (
+      hydrated && (
+        <div className="flex h-full w-full grow flex-col items-center justify-center">
+          <span className="mb-10 text-4xl text-white">
+            This part of the chapter is currently locked!
+          </span>
+          <Button href={result} size="small">
+            &larr; Back to Last Unlocked Lesson
```
vinayaksh42, line 23: @GBKS any copy-related suggestion for this part?

GBKS: Locked challenges should simply not be clickable at all.

afterburn (changes requested):
I dont see anything wrong with the code perse but I think the approach is wrong, or at least incomplete.

As @GBKS alluded to earlier, users should not even be able to navigate to locked lessons in the first place. We can (partially) prevent this from happening by modifying the `Tab` components to check if the lesson it points to is available to the user, if it is we render a `Link`, otherwise we render a non-clickable `div` or whatever.

Then if a user attempts to manually navigate to a lesson by entering its url into the browser, we should have a hook that immediately runs on the Lesson page which redirects the user back to the last lesson they **_did_** unlock. In other words, we shouldn't need a page that says "This lesson is currently locked, click here to go back", we should simply redirect the user and not render anything at all, except for maybe a loader if that's required.

vinayaksh42:
update:
- Locked challenges are unclickable in the navbar.
- The user automatically gets redirected to the last unlocked lesson (no more "This lesson is currently locked..." message). instead, it displays "Loading Challenge..." till the time it redirects.

ecurrencyhodler:
1. tACK on functionality. 
2. Can you explain your 2nd statement Vinayak? Why would someone need to get redirected if they can't click on the locked lesson?

vinayaksh42:
> 1. tACK on functionality. 
> 
> 2. Can you explain your 2nd statement Vinayak? Why would someone need to get redirected if they can't click on the locked lesson?

In case the user directly enters the URL to a locked chapter.

components/LessonLocked.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+  useEffect(() => {
+    if (typeof result === 'string') {
+      router.replace(result)
+    } else {
+      setLessonContent(children)
+    }
+  }, [])
+
+  return lessonContent
```
afterburn, line 24: Let's not put jsx into state, I think a better solution is something like this:
```jsx
export default function LessonLocked({ children }) {
  const router = useRouter()
  const result = useLessonUnlocked()

  useEffect(() => {
    if (typeof result === 'string') {
      router.replace(result)
    }
  }, [])

  return <div className="flex h-full w-full grow flex-col items-center justify-center">
    <span className="mb-10 text-4xl text-white">Loading Challenge...</span>
  </div>
```

Also I'm wondering why you're not just doing this in the `/app/[lang]/chapters/[slug]/[lesson]/layout.tsx` file since I doubt we'll reuse this new `LessonLocked.tsx` file much.

hooks/useLessonUnlocked.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export const useLessonUnlocked = () => {
+  let { lang, pageId, chapterId, lessonId } = usePathData()
+  const userInfo = useUser()
+
+  // if someone vists the website for the first time
+  if (userInfo.user === undefined) {
+    return `/${lang}`
+  } else {
```
afterburn, line 19: General rule of thumb for keeping your code clean and easy to understand for other developers that I push alot is to return early (which is what you are doing here), but also to steer away from else blocks if there is no code after that.

```
if (userInfo.user === undefined) {
  return `/${lang}`
}

const userLesson = userInfo.user.progress.lesson
const userChapter = userInfo.user.progress.chapter
const lessonArray = chapters[chapterId].metadata.lessons
const reDirectURL = `/${lang}/${pageId}/${userChapter}/${userLesson}`

if (userChapter.split('-')[1] < chapterId.split('-')[1]) {
  return redirectUrl
}

if (lessonId.includes('intro')) {
  return true
}

if (lessonId.includes('outro')) {
  if (removeAfterHyphen(userLesson) === 'outro') {
    return true
  }
  
  return redirectUrl
}

etc

File: hooks/useLessonUnlocked.ts [hunk 4]
```diff
@@ -0,0 +1,54 @@
...
+    const lessonArray = chapters[chapterId].metadata.lessons
+    const reDirectURL = `/${lang}/${pageId}/${userChapter}/${userLesson}`
+
+    if (userChapter.split('-')[1] < chapterId.split('-')[1]) {
+      return reDirectURL
+    } else {
+      if (lessonId.includes('intro')) {
+        return true
+      } else {
+        if (lessonId.includes('outro')) {
```
afterburn, line 31: Why are we returning `true` if an intro or outro lesson is hit? If I understand correctly that means the user will _not_ be redirected while they should. If I haven't unlocked chapter-2 intro-1 I should also be redirected away right?

afterburn (changes requested):
Thanks @vinayaksh42 I requested some more changes to keep the code a little cleaner/easy to understand/read.

hooks/useLessonUnlocked.ts
vinayaksh42: The user will be redirected if they directly go to chapter-2 intro-1 without unlocking chapter-2 because the first `if` statement checks whether the particular chapter has been unlocked.

I have updated the logic. All the 3 different arrays (intros, lessons, and outros) are combined into a single array, and then a comparison is made based on the index of the saved userProgress and the lesson that the user is trying to access. Before I was checking for intro or outro page because they were not part of the array.

vinayaksh42:
During testing, I discovered a potential issue where if a user has completed all the lessons in chapter 1 and decides to redo the Genesis lesson, their user progress gets reset to Chapter 1, Lesson 1. I believe this should not occur. redoing an old lesson should not change the userProgress.

opening a new issue for this

benalleng:
I found two things that i think should be adjusted
1. This is mostly a question but has any temporary user progress been implemented?
I completed genesis challenge-1 but then attempted to go to transacting-3, instead of being directed to genesis-2 i was taken all the way back to genesis-1
2. Anything that doesn't exist, i.e. something that would normally lead to a 404 error leads instead to a client-side error `/en/chapters/chapter-1/oogabooga` presents us with this error.
```
Application error: a client-side exception has occurred (see the browser console for more information).
```

console.log:
```
TypeError: d.K[u][h] is undefined
```

vinayaksh42:
> I found two things when testing this deployment
> 
> 1. This is mostly a question but has any temporary user progress been implemented?
>    I completed genesis challenge-1 but then attempted to go to transacting-3, instead of being directed to genesis-2 i was taken all the way back to genesis-1
> 2. Anything that doesn't exist, i.e. something that would normally lead to a 404 error leads instead to a client-side error `/en/chapters/chapter-1/oogabooga` presents us with this error.
> 
> ```
> Application error: a client-side exception has occurred (see the browser console for more information).
> ```
> 
> console.log:
> 
> ```
> TypeError: d.K[u][h] is undefined
> ```

1. The progress is still being saved on completion of challenges (genesis-1, transacting-1), we can implement a per-page basis progress saving for local storage but then we will have to re-do it for the backend again.
2. Not sure, if this bug is being introduced by this PR or if it exists already but if you go to https://savingsatoshi.com/en/chapters/chapter-1/dwdwd it gives the same error that you mentioned.

ecurrencyhodler:
Okay we can create a separate issue for the bug. This is a tACK for me. I think we just need @afterburn to review then we can merge.

hooks/useLessonUnlocked.ts
```diff
@@ -0,0 +1,50 @@
+'use client'
+
+import { useUser, usePathData } from 'hooks'
+import { chapters } from 'content'
+
+function removeAfterHyphen(str: string): string {
+  return str.split('-')[0]
+}
+
+function getIndex(arr: string[], searchStr: string): number {
```
afterburn, line 10: I dont think we need a new function if all its doing is `arr.indexOf(searchStr)`, just use that instead of `getIndex` on line 37

hooks/useLessonUnlocked.ts
```diff
@@ -0,0 +1,50 @@
...
+  return arr.indexOf(searchStr)
+}
+
+export const useLessonUnlocked = () => {
+  let { lang, pageId, chapterId, lessonId } = usePathData()
```
afterburn, line 15: This can be a const if I'm not mistaken

hooks/useLessonUnlocked.ts
```diff
@@ -0,0 +1,50 @@
...
+  const userLesson = userInfo.user.progress.lesson
+  const userChapter = userInfo.user.progress.chapter
+  const lessonArray = [
+    ...chapters[chapterId].metadata.intros,
+    ...chapters[chapterId].metadata.lessons,
+    ...chapters[chapterId].metadata.outros,
+  ]
+  const redirectURL = `/${lang}/${pageId}/${userChapter}/${userLesson}`
+
+  if (userChapter.split('-')[1] > chapterId.split('-')[1]) {
```
afterburn, line 31: Some comments would be helpful here

afterburn (changes requested):
Thanks @vinayaksh42 code looks way better now, requested a couple minor changes after which we can merge.

vinayaksh42:
> Thanks @vinayaksh42 code looks way better now, requested a couple minor changes after which we can merge.

I have incorporated all the changes requested. I was testing the implementation and found that when the user has completed genesis-1 challenge and then tries to access a challenge that is currently locked it keeps on sending the user to the homepage. Basically, on the first load of the website, the hook was unable to fetch the local storage due to which it was redirecting to the homepage. I have updated the logic and added a new hook for fetching local storage items using `useEffect`. Now when the data fetched from local storage is undefined it keeps on displaying "loading challenge" until the actual userProgress is fetched.

app/[lang]/chapters/[slug]/[lesson]/layout.tsx
```diff
@@ -1,34 +1,29 @@
...
 import { useRouter } from 'next/navigation'
-import { useEffect, useState } from 'react'
 import Navbar from 'components/Navbar'
 import { lessons } from 'content'
 
 export default function Layout({ children, params }) {
   const router = useRouter()
-  const result = useLessonUnlocked()
-  const [displayLesson, setDisplayLesson] = useState<boolean>(false)
+  const userInfo = useLocalStorage('SavingSatoshiUser')
```
Comment (afterburn) [inline app/[lang]/chapters/[slug]/[lesson]/layout.tsx:10]: Do we really need a hook for this? Whats the advantage of this over calling `LocalStorage.getItem('SavingSatoshiUser')` directly here?

This would be a nice addition if we could pass a fallback value if the value for the given key couldn't be found but otherwise I don't see the point of introducing a hook.

vinayaksh42:
As per the suggestion made by @adamjonas , I have added a check which allows the lesson to be visible if the URL has parameter `dev=true`. 

Any lesson with param `dev=true` will be visible for example: https://saving-satoshi-h00uljucc-saving-satoshi.vercel.app/en/chapters/chapter-1/outro-1?dev=true

afterburn (approved):

PR #333, opened 2023-03-29
SHA-256 Hashing

### LINKED ISSUE #322: Ch.2 L1: SHA-256 challenge
**Current situation**
Our `<InputChallenge>` component has text on top and an Input field on the bottom.

**Desired situation**
In Chapter 2 users will need to be able to do sha256 hashing. They should be able to input text on the top and see the sha256 hashed output on the bottom of their screens (see [Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=2019-26043&t=2hKSLjtWDaVqPH0h-0)).

Investigate whether it makes sense to update the existing `<InputChallenge>` component to support this functionality, or if it makes more sense to create a new component. The new code should be written in a way that allows us to easily apply other hashing algo's (or entirely other kinds of transformations) to the users input.

The sha256 algo can be implemented using a client-side using library such as [sjcl](https://www.npmjs.com/package/sjcl), see a code snippet [here](https://stackoverflow.com/a/48738430) (their docs are somewhat lacking).

benalleng:
Closes #322 

All Ch. 2 Challenge 1 pages have been placed

The `statusBar` now takes in a boolean and a hint boolean to set it's states

Additionally there is a new `formatHash` function that can be reused through all 3 challenges

moving forward the prop `alwaysShow` will display the `statusBar` in all screens allowing us to maintain a one page format like this one

[Preview](https://saving-satoshi-git-fork-benalleng-issue322-saving-satoshi.vercel.app/en/chapters/chapter-2/intro-1?dev=true)

<details>
<summary>Answers</summary>

```
//challenge-1
input length > 8
//challenge-2
'popcorn'
//challenge-3
'trigonometry'
```

</details>


content/lessons/chapter-2/hashing-1.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export default function Hashing1({ lang }) {
+  const routes = useLocalizedRoutes()
+  const t = useTranslations(lang)
+
+  return (
+    <HashChallenge
+      answer={8}
+      next={`${routes.chaptersUrl}/chapter-2/hashing-2`}
+      label={'Enter anything'}
```
afterburn, line 14: If all you are passing is a string you don't need curly braces.

content/lessons/chapter-2/hashing-2.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { useLocalizedRoutes, useTranslations } from 'hooks'
+import { HashChallenge } from 'ui'
+
+export default function Hashing2({ lang }) {
+  const routes = useLocalizedRoutes()
+  const t = useTranslations(lang)
+
+  return (
+    <HashChallenge
+      answer={'0'}
```
afterburn, line 12: Same here

content/lessons/chapter-2/hashing-3.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { useLocalizedRoutes, useTranslations } from 'hooks'
+import { HashChallenge } from 'ui'
+
+export default function Hashing3({ lang }) {
+  const routes = useLocalizedRoutes()
+  const t = useTranslations(lang)
+
+  return (
+    <HashChallenge
+      answer={'00'}
```
afterburn, line 12: Same here

content/lessons/chapter-2/hashing-4.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { useLocalizedRoutes, useTranslations } from 'hooks'
+import { HashChallenge } from 'ui'
+
+export default function Hashing3({ lang }) {
+  const routes = useLocalizedRoutes()
+  const t = useTranslations(lang)
+
+  return (
+    <HashChallenge
+      answer={'000'}
```
afterburn, line 12: Same here

hooks/usePathData.ts
```diff
@@ -1,3 +1,5 @@
+// 'use client'
```
afterburn, line 1: Same here

ui/lesson/HashChallenge/HashChallenge.tsx
```diff
@@ -0,0 +1,151 @@
...
+  label2: string
+  answerHint?: string
+  auto?: boolean
+  hints?: boolean
+}) {
+  const [input, setInput] = useState('')
+  const [userInput, setUserInput] = useState('')
+  const [hintTimer, setHintTimer] = useState(false)
+
+  const inputView = (
```
afterburn, line 38: Looks like this isnt being used anywhere

ui/lesson/HashChallenge/HashPrompt.tsx
```diff
@@ -0,0 +1,42 @@
+import clsx from 'clsx'
```
afterburn, line 1: What is the purpose of this component, if its just used for markup lets move it to `Hasher` or `HashChallenge` instead.

ui/lesson/HashChallenge/HashChallenge.tsx
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  useEffect(() => {
+    const hintTimeout = () => {
+      setTimeout(() => {
+        setHintTimer(true)
+      }, 15000)
+    }
+    answerHint && hintTimeout()
+  })
```
afterburn, line 68: You are not clearing the timeout in the `useEffect`'s return function, there is also no dependency array so this effect will trigger on every render, if that is by design please explain that with a comment.

ui/lesson/HashChallenge/HashChallenge.tsx
```diff
@@ -0,0 +1,151 @@
...
+  label,
+  label2,
+  auto,
+  hints,
+}: {
+  // children: any
+  answer: string | number
+  next: any
+  label: string
+  label2: string
```
afterburn, line 29: Let's rename to something other than `label2` which is more descriptive of what it does, as a rule of thumb ask yourself how others are supposed to know as someone who has never used this component what `label` and `label2` do?

ui/lesson/HashChallenge/HashChallenge.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+import { useEffect, useState } from 'react'
+import { Lesson, LessonTabs, LessonPrompt, StatusBar, Tooltip } from 'ui'
+import LessonHasher from './HashPrompt'
+import clsx from 'clsx'
+
+/**
+ * @answer {string} correct answer to the challenge problem
+ * @next {string} link to next part of chapter
+ * @label {string} label on top of input field
```
afterburn, line 11: Also, `label2` seems to be missing here

ui/common/Hasher.tsx
```diff
@@ -0,0 +1,154 @@
...
+          </p>
+        </>
+      )}
+    </>
+  )
+}
+
+/*
+'000'
+vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvccxcvxcvxcvxcvxcvxcvcvcvcvvbbbbbbbbbbbbbgggggggggggggggggggggggggggggggggggggggggggdddddddddeeeeeeeeeeeeeeeeeeeeeeeeeeerrrrrrrrrrrrrrrrrrrrrrrrrrewwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwdddddddsaaaaaaaaaaaaaaaaaaaaaaqqwwweeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeefffffffffggggggggggggggggggggggggggggggggggggtttttttttttreerwerwerwerwerwerwerwerwerwererererwewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewewwwwwwwwwwwwwwwwwwwwwwwwwqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqwweeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
```
afterburn, line 153: Let's remove this

ui/common/Hasher.tsx
```diff
@@ -0,0 +1,154 @@
...
+                  'overlay-incomplete':
+                    (typeof answer === 'string' &&
+                      hash.startsWith(answer) === false) ||
+                    (typeof answer === 'number' && input.length < answer),
+                  underscore: !hash || !answer,
+                }
+              )}
+            >
+              {hash
+                ? `${hash.substring(0, 4)} ${hash.substring(
```
afterburn, line 87: I'm guessing this code is used to `splitastringlikethis` into `spli tast ring like this`, I don't think this is the way to do that. You should use a loop or something. I had a similar challenge [here](https://github.com/saving-satoshi/saving-satoshi/blob/master/ui/lesson/ScriptingChallenge/Runner/Hasher.tsx#L36), perhaps you can extract that to a helper function that we can reuse in this challenge and in the scripting challenge.

ui/common/Hasher.tsx
```diff
@@ -0,0 +1,154 @@
...
+    !clicked && handleHash(input)
+    if (typeof answer === 'string') {
+      onChange(hash.substring(0, answer.length))
+    } else {
+      onChange(hash)
+    }
+    return () => {
+      clearInterval(miner)
+    }
+  })
```
afterburn, line 60: No dependency array means this runs on every render, seems odd considering you have an interval (though you do clear it propery), can you explain what this does?

ui/common/Hasher.tsx
```diff
@@ -0,0 +1,154 @@
...
+  label: string
+  onChange: any
+  auto: boolean
+  input: string
+}) {
+  const [hash, setHash] = useState('')
+  const [clicked, setClicked] = useState(false)
+
+  const handleAutoChange = () => {
+    let randomNum = (Math.random() * 1000000000000000000).toString()
```
afterburn, line 25: If you dont mutate `randomNum` you can make it a `const`.

ui/common/Hasher.tsx
```diff
@@ -0,0 +1,154 @@
...
+    setHash(sjcl.codec.hex.fromBits(sjcl.hash.sha256.hash(randomNum)))
+    console.log(randomNum)
```
afterburn, line 27: Please remove `console.log`.

afterburn (changes requested):
Thanks @benalleng, requested some changes and asked for some explanations here and there.

ecurrencyhodler:
Tested. 

- [ ] The title is off on the top left. It should be "Zeros". And on the right i don't think there should be a lock icon when a player is doing lesson 1. 

Screenshot 2023-04-04 at 11 53 12 PM

- [ ] Also when a player completes a challenge, it should stick. In this video I completed the challenge but when I added another character, it went back to "hmm...that isn't right".

https://user-images.githubusercontent.com/32662508/229833403-915e9203-cff8-499d-9d84-f431488a4207.mov

- [ ] I think the expected behavior here is that when I click on trigonometry, it should be copied and I should be able to paste it. But I also want @GBKS to give it a design review. This design review shouldn't be a blocker to merging though. I can create a separate issue later if we are ready to merge and if christoph isn't back yet. 

https://user-images.githubusercontent.com/32662508/229834000-1d0460cc-f6f2-4bf1-a65e-1abdd00c5e2b.mov

- [ ] Also on this last part, I shouldn't be able to type in anything if I click the button to hash.

https://user-images.githubusercontent.com/32662508/229834169-0dfd5490-9b0b-45ff-b70d-3ce7dc39e042.mov

Finally, I waited a good 2 minutes and still didn't find anything. lol.

benalleng:
> I went ahead and just made the input lose focus once completed so that you can break it if you really want to

ui/lesson/HashChallenge/HashChallenge.tsx
benalleng: I intend for the hintTooltip to show after 15 seconds, with the timer starting only after the user begins to put in an input> I have made <s>some</s> many adjustments to the useEffect and I think I got it!

ui/common/Hasher.tsx
benalleng: ATM I have a small regex function that replaces that long mess. But I like your idea of implementing your helper function into a separate function to reuse

ecurrencyhodler:
Okay so for the first lesson, once the challenge is completed I'm able to type more but the challenge remains completed.

For the second part of lesson 1, text is locked and I'm unable to type more. 

IMO there should be consistency. I prefer the latter approach. 

Also for the triginomotry hint, I don't think we need to hide the answer. It's already hidden inside the hint which only gets displayed after someone is blocked.
Screenshot 2023-04-06 at 9 07 46 AM

benalleng:
I intended for all of the lessons to behave the same, they 'should' all behave the same and lose focus after completing the challenge and any more typing does not impact the next button

ecurrencyhodler:
Okay so it does look like the challenges do lock. But when I click to enter more characters, I'm able to continue. I think this is what we want. 

But the color of the hashes are different from pt1 and pt 2 when I continue typing in characters. 

Here's pt 1 where the hash stays green.

https://user-images.githubusercontent.com/32662508/230785326-d701802a-f4e0-43d8-846f-7517ef225f03.mov

Here's pt 2 where the hash goes white and then turns green when I pass the challenge again.

https://user-images.githubusercontent.com/32662508/230785330-7ffca7f5-0b5d-4eda-bb8d-27b7a19ce190.mov

benalleng:
> The title is off on the top left. It should be "Zeros". And on the right i don't think there should be a lock icon when a player is doing lesson 1.

This should now be fixed based on #329

Also the last part has been removed

ecurrencyhodler:
I tested it worked great!

One question I have is what exactly is the trigger for the hint to appear from pt.3 of Ch.2 L1?

>Do you think it should remain green even when on a wrong answer? IMO it shows the user how there are some other correct answers. what do you think?

As a player I think that would be confusing unless there's specific instructions/info telling me that's what's going on.

GBKS:
Nice work here. Just tested it and it works great overall, just needs a bit of tweaking.

For the mobile layout, we can avoid the whole screen height issue and simply let elements stack vertically with padding that give it the right amount of information density.

image

The hash on the bottom right seems to break lines earlier than it should.

One of the spacing issues in all the hashing challenges on desktop is that the text container at the bottom has top-padding applied, which makes it not-centered (it's probably only meant for mobile).

---

I like the idea of the hint. Would be nicer if it did not surprisingly interrupt the layout flow. In the mock below, it would use an absolute position, and maybe slowly fade in over several seconds.

image

---

Some other thoughts and questions:

- It seems a bit much to instantly show "That's not quite right" immediately after the user has pressed one key. Maybe we can introduce an intermediate ("Keep going...")?
- This is especially appropriate for "hashing 2", where the user just freely inputs things without a specific goal
- Should it be possible to enter spaces?
- What's the reason for a 24 character limit?
- Would it be possible to center the top text and input on desktop?
- The input field in "hashing 2" is too short and cuts off text. Probably because it's "inline-block". Could that be stretched wider?
- The input and hash letter-spacing is too much. In Figma it's defined as 5%, while in code it's set to 5px, the difference in unit is probably where the difference stems from
- Is the red background color being worked in a separate issue/PR?
 
---

Hope this is helpful, this comment is kind of a mess.

ecurrencyhodler:
Wow thanks christoph. 

> Is the red background color being worked in a separate issue/PR?

Yes in #360

benalleng:
> It seems a bit much to instantly show "That's not quite right" immediately after the user has pressed one key. Maybe we can introduce an intermediate ("Keep going...")
    
> Should it be possible to enter spaces?
I don't see why not, in this case specifically any character should be allowed as SHA256 is able to hash any data
    
> What's the reason for a 24 character limit?

this is sort of to prevent cheating on hashing-3 where the objective is to be difficult but simply holding down a certain key forever seems to kind of cheat it

I'll continue to work through your comments till i have it finished up!

afterburn:
@benalleng got some conflicts

benalleng:
I know there is an issue with my codebase, something broke my fonts loader. I have fixed this in my future PR's but it should not affect master when this is merged

benalleng:
@afterburn just following up with a check to see if we are good to merge this PR

content/lessons/chapter-2/hashing-1.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <p className="mt-8 text-lg md:text-xl">
+        {t('chapter_two.hashing_one.paragraph_four')}
+      </p>
+
+      <p className="mt-8 text-lg md:text-xl">
+        {t('chapter_two.hashing_one.paragraph_five')}
+      </p>
+
+      <Button
+        href={`${routes.chaptersUrl}/chapter-2/hashing-2`}
```
afterburn, line 39: Instead of using a href here are we not able to use the new `useSaveAndProceed` hook I made? Using that would make it so that once a user presses this next button their progress is being stored, that doesn't seem to be the case now.

content/lessons/chapter-2/hashing-3.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <p className="mt-8 text-lg md:text-xl">
+        {t('chapter_two.hashing_three.paragraph_two')}
+      </p>
+
+      <p className="mt-8 text-lg md:text-xl">
+        {t('chapter_two.hashing_three.paragraph_three')}
+      </p>
+
+      <Button
+        href={`${routes.chaptersUrl}/chapter-2/hashing-4`}
```
afterburn, line 45: Same here

content/lessons/chapter-2/hashing-5.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <p className="mt-2 text-center text-xl md:text-2xl">
+          {t('chapter_two.hashing_five.heading')}
+        </p>
+
+        <p className="mt-2.5 break-keep text-center text-lg md:text-xl">
+          {t('chapter_two.hashing_five.paragraph_one')}
+        </p>
+        <div className="flex justify-center">
+          <Button
+            href={`${routes.chaptersUrl}/chapter-2/hashing-6`}
```
afterburn, line 28: And here.

ui/lesson/HashChallenge.tsx
```diff
@@ -0,0 +1,183 @@
...
+          </div>
+        </div>
+      </div>
+
+      <StatusBar
+        alwaysShow
+        full
+        inProgressMessage="Keep going..."
+        input={success}
+        expected={'true'}
```
afterburn, line 178: @benalleng no curly braces needed for strings my friend

ui/lesson/HashChallenge.tsx
```diff
@@ -0,0 +1,183 @@
...
+}) {
+  const { saveProgress } = useProgressContext()
+  const lang = useLang()
+  const t = useTranslations(lang)
+  const pathName = usePathname()
+
+  const [input, setInput] = useState('')
+  const [userInput, setUserInput] = useState('')
+  const [hintTooltip, setHintTooltip] = useState(false)
+  const [success, setSuccess] = useState<string>('')
```
afterburn, line 42: This can be a bool

afterburn (changes requested):
@benalleng thanks, left some comments

benalleng:
I went ahead and reworked the statusBar, I made it accept two things, a success value and hints so that we can cover all the expected states. begin, inProgress, error, and success

afterburn (approved):

PR #346, opened 2023-04-02
Feedback Link

### LINKED ISSUE #343: Add help button and link to form
Let's add a help button on the top right of the screen like we had in previous designs. When a player clicks on the button, a separate tab opens and they're directed to [this google form](https://forms.gle/NVcg9ukPvUBYjw1u7). This will help us keep tabs on where people are getting stuck.

This feature request was born out of feedback we received during the demo where sometimes people would get stuck in the middle of the chapter but didn't know where to submit feedback about it.

benalleng:
Closes #343

Also cleans up some NavBar item classes to be more consistent

### DIFF
--- a/components/Navbar/HelpLink.tsx
+++ b/components/Navbar/HelpLink.tsx
@@ -1,0 +1,34 @@
+'use client'
+
+import { useLang, useTranslations } from 'hooks'
+import Link from 'next/link'
+import { Tooltip } from 'ui'
+
+export default function HelpLink() {
+  const lang = useLang()
+  const t = useTranslations(lang)
+  return (
+    <div className="flex-l flex h-full items-stretch">
+      <Tooltip
+        id={`navbar-tab-tooltip-3`}
+        position="bottom"
+        offset={0}
+        content={
+          <div className="flex">
+            <span className="whitespace-nowrap">
+              {t('navbar.help_tooltip')}
+            </span>
+          </div>
+        }
+      >
+        <Link
+          className="relative flex h-full w-12 cursor-pointer items-center justify-center border-r border-white/25 text-center text-2xl text-white transition duration-100 ease-in-out hover:bg-black/25 md:w-16"
+          target="_blank"
+          href="https://forms.gle/NVcg9ukPvUBYjw1u7"
+        >
+          ?
+        </Link>
+      </Tooltip>
+    </div>
+  )
+}
--- a/components/Navbar/UserButton.tsx
+++ b/components/Navbar/UserButton.tsx
@@ -13,35 +13,6 @@ export default function UserButton() {
   const [openSignInModal, setOpenSignInModal] = useState(false)
   const [openSignUpModal, setOpenSignUpModal] = useState(false)
   const { user, isLoggedIn, isRegistered } = useUser()
-  const { account, isLoading } = useAuthContext()
-
-  if (!isLoading) {
-    console.log(account)
-  }
-
-  const renderIconButton = () => {
-    if (!hasMounted) {
-      return <div />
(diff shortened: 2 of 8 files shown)


ecurrencyhodler:
One nit is that we should probably change "get stuck?" to "need help?"

I think it translates easier to other languages. Other than that LGTM!

components/Navbar/HelpLink.tsx
```diff
@@ -0,0 +1,29 @@
+'use client'
+
+import Link from 'next/link'
+import { Tooltip } from 'ui'
+
+export default function UserButton() {
```
afterburn, line 6: File is called HelpLink, but function is called UserButton

components/Navbar/HelpLink.tsx
```diff
@@ -0,0 +1,29 @@
...
+        offset={0}
+        content={
+          <div className="flex">
+            <span className="whitespace-nowrap">Get stuck?</span>
+          </div>
+        }
+      >
+        <Link
+          className="relative flex h-full w-12 cursor-pointer items-center justify-center border-r border-white/25 text-center text-2xl text-white transition duration-100 ease-in-out hover:bg-black/25 md:w-16"
+          target={'_blank'}
```
afterburn, line 21: You don't need curly braces if all you are passing is a string

components/Navbar/HelpLink.tsx
```diff
@@ -0,0 +1,29 @@
...
+export default function UserButton() {
+  return (
+    <div className="flex-l flex h-full items-stretch">
+      <Tooltip
+        id={`navbar-tab-tooltip-3`}
+        position="bottom"
+        offset={0}
+        content={
+          <div className="flex">
+            <span className="whitespace-nowrap">Get stuck?</span>
```
afterburn, line 15: Missing translation

components/Navbar/UserButton.tsx
```diff
@@ -50,8 +52,11 @@ export default function UserButton() {
   }
 
   return (
-    <div>
-      <div className="ml-4 h-10 w-10">{renderIconButton()}</div>
+    <div className="flex-l flex h-full items-stretch">
+      <span aria-describedby="navbar-tab-user">
+        <span>{renderIconButton()}</span>
```
afterburn, line 57: Lets try to stay away from these types of render functions, I feel they make it difficult to grasp the overall html structure at a glance for other contributors.

I would also strongly suggest to keep the markup separate from the component's logic. In this instance I would classify `toggleModal` and `hideModal` as component logic, whereas `renderIconButton` is markup so it should come after `toggleModal` and `hideModal`. Ideally we just inline it like this though:

```jsx
<span aria-describedby="navbar-tab-user">
  {!hasMounted && <div />}
  {hasMounted && isLoggedIn && (
    <button
      id="navbar-tab-user-icon"
      onClick={() => setOpenSignInModal(true)}
      aria-label="profile"
      className="text-grey-300 relative flex h-full w-12 cursor-pointer items-center justify-center md:w-16"
    >
      <Avatar avatar={user.avatar} size={30} />
    </button>
  )}
  {hasMounted && !isLoggedIn && (
    <button
      id="navbar-tab-user-login"
      onClick={() => toggleModal(true)}
      aria-label="profile"
      className="text-grey-300 relative flex h-full w-12 cursor-pointer items-center justify-center md:w-16"
    >
      <UserIcon />
    </button>
  )}
</span>
```

afterburn (changes requested):
Thanks @benalleng, requested a couple minor changes.

ecurrencyhodler (approved):
LGTM

Issue #323, opened 2023-03-21
Ch.2 L3: Mining challenge


afterburn:
In Chapter 2 Challenge 3 users will need to perform mining. We need to implement a UI that let's users start and stop mining, it should also display some statistics on the right (see [Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=2873-29990&t=QiZtLSzbb96QH3aE-0)).

My current train of thought is that we don't actually want to generate hashes until one is found that starts with five zero's, this is resource intensive and people might get the impression that we're actually trying to mine bitcoin for ourselves using their machine. I think it makes more sense to simulate the mining process instead.

This `<Mining>` component needs to periodically update the statistics to give the impression that something is happening, the numbers can be random as long as they make logical sense. After a fixed time has passed we can show a hardcoded hash starting with five zero's to the user.

There are a couple of elements in this `<Mining>` component that lend themselves well to reusability:
- The progress bar should become a standalone component.

<img width="644" alt="image" src="https://user-images.githubusercontent.com/10757768/226626166-a278a5b7-d1bf-449c-80fa-058fef14f564.png">

- The stats card should become a standalone component.

![image](https://user-images.githubusercontent.com/10757768/226625882-d4d5c347-b165-4244-9461-097451727fdb.png)

ecurrencyhodler:
@GBKS after our call on tuesday, we decided that vinayak would just go ahead and choose an arbitrary amount of time for a player to wait before they complete the mining challenge. This is to prevent the player from getting stuck mining for hours. Kevin will do the same for Lesson 2.

When we test Lesson 2 and Lesson 3 next week, we can see if this amount of time needs to be adjusted for both challenges.

GBKS:
That makes sense. Within that time span, we may still want to introduce some randomness. For example, if we mine 1,000 blocks, to would feel more "realistic" if the time between mined blocks is randomized, and not consistent. We could also randomize the number of transactions confirmed per block.

PR #349, opened 2023-04-04
Mining Challenge

vinayaksh42:
Fixes: #323 

Changes:
- ProgressBar UI component
- InfoBox UI component
- structure for mining challenge

[preview](https://saving-satoshi-git-fork-vinayaksh42-issue323-savingsatoshi.vercel.app/en/chapters/chapter-2/mining-1?dev=true)

To-Do:
- Plan and implement the duration of different mining sessions (https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=2019-25736&t=MqfrdugwEKfkRuMC-0)
- ~~Chapter 2 theme~~


ui/common/InfoBox.tsx
```diff
@@ -0,0 +1,29 @@
+'use client'
+
+function addCommas(number) {
```
GBKS, line 3: Browsers have [a built-in API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) you can use for this.

vinayaksh42: thanks for the info about this built-in API, I will add it to the code

GBKS:
Just tried it out. Great start, super helpful to see this. There are definitely a bunch of details we can tweak to really make this come to life. I'll try to provide more detailed feedback this week.

How do you like it? Any changes you'd make?

vinayaksh42:
> Just tried it out. Great start, super helpful to see this. There are definitely a bunch of details we can tweak to really make this come to life. I'll try to provide more detailed feedback this week.
> 
> How do you like it? Any changes you'd make?

the first mining session feels to be a little longer than it should be, either we can add extra explanation text so that the user is engaged in reading or we can reduce the time to 45 seconds or 1 minute. I also noticed that the numbers don't match up with the text on the left, for example, it says something about 2,000,000 transactions, and the mockup has 287,000,000 transactions. The timing and the numbers will surely require some adjustments.

ui/common/InfoBox.tsx
```diff
@@ -0,0 +1,28 @@
+'use client'
```
afterburn, line 1: Rename to something like <Statistic>

ui/common/ProgressBar.tsx
```diff
@@ -0,0 +1,30 @@
+'use client'
```
afterburn, line 1: I created a progress bar [here](https://codesandbox.io/s/progress-bar-7yheeh?file=/src/App.js) that has gpu animation, lets use that.

content/lessons/chapter-2/mining-1.tsx
```diff
@@ -0,0 +1,326 @@
...
+    )
+    setTimeout(() => {
+      setStep(4)
+      setBlocks(1000)
+      setNonceHighlight(true)
+      setHashPowerHighlight(true)
+    }, time)
+  }
+
+  const stepOne = (
```
afterburn, line 138: Lets inline these steps in the returned jsx

afterburn (changes requested):
Hey @vinayaksh42 thanks for working on this issue, I requested some changes.

ui/common/ProgressBar.tsx
vinayaksh42: @afterburn According to the new design for the mining challenge, I think we might have to rework the progress bar. figma mockup - https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=3427-35637&t=L7QZCyGO0OpLQyT5-0

afterburn: Yeah thats fine, just make sure you use the setup I provided in the codesandbox but feel free to modify it to match the design. Just want to make sure all our animations run on the gpu.

ecurrencyhodler:
Few notes after testing:
- I absolutely love the ui and animations. Seeing the #'s run while reading about what's happening was a powerful experience for me.
- pt. 1 of the mining challenge I think is too long. We should make it faster. Maybe half the amount of time.
- pt. 2 of the mining challenge of normal mining is also too long. We should cut it down to 30 seconds or so before the option to 10x appears. 
- The design is also a bit off. The text here is too close to each other
Screenshot 2023-05-05 at 12 51 40 PM

- I got stuck on the last part of the challenge. Nothing happens when I click "continue." See video.

https://user-images.githubusercontent.com/32662508/236557649-90058b9d-af9a-45d7-99d2-6cb6b8e39ba9.mov

Lastly, @GBKS what is "X" signal in this section?
Screenshot 2023-05-05 at 12 50 14 PM

vinayaksh42:
based on feedback received from @ecurrencyhodler 
1. Part-1 can be reduced to 30 or 45 seconds.
2. Part-2 (before 10x) is already set to 30 seconds so maybe we should reduce it to 15 or 20 seconds.
3. for the text we can reduce the font size  (@GBKS suggestions needed for this issue)
4. The continue will be linked to the final outro screen of chapter-2. We can resolve this in a separate PR as well.

ecurrencyhodler:
I like 15 seconds for pt. 2.

I actually don't know if anyone has worked on the outro yet. I don't think it has. I'll create an issue.

ecurrencyhodler:
@vinayaksh42 for the text size. When I look at the design the table size is actually wider than what's in this PR. I think that might be better fix so it's more closely in line with figma.

This is the PR
Screenshot 2023-05-06 at 3 08 34 PM

This is Figma
Screenshot 2023-05-06 at 3 08 53 PM

ecurrencyhodler:
It looks better than before but still is a bit crowded. I don't want the font size to be a blocker. We can address this in a follow issue/pr.
Screenshot 2023-05-07 at 5 37 24 PM


Also I waited for the 10x button in chapter-2/mining-1 and it was longer than 15 seconds. Did you update that part or not yet?

GBKS:
Awesome progress. I just took screenshots and annotated them in Figma [here](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=1194%3A14198&t=040If94OG01Ai74a-1). Still a few details to adjust.

Regarding font size, if you match the designs, the layout works on small phones (iPhone SE with 375px width). This PR currently does not (wrong font, font size is 16 but should be 15, left/right padding is 24 but should be 15, etc). Please adjust and then we can see if it's still problematic. This is where the details really matter and add up.

@ecurrencyhodler the "X" is an indicator whether the generated hash does start with 10 zeroes. It turns to a green check when a good hash was found. There is a bit missing that should make this more obvious. When we're mining all the 1,000 blocks, this should flicker to the green check whenever a block was found. That would be 999 times in this short time span, which is a lot. We need to fudge this a bit to get the effect we want (a very quick flickering between red x and green check). I also added a comment in Figma to add a tooltip (not a visual tooltip, just a "title" attribute) so when a user hovers the x/check, they get an explanation what it means. Once this is in, let's test again and see if it feels right.

benalleng:
Thanks @sabakhilji This PR only addresses the mining challenge, lets move any non-related issues you find through QA to the issues board. I have created a list here in #431 and you are welcome to add your findings there so we have an area we can all look at to track QA progress

ui/common/MiningStatisticsHash.tsx
```diff
@@ -0,0 +1,418 @@
...
+            '!bg-[#FBEBC6] shadow-[0_0px_8px_rgba(251,235,198,0.5)]':
+              hashPower > 39000,
+          })}
+        ></div>
+        <div
+          className={clsx('h-[15px] w-[5px] rounded-full bg-black/25', {
+            '!bg-[#FBEBC6] shadow-[0_0px_8px_rgba(251,235,198,0.5)]':
+              hashPower > 40000,
+          })}
+        ></div>
```
shaavan, line 414: I was analyzing the code and noticed this section of repetitive code. 

Using `.map()` will be a good idea here.

You can define a const array of numbers at the beginning of the code.

```js
const hashDiff = Array.from({length: 40}, (_, i) => i + 1)
```

And use it to map these divs.

```suggestion
    {hashDiff.map((i, index) => {
          return (
            <div key={index}
              className={clsx('h-[15px] w-[5px] rounded-full bg-black/25', {
                '!bg-[#FBEBC6] shadow-[0_0px_8px_rgba(251,235,198,0.5)]':
                  hashPower > i * 1000,
              })}
            ></div>
          )
        })}
```

shaavan (commented):
The Mining challenge looks beautiful, and all the subtle animations add depth to the experience.

Fantastic work to both @GBKS and @vinayaksh42 for making this possible!

I am going through the code changes, and it will take me a little while to thoroughly go through it. For now, I have one suggestion to improve the brevity of the code.

ui/common/MiningStatisticsHash.tsx
vinayaksh42: Thanks for the suggestion. Using a map will be more optimized and will require less work if in the future we want to change the specification of the loading bar/ progress bar. I will try to convert it into a reusable component

tailwind.config.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
           }
           return `rgb(var(--green))`
         },
+        red: '#BC0E13',
```
shaavan, line 33: I think a better approach would be to define this color in `styles/global.css` and use it to define color the way `back` and `green` are defined preceding it.

This will allow us to use an opacity filter along with color in case we need it, and will also maintain symmetry of how we have defined custom colors till now.

shaavan (commented):
Code Review ACK modulo small nit.


@vinayaksh42 @GBKS 

I have one question on the timing of enabling of `10x` button.

- The first hash-run takes a long time to run with no visual clue on the progress or how long it will take to mine one block.
- This makes sense to show the time and energy required to mine a block.
- However, on a second run-through, when we task ourselves to mine 1000 blocks, the process repeats, and it takes a little too long for the `10x` button to be activated. I could sense a feeling of impatience waiting for the 10x button to be enabled.
- We should experiment with a slightly lesser time difference for the 10x button to be enabled.

ecurrencyhodler:
Two small copy nits. Other than that it's a tACK.

"It really helped to have that boost in hash power i.e. the number of hashes your computer can try in a second."
Screenshot 2023-05-12 at 9 38 50 AM

IMO we should replace "chip" with "machine". 
Screenshot 2023-05-12 at 9 34 22 AM

tailwind.config.js
vinayaksh42: nice catch, I will add all the custom colors used in this challenge to tailwind config file as they might be re-used in future challenges.

lib/formatHash.tsx
```diff
@@ -24,7 +26,16 @@ export default function formatHash(
...
             return (
-              <span key={idx} className="text-white">
+              <span
+                key={idx}
+                className={
+                  hashStatus
+                    ? hashStatus === 1
+                      ? 'text-[#EDA081]'
+                      : 'text-white'
+                    : 'text-[#5DBC59]'
```
afterburn, line 36: How about we use `clsx` for this, something like
```ts
clsx({
  'text-[#5DBC48]': !hashStatus,
  'text-[#EDA081]': hashStatus && hashStatus === 1,
  'text-white': hashSstatus && hashStatus !== 1
})
```

shared/icons/Cross.tsx
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <svg
+      width="15"
+      height="15"
+      viewBox="0 0 15 15"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      className={className}
+    >
+      <title>{title}</title>
```
afterburn, line 11: Strikes me as odd that a SVG icon should have a title, what is this being used for?

Also, the idea behind this icons folder is that each icon can be colored using `text-<color>`, for that you would need to set stroke to currentColor i.e. `stroke="currentColor"`.

ui/common/ProgressBar.tsx
```diff
@@ -0,0 +1,21 @@
...
+    <div className="relative h-[30px] w-full rounded-[5px]  bg-black/20">
+      <div
+        className={clsx(
+          'absolute top-0 left-0 h-full transform-gpu rounded-[5px] bg-[#FBEBC6] transition-transform',
+          {
+            '!bg-[#5DBC59]': progress === 100,
+          }
+        )}
+        style={{
+          width: `${progress <= 0.5 && progress > 0 ? 0.5 : progress}%`,
```
afterburn, line 16: Earlier in this PR I posted a link to a codesandbox that has a progress bar which animates using transforms which do not cause paints when `progress` changes, modifying the `width` does cause paints. If you can, please try to use transforms only to make full use of the GPU.

You can turn this on in your devtools to see whether your animation causes any paints:
image

And you can learn more about this subject [here](https://web.dev/animations-overview/) and [here](https://web.dev/animations-guide/).

afterburn (changes requested):
Hey @vinayaksh42 good effort, requested some changes after which I think we can merge. Feel free to pick up the gpu animation one in a separate PR.

shared/icons/Cross.tsx
vinayaksh42: the title attribute is being used for providing an explanation to the cross icon - further explanation - https://github.com/saving-satoshi/saving-satoshi/pull/349#issuecomment-1537974549

ui/common/ProgressBar.tsx
vinayaksh42: based on that sandbox I added `transform-gpu` and `transition-transform`. I tried out the Paint flashing devtool in Chrome and it still displays the progress bar with green highlights.

vinayaksh42:
> Hey @vinayaksh42 good effort, requested some changes after which I think we can merge. Feel free to pick up the gpu animation one in a separate PR.

I have addressed 2 out of 3 changes. We should create a new issue for GPU animation issue and for adjusting the mining time duration.

ui/common/ProgressBar.tsx
afterburn: That's probably due to rerendering of the component itself, lgtm for now, I can also have a look later

afterburn (approved):

GBKS:
Just did another review and here's my feedback.

First, shorten time of step 1 to 15 seconds and step 3 to 30 seconds. This has been requested by several people, please make this change.

Second, I noticed that the small hash is displayed in columns - should be rows. See image below.

image

A couple of polish things that I already brought up previous:
- When the text on the left changes, please fade it out and the new text then in
- At the end, when boxes get their highlights, please fade those highlights in and out

Something we talked about in the Tuesday call is that step 2 is not realistic. The user has the impression that they are still mining, but the UI never finds any blocks. It's basically running in idle mode. They should still find blocks, and the 10x button should literally just make that process 10 times faster. This could be a follow-up PR.

To move forward, I think only the first point here (shortening the time) is important for this PR. The other feedback can be addressed in a follow-up PR.

Almost there. Thanks for tackling this big one.

vinayaksh42:
- @GBKS and @afterburn I have reduced the mining time for the first mining session from 25 seconds to 15 seconds. The 3rd mining session is actually running based on when all the blocks are found and right now it is under 30 seconds even tho in the code it is 60 seconds but the mining stops when it hits 1000 blocks.

As for the other changes:
- Changing the sequence of hash from column to row will require changes to `lib/formatHash.tsx`which is also being used in one of the other challenge made by @benalleng . we should move this change to a new issue/PR

I am creating a new issue for the following changes:

- Text fade in and fade out
- Box highlight fade in and fade out
- Making the 2nd mining session realistic


@afterburn I think this PR is now ready for a final code review. The only change made since the last time is very small tho.

vinayaksh42:
I have opened the issue - https://github.com/saving-satoshi/saving-satoshi/issues/454 for further improving the UX of this challenge

tailwind.config.js
shaavan: This comment was left to be resolved in the subsequent updates of PR.

Since it's not a blocker, let's take this up with #454.

shaavan (approved):
ACK

Amazing job, @vinayaksh42!! 💯

I have added a remark for one of the unresolved comments [here](https://github.com/saving-satoshi/saving-satoshi/pull/349#discussion_r1200294949).
Since it's a non-blocker, we can take that up later in development.

GBKS:
Awesome, let's get this thing merged in 🪩🥳🎉

afterburn (approved):

Issue #281, opened 2023-03-11
Handle 404 redirects outside of chapters


ecurrencyhodler:
When someone lands on a page that can't be found, they are shown this image. We should have a friendly redirect or display a landing page that points them back to savingsatoshi.com.
<img width="1190" alt="Screenshot 2023-03-11 at 10 17 03 AM" src="https://user-images.githubusercontent.com/32662508/224505095-fd47c68a-c62b-4761-ac6b-e036f79b3e4d.png">

It looks like we have something in place if a user lands in a url in the chapter section:
<img width="1193" alt="Screenshot 2023-03-11 at 10 19 15 AM" src="https://user-images.githubusercontent.com/32662508/224505140-62ee77cc-b201-4431-9dcb-80c457cde263.png">

First we need to discuss how we want to handle 404 redirects. Then we can send the issue to design if needed.

This is not necessary for the demo.

afterburn:
We can have a catch-all 404 page instead of handling it on a per page basis, maybe @GBKS can make a fun design.

Something like this:
![Copy-of-Copy-of-Copy-of-SVG-in-HTML-10.png](https://user-images.githubusercontent.com/10757768/224505561-81bfcd95-1122-47bc-9563-7cc8b9c9567f.png)

But related to Bitcoin

ecurrencyhodler:
Holocat yelling at them would be pretty funny.

GBKS:
Here's an old error page from the times of the Legends of Lightning pitch.

![image](https://user-images.githubusercontent.com/695901/224974597-61bb0922-b8e4-4a06-89b8-a89ceecda004.png)

sabakhilji:
Hey is this issue open can I work on it?

PR #355, opened 2023-04-04
Handle 404 redirects outside of chapters

sabakhilji:
Handle 404 redirects outside of chapters #281


benalleng:
For this deployment to work you are going to need to remove the `package-lock.json` and `yarn.lock` files from being tracked. 
[Here](https://stackoverflow.com/questions/38743912/remove-a-file-from-a-git-pull-request) is a quick stackoverflow helping out with this

Issue #319, opened 2023-03-21
Implement new Account system


afterburn:
**Current situation**
We're currently using LocalStorage to store and retrieve user account data.

**Desired situation**
We want to use `https://api.savingsatoshi.com`'s endpoints to save/retrieve account data.

For more information about how to use the backend, see [this PR](https://github.com/saving-satoshi/saving-satoshi/pull/330).

afterburn:
Closed in https://github.com/saving-satoshi/saving-satoshi/pull/357

Issue #158, opened 2023-02-17
Refactor Progress


afterburn:
**Describe the enhancement**
Replace all the logic pertaining to user progress with endpoints from Backend V1.

Issue #331, opened 2023-03-29
User login


afterburn:
Right now the same user will always be logged in because in `/providers/AuthProvider.tsx` I have a hardcoded `private_key`, we should obviously pass the user's `private_key` instead.

I think @GBKS and whoever has a strong opinion on the matter should think about the best way to approach the login flow from a user's perspective. After creating an account users will need a way to enter their `private_key` to sign in. We currently have the modal below which users can use to restore their account, but with the new backend it's no longer possible to retrieve `private_key`'s from local storage, this means we cannot login users automatically after their access_token expires, and so the question is whether the modal below is the one we want to present to our users for logging in again, or whether we want a different way for users to enter their `private_key` to 'log in'.

<img width="422" alt="image" src="https://user-images.githubusercontent.com/10757768/228575707-d8944835-9b51-4c3c-8644-c3a4e3ac70d2.png">

GBKS:
This still looks good to me.

ecurrencyhodler:
Closing this issue for now. We can revisit in later milestones.

ecurrencyhodler:
Re-opening this issue. Kevin will provide more context.

afterburn:
With our new backend system the account creations process goes like this:

1. User navigates to savingsatoshi.com
2. User clicks on user button, the "Sign in" modal pops up:

![image](https://user-images.githubusercontent.com/10757768/231211449-905411e3-2c15-4521-b6d9-55b4f226cb41.png)
3. The user has no account yet, so he/she clicks on "Dont have an account?", which opens the "Sign Up" modal:

![image](https://user-images.githubusercontent.com/10757768/231211886-85e8a33e-7765-40af-ad26-69aa7ecf00ee.png)
![image](https://user-images.githubusercontent.com/10757768/231211965-1945ecbe-cd1d-4da1-804e-356d00e57b34.png)

4. User creates an account by:
  a.  Choosing an avatar
  b. Either generate a private key (like we did before) and copy + save it, or enter a private key themselves.
5. User presses "Create account" button, which sends the data that was input to the backend, which creates a record in the DB.
6. The user clicks on his/her avatar, which opens the "Account" modal:
![image](https://user-images.githubusercontent.com/10757768/231212248-4daf5afd-364a-40e6-a088-3d85954da58e.png)
This allows users to logout, and in the future perhaps manage account settings (change avatar, etc).

[...]

PR #357, opened 2023-04-06
Account & Progress system

### LINKED ISSUE #139: Refactor progress tracking mechanism
**Describe the enhancement**
We should refactor _when_ we update user progress. Currently, we only save user progress when they successfully complete a Terminal challenge (after validating their input) or when they reach the end of a chapter.

I think this setup is flawed because contributors might not want to use the Terminal challenge template, especially for more complex challenges. Ideally we move the progress tracking to a higher level so that the content does not have to deal with it directly, but I think we'll always need a way to signal from within the content that a users progress status has changed.

Instead of our current approach (`setUserProgress`) I propose we use React context instead. An example of how this could look in our content:

```jsx
export default function Genesis2() {
  const { setLessonStatus } = useProgressContext()
  
  const onChange = (input) => {
    if (givenInput === expectedInput) {
      setLessonStatus(lessonId, Status.Complete)
    }
  }
  
  ...
}
```

The benefit of React context over our current approach is that we can listen for changes from anywhere in our app, so the Tabs in the Navbar will immediately update once `setLessonStatus` is called.

afterburn:
This PR closes #139, #158 and #319.


afterburn:
@benalleng @ecurrencyhodler please test the functionality, I have [an issue](https://github.com/saving-satoshi/saving-satoshi/issues/331) open for the design and plan to address that in a follow up PR depending on how fast the design issues have been resolved.

ecurrencyhodler:
If I understand this screen correctly, if someone enters in a personally generated privkey an account will get created. Even if that is the case, I don't think we should have copy that tells them that. 

To make it simpler, we should just say something like: "Enter your private key to below to sign into your account and restore your progress".
Screenshot 2023-04-11 at 9 23 45 AM

Then if they don't, they can click the "Don't have an account" link.

afterburn:
> If I understand this screen correctly, if someone enters in a personally generated privkey an account will get created. Even if that is the case, I don't think we should have copy that tells them that.
> 
> To make it simpler, we should just say something like: "Enter your private key to below to sign into your account and restore your progress".
> 
> Then if they don't, they can click the "Don't have an account" link.

@ecurrencyhodler You can only enter a registered `private_key` into the "Sign In" modal, so one that is known in our database.

The "Sign Up" modal allows you to create an account (aka register a `private_key`) by either letting us generate a `private_key` for you, or you can enter one yourself. That `private_key` can then be used to sign in via the "Sign In" modal

app/[lang]/chapters/[slug]/[lesson]/page.tsx
```diff
@@ -23,5 +73,47 @@ export default function Page({ params }) {
...
+  if (!account || !progress) {
+    return (
+      <div className="flex h-full w-full grow flex-col items-center justify-center">
+        <span className="mb-10 font-nunito text-2xl text-white">
+          Sorry, you need to be signed in to access this lesson
+        </span>
+        <Button onClick={handleSignInClick} size="small">
+          Sign in
+        </Button>
+      </div>
```
Comment (benalleng) [inline app/[lang]/chapters/[slug]/[lesson]/page.tsx:99]: Assuming the user has to sign in to proceed can we just render the `Modal.SignIn` here to save a click instead of prompting them?

components/Modals/SignUp.tsx
```diff
@@ -4,40 +4,59 @@ import Avatar from 'components/Avatar'
...
-    saveLocally()
-    setUserRegistered()
-    onConfirm()
+  async function handleConfirm() {
+    try {
+      setLoading(true)
+
+      await register(privateKey, `/assets/avatars/${avatar}.png`)
+
+      onClose()
```
benalleng, line 45: With this happening on every create account button press is there a concern about account generation spam? 
Even just trying it out I think I made several accounts as I didn't copy my private key the first time

components/Modals/SignUp.tsx
```diff
@@ -80,21 +99,53 @@ export default function SignUpModal({ open, onClose, onConfirm }) {
...
+            <p className="mt-5 text-base">{t('modal_signup.input')}</p>
+          </>
+        )}
 
-        <p className="mt-5 text-base">{t('modal_signup.paragraph_two')}</p>
         <button
-          onClick={confirm}
-          className="mt-4 w-full rounded-md border border-white px-4 py-2 text-xl text-white"
+          onClick={handleConfirm}
+          disabled={loading}
```
benalleng, line 144: Perhaps we should add a simple "I have backed up my private key checkbox" before allowing the user to proceed might help.

Also worth considering just immediately signing them in after this step. If they don't write down their private key they will still have to restart once they are logged out but worth considering

ui/chapter/ChallengeListItem.tsx
```diff
@@ -27,17 +33,17 @@ export default function ChallengeItem({
...
-          'hover:bg-black/20': status && status.unlocked,
+          'bg-black/15': isUnlocked && !isCompleted,
+          'hover:bg-black/20': isUnlocked,
         }
       )}
     >
       <span className="pr-1 opacity-50">{position + '. '}</span>
       {t(title)}
-      {status && !status.unlocked && (
+      {isUnlocked && (
```
benalleng, line 43: should this be?
``` suggestion
      {!isUnlocked && (
```

ui/chapter/End.tsx
```diff
@@ -54,7 +24,7 @@ export default function End({
       <div className="left-unset absolute bottom-0 ml-auto w-full bg-gradient-to-b from-transparent via-[#00000040] to-[#00000080] p-4 pb-12 md:p-16">
         <div className="md:max-w-[500px]">
           <h1 className="text-3xl font-bold text-white md:text-5xl">
-            {t(title)}
+            {t(title)}z
```
benalleng, line 27: typo
``` suggestion
            {t(title)}
```

benalleng (changes requested):
Works great! 
I tried my best to focus on functionality and most of my comments are small or questions about how we want it to be

Also i think we should think about where the signup prompt should be ultimately. Is having it as a far back as the first lesson page the right spot for it?

components/Modals/SignUp.tsx
afterburn: Yeah that's a valid concern, something I wanted to save for a follow up PR but a few things we could do are:
- Add rate limiting on backend
- Add a captcha

components/Modals/SignUp.tsx
afterburn: I think the checkbox is a good idea, I'll add that. In general your second proposal is often considered a security risk but since we're not dealing with things like email verification I think we can do it.

afterburn:
> Also i think we should think about where the signup prompt should be ultimately. Is having it as a far back as the first lesson page the right spot for it?

@benalleng, the problem is with the new system we can no longer store things client-side. So if a user gets to ch1 transacting-3 and navigates away, all progress is lost. We can prevent that from happening by doing a save whenever a lesson is completed, so when the user refreshes his/her latest progress is loaded again.

To be able to save progress after each lesson we need to have the user be signed in though, so it's a tradeoff. Alternatively we could store some progress in localStorage, then once the end of chapter 1 is hit we send all that progress to the backend, but this is more complex and time-consuming to implement.

I personally lean more towards having people sign in before they can start the game, similar to how other learning platforms (and games) do it, but I can understand we might want to "lure" people in a bit before having them register their private key. 

I'm curious to learn what others think, I can bring it up on tuesday and modify the behavior after we reach consensus.

benalleng (commented):
I received an error (has happened more than once but very rarely)

image

Request object:
```
XHRPOSThttps://api.savingsatoshi.com/v1/auth/register
[HTTP/1.1 400 Bad Request 30ms]

	
avatar	"/assets/avatars/4.png"
private_key	"a0d5a5162eead0384c38821fda5794ef4031de7cd3d976827c1c83bafa675fb"
```

Private key length: 63

This was from the create account button. 

not sure what is causing this problem as you clearly state the private key length as 32bytes

benalleng:
> lean more towards having people sign in before they can start the lessons

I was actually thinking even earlier if that wasn't clear, like pushing the user to login before going past the homepage

lib/crypto.ts
```diff
@@ -0,0 +1,32 @@
...
+  let publicKey = null
+  let privateKey = null
+
+  do {
+    const privateKeyBuf = crypto.randomBytes(32)
+    privateKey = Secp256k1.uint256(privateKeyBuf, 16)
+    publicKey = Secp256k1.generatePublicKeyFromPrivateKeyData(privateKey)
+  } while (
+    !publicKey ||
+    privateKey.gte(
```
benalleng, line 16: Is this supposed to be `.get()`?

benalleng: Don't think this is causing the 31 bytes error but trying look see why that might happen

lib/crypto.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { KeyPair } from 'types'
+
+export function generateKeypair(): KeyPair {
+  let publicKey = null
+  let privateKey = null
+
+  do {
+    const privateKeyBuf = crypto.randomBytes(32)
+    privateKey = Secp256k1.uint256(privateKeyBuf, 16)
```
benalleng, line 12: Any chance this cuts off leading zeros to cause the length problem?

lib/crypto.ts
afterburn: Nope, gte is correct, I copied this from our existing `lib/user.ts` file which has been removed in this PR. I think it stands for greater than or equal to. I'll have a look at this code again because I think it can be simplified.

lib/crypto.ts
afterburn: I refactored this file, we're now using [paulmillr/noble-secp256k1](https://github.com/paulmillr/noble-secp256k1) instead of [lionello/secp256k1-js](https://github.com/lionello/secp256k1-js) which seems to be simpler to use and seems a little "established". Please test again when you have the time.

benalleng (approved):
Tested!
I Can no longer reproduce the error with private key length
The useEffect error has also been solved.
It all LGTM!

PR #366, opened 2023-04-10
Progress tracking and story inclusion

### LINKED ISSUE #296: UX: User can skip/miss the story if they use the challenges tab
**Demo QA**

**Environment**: Linux (Ubuntu), Chrome (private browser, so cache is totally cleared)

**Background**: On the [chapters page](https://savingsatoshi.com/en/chapters), a new user has two ways of starting their journey:

- they can click "Start chapter 1" which takes them to https://savingsatoshi.com/en/chapters/chapter-1/intro-1, and they get the backstory
- they can click "Challenges", then "Genesis", which takes them right to the challenge: https://savingsatoshi.com/en/chapters/chapter-1/genesis-1

**Issue**: I think this is more of a UX question. Do we want this second flow to be possible? For the user to bypass the story screens?

**Possible solution**: If the user has not completed the Genesis challenge (if they don't have a check mark next to it in the list of challenges), clicking the challenge will take them to https://savingsatoshi.com/en/chapters/chapter-1/intro-1 . If they have completed the challenge, the current behavior is fine, and they should be taken straight to the challenge, https://savingsatoshi.com/en/chapters/chapter-1/genesis-1

I imagine this is could be a good first issue if we decide to implement the possible solution that is proposed.

### LINKED ISSUE #335: Background color for lessons on top right is different during story intro
When a player is in ch.1 lesson 1 during the story part, the background color on the top right is not dark. However when the player gets to the actual lesson content, it becomes dark. Is this intentional?

*During story part of the lesson*
Screenshot 2023-03-30 at 7 24 40 AM

*During the actual lesson content*
Screenshot 2023-03-30 at 7 29 47 AM

benalleng:
closes #335, Closes #296

This allows the navbar to properly track where the user is in the chapter. This PR treats the intro pages as part of the first challenge in each chapter in this case (genesis-1). Additionally this means clicking on the chapter-list-items on the chapters page or the challenge tabs no longer skip the story This also fixes some progress tracking errors

shoutout to @shaavan for the work he did on `lib/progress.ts` to get the logic working!!

This also restores outro-2 which we can use to have people fill out the feedback until we finish Chapter 2


afterburn:
Hey @benalleng unfortunately my shitty internet connection made me miss out on your explanation on what issues this PR solves, would you mind posting that here?

benalleng:
@afterburn I went ahead and detailed this change, additionally i went ahead and made similar comments to all of my open PR's

afterburn (approved):
Ah I get it now, thanks for the explanation.

benalleng:
I'll take a closer look at the logic to see if we can leave them be intro-1/2

ecurrencyhodler:
tACK! Looks we need a review from @afterburn

components/Navbar/NavbarDesktop/Tab.tsx
```diff
@@ -39,14 +39,33 @@ export default function Tab({
...
     progress,
     getLessonKey(slug, challenge.lessonId)
   )
 
-  const challengeId = isRouteLesson ? pathData.pop().split('-')[0] : undefined
+  const challengeId = isRouteLesson
+    ? pathData
+        .pop()
+        .split('-')[0]
+        .replace('intro', chapters[slug].metadata.challenges[0].split('-')[0])
```
afterburn, line 60: What is happening here? It seems to take intro-1, and replaces `intro` with the chapter's first challenge lessonId so it would become something like `scripting-1`?

benalleng: This allows the intro pages to be treated as challenges by the isActive variable, while not moving the intro pages to be a part of a specific lesson

afterburn: Okay, feels a bit hacky but its fine, we can always clean up later.

afterburn (approved):

PR #367, opened 2023-04-10
Scripting challenge mobile view

### LINKED ISSUE #341: Implement mobile version of ScriptingChallenge
I haven't paid any attention to the mobile version of the `ScriptingChallenge` so presumably it will need some work to get it to match the design [here](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=3013-28707&t=EBUyGkVSv51VyXPo-4). 

@GBKS asked me for some early feedback and one thing I think the design still needs is a Run button on the code tab itself, I'll assign this ticket to @GBKS and when he unassigns the issue that means its free for someone to work on as the design will then be finalized.

benalleng:
Closes #341 

[Preview](https://saving-satoshi-git-fork-benalleng-issue341-savingsatoshi.vercel.app/en/chapters/chapter-2/scripting-2?dev=true)

TODO
- state gets reset on window size change


afterburn:
Looks like something broke for the desktop view:
image

Should be:
image

afterburn:
I also suggested we add a "Run" button that remains visible across all tabs/views, so users don't have to switch to the execute tab manually after they wrote some code, similar to how Replit for mobile does it:

image

Perhaps you can discuss this with @GBKS though, I think he has some opinions on this as well. But I think we should also address whatever comes out of that in this PR.

GBKS:
Yes sir, I already added this additional "run" button in Figma, [right here](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=3101%3A31466&t=YJXo8HFQkRguNbLm-1).

ui/lesson/ScriptingChallenge/ScriptingChallenge.tsx
```diff
@@ -57,25 +76,26 @@ export default function ScriptingChallenge({
...
+    <Lesson
+      direction={
+        isSmallScreen ? LessonDirection.Vertical : LessonDirection.Horizontal
+      }
+    >
+      <LessonTabs items={tabData} classes="px-4 py-2 w-full" stretch={true} />
+      {children}
 
-      <div className="flex flex-col">
+      <div className="flex max-h-[92.5vh] max-w-full grow flex-col border-white/25 font-space-mono text-white md:max-w-[50%] md:basis-1/3 md:border-l">
```
GBKS, line 87: What's the reason for the 92.5% viewport height via `max-h-[92.5vh]`? It leaves a bit of empty space at the bottom (on desktop).

Do you need `max-w-full`?

content/lessons/chapter-2/scripting-2.tsx
```diff
@@ -128,12 +128,14 @@ export default function Scripting2({ lang }) {
...
       successMessage={t('chapter_two.scripting_two.success')}
     >
-      <span className="font-nunito text-xl text-white">
-        {t('chapter_two.scripting_two.paragraph_one')}
-      </span>
-      <span className="font-nunito text-xl text-white">
-        {t('chapter_two.scripting_two.paragraph_two')}
-      </span>
+      <LessonInfo>
+        <span className="font-nunito text-xl text-white">
```
GBKS, line 132: Should these be `<p>` tags instead?

ui/lesson/ScriptingChallenge/ScriptingChallenge.tsx
benalleng: Good catch! They aren't necessary, these were here as an artifact from trying to get the terminal window to play nice

GBKS:
Thank you for working on this. Here's what I see on mobile. I'm ignoring the details of the scripting component (like it no picking up the theme color and the play button being the wrong size) because I think that's not being touched in this PR.

There should be a bit of space between the two text paragraphs, I think we already have this in the other screens.

image

And here on desktop. The scripting component is a smidge too short because it's set to 92.5% of viewport height. It should just stretch to the height of the container. It's generally best to size based on the container element rather than the viewport, except for some special cases.

image

Some really small stuff:
- The scripting component tabs should have a white line above on mobile
- In the execute tab, we don't need to show the language toggle
- The tab texts slightly shift when (de)selecting, probably because the styling switches from borders to backgrounds (just guessing)

GBKS:
Thanks for addressing these, looking good, the 92.5% is gone.

Something I just noticed is that the success box ("Nicely done! Next") shows up on the code tab when you click "Run". You don't even get to see the result of the execution because it doesn't switch the tab over. Could we make it so the "Run" on the "Code" tab also switches you to "Execute" before running the script?

ui/lesson/ScriptingChallenge/Editor.tsx
```diff
@@ -54,20 +63,37 @@ export default function Editor({
     setLoading(false)
   }
 
+  useEffect(() => {
```
afterburn, line 66: We should use the `useMediaQuery` hook for this

ui/lesson/ScriptingChallenge/Runner/index.tsx
```diff
@@ -40,8 +41,12 @@ export default function Runner({
   const [loading, setLoading] = useState<boolean>(true)
   const [isRunning, setIsRunning] = useState<boolean>(false)
   const [result, setResult] = useState<any | undefined>(undefined)
+  const [success, setSuccess] = useState('false')
```
afterburn, line 44: Why is this a string and not a boolean? This would prevent you from having to check `success === 'true' &&`, instead you can just do `success &&`.

ui/lesson/ScriptingChallenge/ScriptingChallenge.tsx
```diff
@@ -0,0 +1,120 @@
+'use client'
```
afterburn, line 1: Why did you move this from `index.tsx` to this file?

ui/lesson/index.ts
```diff
@@ -5,7 +5,7 @@ import LessonTerminal from './Terminal'
 import LessonTabs from './Tabs'
 import TerminalChallenge from './TerminalChallenge'
 import InputChallenge from './InputChallenge'
-import ScriptingChallenge from './ScriptingChallenge'
+import ScriptingChallenge from './ScriptingChallenge/ScriptingChallenge'
```
afterburn, line 8: See previous comment, this is unnecessary

afterburn (changes requested):
Thanks for picking this up @benalleng, requested some changes.

benalleng:
Got it I'll take a look! any idea why the fonts aren't loading in my deployment?

ui/lesson/ScriptingChallenge/Runner/index.tsx
benalleng: In most cases the status bar checks for a string has 4 conditions: empty, wrong, incomplete, and complete. Because we structured this challenge to only have two conditions I didnt want to redesign statusBar so i just passed true false strings

afterburn:
@benalleng not sure, I don't see any code changes that couldve caused that so I think we can ignore it.

afterburn (approved):

Issue #339, opened 2023-03-30
Incorrect Runner state when awaiting user input


afterburn:
**Current situation**
![image](https://user-images.githubusercontent.com/10757768/228918801-d5b327b4-493b-47b8-87b3-1917896fb7dc.png)

**Desired situation**
[Link to Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=2873-28026&t=puLldux6lTUTg04i-4)

afterburn:
As I mentioned in #390, this issue wasnt solved by the pr.

PR #390, opened 2023-04-18
Styling fix for initial state of scripting challenge

vinayaksh42:
fixes: #339

[preview](https://saving-satoshi-git-fork-vinayaksh42-issue339-saving-satoshi.vercel.app/en/chapters/chapter-2/scripting-2?dev=true)

### DIFF
--- a/ui/lesson/ScriptingChallenge/Runner/Hasher.tsx
+++ b/ui/lesson/ScriptingChallenge/Runner/Hasher.tsx
@@ -81,10 +81,12 @@ export default function Hasher({
   return (
     <div
       className={clsx(
-        'flex h-60 flex-col gap-4 overflow-y-auto border-t border-white border-opacity-30 p-4 font-mono text-white',
+        'flex flex-col gap-4 overflow-y-auto border-t border-white border-opacity-30 p-4 font-mono text-white',
         {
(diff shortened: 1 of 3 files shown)


afterburn (changes requested):
Code looks fine, but I noticed that the loading state of the runner looks like this:
image

While the loaded state looks like this:
image

Would be nice to have them be the same height.

afterburn:
Thanks @vinayaksh42 the height of the loader is now the same as the idle state of the Runner. I don't think this PR fixes #339 though, because there still is the "Custom validation message" which should be dynamically changed based on the current hash that is being validated. If you want to work on that issue in this PR feel free, otherwise let me know and we can merge this but keep #339 open.

afterburn (approved):

vinayaksh42:
> Thanks @vinayaksh42 the height of the loader is now the same as the idle state of the Runner. I don't think this PR fixes #339 though, because there still is the "Custom validation message" which should dynamically change based on the current hash that is being validated. If you want to work on that issue in this PR feel free, otherwise let me know and we can merge this but keep #339 open.

Sure, I will create a new PR for adding the "custom validation message"

afterburn:
@vinayaksh42 its okay I fixed this in #398

PR #399, opened 2023-04-21
Remove Mandatory signup screen + Introduced local storage when account is not created

### LINKED ISSUE #331: User login

shaavan:
Resolves #331

- Removes the mandatory signup screen.
- Introduce an option for loading/saving progress locally when the user has not created the account.

### DIFF
--- a/api/local/setProgress.ts
+++ b/api/local/setProgress.ts
@@ -1,0 +1,8 @@
+export default async function setProgress(progress: string) {
+  try {
+    localStorage.setItem('SavingSatoshiProgress', JSON.stringify(progress))
+  } catch (errors) {
+    console.error(errors)
+    return undefined
+  }
+}
--- a/api/local/setProgress.ts
+++ b/api/local/setProgress.ts
@@ -1,6 +1,6 @@
 export default async function setProgress(progress: string) {
   try {
-    localStorage.setItem('SavingSatoshiProgress', JSON.stringify(progress))
+    localStorage.setItem('SavingSatoshiProgress', progress)
   } catch (errors) {
     console.error(errors)
     return undefined
--- a/api/index.ts
+++ b/api/index.ts
@@ -1,2 +1,3 @@
 export * as auth from './auth'
 export * as progress from './progress'
+export * as local from './local'
--- a/api/local/getProgress.ts
+++ b/api/local/getProgress.ts
@@ -1,0 +1,14 @@
+// TO BE Edited
+
+export default async function getProgressLocal() {
+  try {
+    const res = (
+      localStorage.getItem('SavingSatoshiProgress') ?? 'CH1INT1'
+    ).replace(/['"]/g, '')
+
+    return res
+  } catch (errors) {
+    console.error(errors)
+    return undefined
+  }
+}
--- a/api/local/getProgress.ts
+++ b/api/local/getProgress.ts
@@ -2,9 +2,7 @@
 
 export default async function getProgressLocal() {
   try {
-    const res = (
-      localStorage.getItem('SavingSatoshiProgress') ?? 'CH1INT1'
-    ).replace(/['"]/g, '')
+    const res = localStorage.getItem('SavingSatoshiProgress') ?? 'CH1INT1'
 
     return res
   } catch (errors) {
--- a/api/local/index.ts
+++ b/api/local/index.ts
@@ -1,0 +1,2 @@
+export { default as getProgressLocal } from './getProgress'
+export { default as setProgressLocal } from './setProgress'
--- a/app/[lang]/chapters/[slug]/[lesson]/page.tsx
+++ b/app/[lang]/chapters/[slug]/[lesson]/page.tsx
@@ -88,11 +88,11 @@ export default function Page({ params }) {
     )
   }
 
-  const lastUnlockedLessonPath = getLastUnlockedLessonPath(progress)
+  const lastUnlockedLessonPath = getLastUnlockedLessonPath(progress!)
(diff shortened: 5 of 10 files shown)


shaavan:
The PR is correctly building and running on my system, but the Vercel deployment is failing for some reason.

@afterburn, would you please take a look into this? Thanks!

benalleng:
Tested on my local machine, something to look into is how we can initialize from the localprogress once logging in, at the moment it seems to reset user progress back to the beginning after logging in

afterburn:
Hey @shaavan thanks for working on this issue. I unfortunately dont have access to our Vercel team lol so I reached out to @adamjonas. Will get back to you asap.

shaavan:
@benalleng 

> At the moment it seems to reset user progress back to the beginning after logging in

Yes! This is something I thought about.
I wanted to get some initial reviews on the approach before implementing the account progress and local progress interaction.

shaavan:
@afterburn

> I unfortunately dont have access to our Vercel team lol so I reached out to @adamjonas. Will get back to you asap.

All right! Thanks, Kevin!

afterburn:
hey @shaavan, if you push another commit the vercel build should redeploy

shaavan:
> if you push another commit the vercel build should redeploy. I think maybe the issue was that there were too many open PR's.

Sure, let me give it a try!

shaavan:
Repushed the commit to restart Vercel Build.

Edit: The build is still failing 🥲

@adamjonas
A gentle ping. Please let us know the reason for the build failure in Vercel. Thanks!

shaavan:
Updated from b5fc8168bdfee6e94912c27e466f2334be3e55df to cabb691ad50d9cd72fa592891642cd60e490f804([pr399.03](https://github.com/shaavan/saving-satoshi/commits/pr399.03) -> [pr399.04](https://github.com/shaavan/saving-satoshi/commits/pr399.04), [diff](https://github.com/shaavan/saving-satoshi/compare/pr399.03..pr399.04))

Changes:

1. Rebased over master.
2. Fixed build issue related to:
    1. non-null assertion (`fn(progress)` -> `fn(progress!)`)
    2. bad import of Google fonts (`next/font/google` -> `@next/font/google`)

shaavan:
Changes:

Removed `node_modules` and reinstalled them to download the latest package versions.

shaavan:
Cleaned and reinstalled dependencies again after running the following instructions:

```
rm -rf node_modules
rm -rf .next
rm -rf yarn.lock

yarn install

yarn build
```

benalleng:
> 2\. bad import of Google fonts (`next/font/google` -> `@next/font/google`)

@shaavan I found that after properly upgrading my next version this error is actually not an error, reset your text editor and see if it still represents this as an error

shaavan:
@benalleng 

> I found that after properly upgrading my next version this error is actually not an error

I have noticed this, and with the latest PR update (1a33a6eeb878c68942e3c66ec471df634bbb598d), I reverted this change to work with the latest _Next_ version.

api/local/setProgress.ts
```diff
@@ -0,0 +1,8 @@
+export default async function setProgress(progress: string) {
+  try {
+    localStorage.setItem('SavingSatoshiProgress', JSON.stringify(progress))
```
afterburn, line 3: Not sure why you are using `JSON.stringify` here, this will result in extra `"` characters around the lessonId. This in turn forces you to use `.replace(/['"]/g, '')` [here](https://github.com/saving-satoshi/saving-satoshi/pull/399/files#diff-1a2ca889da37ca6ad47d64eaae05a834b5886e6a8bba5b1c2ea555045c97dccdR7).

afterburn (dismissed):
Hey @shaavan, requested a small change after which I will merge your PR. I checked it out locally and couldn't find any issues so we'll ignore the failing Vercel build.

api/local/setProgress.ts
shaavan: Not using Json.stringify() causes the following error.

image

That's why using `JSON.stringify()` while saving and using `.replace(/['"]/g, '')` while reading is the cleanest solution in our case.

afterburn: Hey @shaavan I checked out your branch locally, made the changes I suggested and was unable to reproduce your issue. 

The destructuring error happens if you try to do something like this:
```ts
const { path } = undefined
```
Which I dont see you doing anywhere in your code, perhaps minification has changed the names of variables but I'm wondering why you're seeing minified code in your browser window. Can you provide some steps to reproduce?

shaavan: > Can you provide some steps to reproduce?

1. I made the changes you suggested. (Same as bd82a76b651ee822a133fb0b90dd916a2fcec929)
2. Compiled and ran the branch (`npm run build` `npm run start`)
3. I navigated to [Intro-1 page](http://localhost:3000/chapters/chapter-1/intro-1) and noticed this error.

shaavan:
> Pushed the changes I suggested, build is now also working.

Finally, successful build!! Yay! 🥳

ecurrencyhodler:
tACK. Great work.

ecurrencyhodler (approved):

PR #403, opened 2023-04-22
Cleanup: TS fixes, package bumps and  icons fixes

afterburn:
This PR:
- Bumps next.js to the latest version
- Removes unnecessary packages
- Refactors the way we load icons
- Fixes all TypeScript errors


afterburn:
@ecurrencyhodler @benalleng please only test for regression compared to the `master` branch

ecurrencyhodler:
Mostly a tack. But I did see this extra white space button on the outro that shouldn't be there.
Screenshot 2023-04-24 at 11 39 12 AM

benalleng:
while that button exists only in this branch, it will not be a problem for long, it only exists because we currently don't want users to go straight to chapter 2 as it is not yet finished, its a pseudo continue button at the moment

benalleng:
Also a tACK from me, all the icons load as expected!

ecurrencyhodler (approved):

PR #405, opened 2023-04-25
New PR test with fresh codebase

benalleng:
This pr is to see if the build still fails if the codebase is swapped for a fresh version upgraded to the new next version.


benalleng:
unfortunately it worked because I hadn't pushed a commit I thought I had in original PR. This might not work for shaavan's PR

PR #407, opened 2023-04-26
Update Chapter End copy

### LINKED ISSUE #406: Update copy in account login flow
Our copywriter suggested the following copy changes to the "Save Progress" flow.

This flow appears at the end of Ch.1 when the player is prompted to save their progress.
Screenshot 2023-04-25 at 2 23 14 PM

Then they are taken to this screen:
Screenshot 2023-04-25 at 2 03 48 PM

I believe this is a new flow that must be created separately from the "create account" flow we currently have. @GBKS is this flow intended to be separate from creating an account?

benalleng:
This will close #406 when complete, but i wanted to get some testing about how we like the possible flow

TODO:
- Make the save my progress modal its own component
- make the continue button logic run after completing the saveProgress modal


ecurrencyhodler:
Okay I ran into a couple of issues. 

Text in the profile page needs to get updated:
Screenshot 2023-05-02 at 2 20 12 PM

Also the progress isn't getting saved for some reason and ch.2 is locked even though it goes to the right page (Chapters overview).


https://user-images.githubusercontent.com/32662508/235789781-2b63dfd5-ca29-4e1b-907c-f39ce166db81.mov

benalleng:
> Also the progress isn't getting saved for some reason and ch.2 is locked even though it goes to the right page (Chapters overview).

This is expected, will be fixed by @421

components/Modals/EndSignUp.tsx
benalleng, line 1: @afterburn I have a question about passing a new parameter to `SignUp.tsx` vs creating a whole new component in `EndSignUp.tsx`. Since we are using a type interface to describe modals how could I pass a new variable through to `SignUp.tsx` such that we can tell it when to run the save and return feature. Do i need to change the modal type interface to allow for this new variable?
```
const SignUp = (save: boolean) => {
save && useSaveAndReturn()
}
```

ecurrencyhodler:
tACK
tack

ecurrencyhodler (approved):

app/[lang]/chapters/[slug]/[lesson]/page.tsx
```diff
@@ -66,65 +67,45 @@ export default function Page({ params }) {
...
 
   const Lesson = chapterLessons[params.lesson].default
 
   if (dev) {
     return (
-      <>
-        <Head />
-        <Lesson lang={params.lang} />
-      </>
+      hydrated && (
```
Comment (afterburn) [inline app/[lang]/chapters/[slug]/[lesson]/page.tsx:81]: Is there any particular reason we want to wait for this component to be hydrated, seems like all the data that is being rendered should be available on intial render?

components/Modals/EndSignUp.tsx
afterburn: Hey @benalleng apologies for the late reply. Not sure I understand your question correctly but if your goal is to perform a save after sign up, but only sometimes, I think the idiomatic thing to do inReact is to provide a callback prop:

Parent component:
```ts
export default function Parent() {
  const handleSignUpComplete = () => {
    // Execute saving logic
  }

  return <SignUp
    onSignUpComplete={handleSignUpComplete}
  />
}
```

SignUp component:
```ts
export default function SignUp({ onSignUpComplete }) {
  const handleConfirm = () => {
    // Execute sign up logic
    if (typeof onSignUpComplete === 'function ) {
      onSignUpComplete()
    }
  }
}
```

afterburn (changes requested):
Hey @benalleng, lets see if we can get rid of the `EndSignUp` modal, we should be able to accomplish the same things using the existing `SignUp` modal with some minor changes.

I also noticed you have a tendency to add fixes to the PR that go beyond the scope of the issue and title of the PR. While I appreciate the effort, lets try to not do that anymore. Iinstead you can create separate PRs to address the issues you run into, if you feel the need. This simplifies the review process and helps us make sure we fully test your PR (right now we might miss some things during testing because they werent specified in the PR title/description).

components/Modals/SignUp.tsx
```diff
@@ -36,6 +39,20 @@ export default function SignUpModal({ open, onClose }) {
 
   const handleSetPrivateKey = (pk: string) => {
     setPrivateKey(pk)
+    setCopyAcknowledged(true)
+  }
+
+  const copy = () => {
```
afterburn, line 45: We can use the  `navigator.clipboard.writeText` api for this. Think it also makes sense to move this to the utils folder so we can more easily reuse it later.

components/Modals/SignUp.tsx
```diff
@@ -115,15 +134,14 @@ export default function SignUpModal({ open, onClose }) {
               {privateKey && (
                 <>
                   <code className="mb-2 whitespace-pre-wrap break-all text-base">
-                    {privateKey}
+                    {privateKey.toUpperCase()}
                   </code>
-                  <CopyButton content={privateKey}>
-                    {t('shared.copy')}
-                  </CopyButton>
+                  <Button round size="tiny" style={'w-full'} onClick={copy}>
```
afterburn, line 139: String props do not need `{}` around them.

components/Modals/EndSignUp.tsx
benalleng: Oh! that's super simple! I was lost in thinking i had to redesign the entire Modal context type to allow for a new callback

components/Navbar/HelpLink.tsx
```diff
@@ -9,8 +9,10 @@ export default function HelpLink({ params }: { params: any }) {
   const t = useTranslations(lang)
 
   const { slug, lesson: lessonId } = params
-  const { theme = chapters[slug].metadata.theme } =
-    lessons[slug][lessonId]?.metadata ?? 'bg-back'
+  const theme =
+    lessons[slug]?.[lessonId]?.metadata.theme ??
+    chapters[slug]?.metadata.theme ??
+    'bg-back'
```
afterburn, line 15: Lets put a comment here "If theme was specified on lesson it should take priority over a theme that was specified on a chapter, otherwise fallback to bg-back" or something similar.

hooks/useSaveAndReturn.ts
```diff
@@ -0,0 +1,40 @@
...
+      } else {
+        await saveProgressLocal(nextLessonKey)
+      }
+    }
+
+    router.push(
+      lang +
+        '/chapters#chapter-' +
+        (parseInt(currentLessonKey.charAt(2)) + 1).toString()
+    )
```
afterburn, line 36:
```ts
const chapterIndex = parseInt(currentLessonKey.charAt(2)) + 1).toString()

router.push(`${lang}/chapters#chapter-${chapterIndex}`)
```

ui/chapter/End.tsx
```diff
@@ -1,50 +1,57 @@
...
-            </Button>
-            {/*Secret Button*/}
-            <Button
-              onClick={saveAndProceed}
-              classes="w-full md:w-auto bg-transparent text-transparent hover:bg-transparent hover:cursor-default"
-            >
-              {t('shared.next')}
+            <Button onClick={handleClick} size="small">
+              {(!account && t('chapter_one.end.save')) ||
+                (account && t('shared.next'))}
```
afterburn, line 54:
```jsx
{!account && t('chapter_one.end.save')}
{account && t('shared.next')}
```

afterburn:
> > I also noticed you have a tendency to add fixes to the PR that go beyond the scope of the issue and title of the PR.
> 
> Right, sorry about that. I'll try to do better about splitting it all up moving forward

Oh no worries, no need to apologize either. Thanks for all your hard work :)

afterburn (approved):

afterburn:
@ecurrencyhodler please retest and ack, then we can merge, thanks

benalleng:
An error seems to have popped up whereby the progress is not being properly saved when transitioning from local to account storage

components/Modals/SignUp.tsx
```diff
@@ -50,6 +68,7 @@ export default function SignUpModal({ open, onClose }) {
         //The following line saves the latest localStorage progress to the backend.
         saveProgress(progress)
         onClose()
+        onSignUpComplete && saveProgress(getNextLessonKey(progress))
```
benalleng, line 71: Got it working! @afterburn I'd like one more re-review just to confirm this is an ok method to go about it,
changes made here...
Once the modal closes i save the progress to the next lessonkey whereby this allows the user to then proceed to the next key. I decided to not make a switch whereby the normal `saveProgress()` would not run if onSignUpComplete is not true because I felt that this ensures a linear progress path, i.e. no skipping progress in case there is an error.
Let me know what you think

sabakhilji:
Also after completing chapter 1, I landed on this page is this assumed to be normal I am not sure about it.

Screenshot (96)

components/Modals/SignUp.tsx
afterburn: What exactly is the problem? If a user completes chapter 1, and then registers, we should set his/her progress key to `CH2INT1`. I'm guessing thats what you are attempting to do here with `saveProgress(getNextLessonKey(progress))` right?

benalleng: it instead saves the local progress to `CH2INT1` but the backend gets saved to `CH1OUT2`

benalleng: Ok, I went ahead and tried `saveProgress(onSignupComplete ? nextLessonKey : progress)` after initializing nextLessonKey as `const nextLessonKey = getNextLessonKey(progress)`
And that seems to have worked

components/SignIn.tsx
```diff
@@ -83,6 +85,13 @@ export default function SignIn({
           </button>
         </div>
       </div>
+      <SignUpModal
```
afterburn, line 88: I noticed you moved the `SignUpModal` away from `components/Layout.tsx` and you're now managing whether its `open` or not through a `useState` in this component, lets not do that. I moved all modals to `components/Layout.tsx` so we could manage their `open` states through the `ModalProvider`.

You should only ever have to use:
```ts
const modals = useModalContext()

modals.open(Modal.SignUp)
modals.close(Modal.SignUp)
```

By moving it here and managing its state with a local `useState` you are breaking the `ModalProvider`.

afterburn:
@benalleng I'm having trouble getting the sign up modal to show up after completing chapter 1 without being logged in, do I need to do something manually to trigger it?

components/SignIn.tsx
benalleng: in this case how do I also provide it an optional component to know whether or not this should run `saveAndReturn()` with `onSignUpComplete` using the saveProgress button vs when logging in via the user button in the NavBar

afterburn: The system did not support it yet so I added it in [this PR](https://github.com/saving-satoshi/saving-satoshi/pull/430) but you can pass a second argument to `open`, e.g. 

```ts
modals.open(Modals.SignUp, { a: 1 })
```

Then in `SignUp.tsx` you can do:

```ts
export default function SignUpModal({ state, onClose }) {
  const { a } = state.meta
  console.log(a) // Outputs: 1
}

**Comment by afterburn** [comment]
Sorry, your deployment is failing due to this code in `SignIn.tsx`, which should be removed:
```ts
<SignUpModal
  open={modalOpen}
  onClose={() => {
    setModalOpen(false)
    close(Modal.SignIn)
  }}
/>
```

PR #412, opened 2023-04-26
Fix modals

### LINKED ISSUE #409: Add z-index to modals
We should add `z-index` to the element that has a `fixed` css class wrapping the modal itself:

image

afterburn:
This PR closes #409, removed our dependency on `react-modal` in favor of a custom implementation that we can modify how we see fit and it plays better with Next.js.

### DIFF
--- a/components/Layout.tsx
+++ b/components/Layout.tsx
@@ -12,20 +12,18 @@ export default function Layout({ children }: { children?: React.ReactNode }) {
   return (
     <>
       {children}
-      <>
-        <SignInModal
-          open={modals[Modal.SignIn]}
-          onClose={() => close(Modal.SignIn)}
-        />
-        <SignUpModal
-          open={modals[Modal.SignUp]}
-          onClose={() => close(Modal.SignUp)}
-        />
-        <AccountModal
-          open={modals[Modal.Account]}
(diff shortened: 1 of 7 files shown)


ecurrencyhodler:
tACK.

shaavan (approved):
ACK

Works perfectly. Code changes are clean, concise, coherent, and cool! :)

benalleng (approved):
LGTM! do you know if this solves a warning that is created when opening modals in current development mode? 
"Warning: react-modal: App element is not defined. Please use `Modal.setAppElement(el)` or set `appElement={el}`. This is needed so screen readers don't see main content when modal is opened. It is not recommended, but you can opt-out by setting `ariaHideApp={false}`."

afterburn:
> LGTM! do you know if this solves a warning that is created when opening modals in current development mode? "Warning: react-modal: App element is not defined. Please use `Modal.setAppElement(el)` or set `appElement={el}`. This is needed so screen readers don't see main content when modal is opened. It is not recommended, but you can opt-out by setting `ariaHideApp={false}`."

@benalleng It solves it in the sense that the prop no longer exists. We can spend some time on making these modals follow proper a11y principals later on since now we have full control over how they work.

PR #414, opened 2023-04-27
sanitize irregular whitespace around |

### LINKED ISSUE #408: Improve UX on user typing
I just found something that is a not a big issue but I think it would improve the UX, both ways are ok but only one is accepted as correct.

image

As in a real unix terminal 

`echo aabbcc | xxd -r -p`  == `echo aabbcc| xxd -r -p` == `echo aabbcc |xxd -r -p`  == `echo aabbcc|xxd -r -p`

shaavan:
Resolves #408

### DIFF
--- a/ui/lesson/TerminalChallenge.tsx
+++ b/ui/lesson/TerminalChallenge.tsx
@@ -66,6 +66,7 @@ export default function TerminalChallenge({
     const sanitizedInput = input
       .replace(/[\t\r\n]/g, '')
       .replace(/\s+/g, ' ')
+      .replace(/\s*\|\s*/g, ' | ')
       .trim()
 
     setLines((lines) => [...lines, { value: sanitizedInput, type: 'input' }])


afterburn (changes requested):
Functionally this works, but I don't think we've taken the right approach to how we parse and validate the user input. I think a better approach would be to tokenize the user input and compare the resulting array of tokens against the expected answer (also an array of tokens).

I think the approach of manually trying to account for every possible combination is not maintainable. I made a js fiddle [here](https://jsfiddle.net/nqzh7du0/2/) that works for the commands we currently need to support.

shaavan:
@afterburn 

> I think a better approach would be to tokenize the user input and compare the resulting array of tokens against the expected answer (also an array of tokens).

I thought about this approach before implementing this one. But I decided to go with the current approach because:

1. The fake terminal shall be used sparingly in the codebase as future challenges will be more complex scripting-based problems requiring a functional terminal.
2. The issue addressed by this PR is a very low-priority issue and would have gone unresolved for quite a while if it required considerable changes.
3. The current change, including the past ones, solves all the whitespace-related issues with the fake terminal.

And for these reasons, I went with the least intrusive code changes to solve this issue.

PR #421, opened 2023-04-28
Allow localProgress to persist after user Signs up or Logs out

### LINKED ISSUE #416: user login needs to save from localProgress
for the UI of the saveProgress function we want at the end of the chapters to work as expected we need it to start from wherever the user's current progress in local storage is.

shaavan:
Resolves #416

### DIFF
--- a/app/[lang]/chapters/[slug]/[lesson]/page.tsx
+++ b/app/[lang]/chapters/[slug]/[lesson]/page.tsx
@@ -110,7 +110,7 @@ export default function Page({ params }) {
 
   const lastUnlockedLessonPath = getLastUnlockedLessonPath(progress)
   const currentLessonPath = `/${pathData.pageId}/${pathData.chapterId}/${pathData.lessonId}`
-  const isRestrictedFromLesson = isLessonCompleted(
+  const isRestrictedFromLesson = !isLessonCompleted(
     getLessonKey(pathData.chapterId, pathData.lessonId),
     progress
   )
--- a/app/[lang]/chapters/[slug]/[lesson]/page.tsx
+++ b/app/[lang]/chapters/[slug]/[lesson]/page.tsx
@@ -108,11 +108,11 @@ export default function Page({ params }) {
     )
   }
 
-  const lastUnlockedLessonPath = getLastUnlockedLessonPath(progress!)
+  const lastUnlockedLessonPath = getLastUnlockedLessonPath(progress)
   const currentLessonPath = `/${pathData.pageId}/${pathData.chapterId}/${pathData.lessonId}`
   const isRestrictedFromLesson = isLessonCompleted(
     getLessonKey(pathData.chapterId, pathData.lessonId),
-    progress!
+    progress
   )
 
   if (
--- a/app/[lang]/chapters/[slug]/[lesson]/page.tsx
+++ b/app/[lang]/chapters/[slug]/[lesson]/page.tsx
@@ -108,18 +108,6 @@ export default function Page({ params }) {
     )
   }
 
-  // If account and progress data have been loaded, but there is no progress obj
-  if (!account || !progress) {
-    return (
-      <>
-        <Head />
-        <div className="flex h-full w-full grow flex-col items-center justify-center">
-          <SignIn lang={params.lang} />
-        </div>
-      </>
-    )
-  }
-
   const lastUnlockedLessonPath = getLastUnlockedLessonPath(progress!)
   const currentLessonPath = `/${pathData.pageId}/${pathData.chapterId}/${pathData.lessonId}`
   const isRestrictedFromLesson = isLessonCompleted(
--- a/components/Modals/SignUp.tsx
+++ b/components/Modals/SignUp.tsx
@@ -8,6 +8,7 @@ import HorizontalScrollView from 'components/HorizontalScrollView'
 import { useTranslations, useLang } from 'hooks'
 import clsx from 'clsx'
 import { useAuthContext } from 'providers/AuthProvider'
+import { useProgressContext } from 'providers/ProgressProvider'
 import { generateKeypair } from 'lib/crypto'
 import { register } from 'api/auth'
 import { Input } from 'shared'
@@ -28,6 +29,7 @@ export default function SignUpModal({ open, onClose }) {
   const [view, setView] = useState<string>(View.Generate)
   const [copyAcknowledged, setCopyAcknowledged] = useState<boolean>(false)
   const [privateKey, setPrivateKey] = useState<string | undefined>(undefined)
+  const { progress, saveProgress } = useProgressContext()
 
   const handleChangeView = (view: View) => {
     setView(view)
@@ -45,6 +47,7 @@ export default function SignUpModal({ open, onClose }) {
       if (privateKey) {
         await register(privateKey, `/assets/avatars/${avatar}.png`)
         await login(privateKey)
+        saveProgress(progress!)
         onClose()
       }
     } catch (ex) {
--- a/components/Modals/SignUp.tsx
+++ b/components/Modals/SignUp.tsx
@@ -47,7 +47,8 @@ export default function SignUpModal({ open, onClose }) {
       if (privateKey) {
         await register(privateKey, `/assets/avatars/${avatar}.png`)
         await login(privateKey)
-        saveProgress(progress!)
+        //The following line saves the latest localStorage progress to the backend.
+        saveProgress(progress)
         onClose()
       }
     } catch (ex) {
--- a/providers/ProgressProvider.tsx
+++ b/providers/ProgressProvider.tsx
@@ -7,7 +7,7 @@ import { getProgressLocal, setProgressLocal } from 'api/local'
 import { useAuthContext } from './AuthProvider'
 
 export const ProgressContext = createContext<ProgressContextType>({
-  progress: undefined,
+  progress: '',
   isLoading: true,
   saveProgress: (key: string) => Promise.resolve(),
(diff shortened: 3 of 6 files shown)


ecurrencyhodler:
Let's get 1 code review. Then I'll QA after we can build this with vercel.

components/Modals/SignUp.tsx
```diff
@@ -45,6 +47,7 @@ export default function SignUpModal({ open, onClose }) {
       if (privateKey) {
         await register(privateKey, `/assets/avatars/${avatar}.png`)
         await login(privateKey)
+        saveProgress(progress!)
```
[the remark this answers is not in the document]
afterburn: Also I think a comment would help here, I believe this is taking the progress that was stored in LocalStorage and sending that to the backend for storing it in the database right? That is not immediately clear from the code, so in cases like that I think it makes sense to add a comment,

afterburn (changes requested):
Small change requested for the non-null assertion.

app/[lang]/chapters/[slug]/[lesson]/page.tsx
```diff
@@ -108,18 +108,6 @@ export default function Page({ params }) {
...
-    return (
-      <>
-        <Head />
-        <div className="flex h-full w-full grow flex-col items-center justify-center">
-          <SignIn lang={params.lang} />
-        </div>
-      </>
-    )
-  }
-
```
Comment (benalleng) [inline app/[lang]/chapters/[slug]/[lesson]/page.tsx:122]: Looking back at this I now see what Jonas was saying, didn't we already remove this code? how did it end up back in the codebase from when it was removed in #399

components/Modals/SignUp.tsx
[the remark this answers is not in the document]
benalleng: where progress could possibly be undefined I think we should just default to setting it to 'CH1INT1'

app/[lang]/chapters/[slug]/[lesson]/page.tsx
Comment (ecurrencyhodler) [reply app/[lang]/chapters/[slug]/[lesson]/page.tsx:122]: This was probably a result of us having 6+ PR's pending for a long time all touching various parts of the codebase. @shaavan can you make sure this PR removes this section like you did in #399 again?

components/Modals/SignUp.tsx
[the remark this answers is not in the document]
shaavan: Addressed in b61b1001da3a6ae75a664201d422e40e1384bb32

Thanks!

app/[lang]/chapters/[slug]/[lesson]/page.tsx
Comment (shaavan) [reply app/[lang]/chapters/[slug]/[lesson]/page.tsx:122]: Addressed in b61b1001da3a6ae75a664201d422e40e1384bb32

Thanks!

providers/ProgressProvider.tsx
```diff
@@ -7,7 +7,7 @@ import { getProgressLocal, setProgressLocal } from 'api/local'
 import { useAuthContext } from './AuthProvider'
 
 export const ProgressContext = createContext<ProgressContextType>({
-  progress: undefined,
+  progress: '',
```
afterburn, line 10: Let's initialize this to `CH1INT1` like @benalleng suggested

providers/ProgressProvider.tsx
```diff
@@ -86,7 +86,7 @@ export default function ProgressProvider({
   return (
     <ProgressContext.Provider
       value={{
-        progress: accountProgress,
+        progress: accountProgress ?? 'CH1INT1',
```
afterburn, line 89: By initializing the progress variable to `CH1INT1` on line 10, you can probably just do `progress: accountProgress` here.

afterburn (changes requested):
Thanks @shaavan, minor change requested

providers/ProgressProvider.tsx
shaavan: I tried this line of code `progress: accountProgress`, but it causes the following error.

```
Type 'string | undefined' is not assignable to type 'string'.
Type 'undefined' is not assignable to type 'string'.
```

So I can either: 
1. change the `accountProgress` type to string or,
2. Use this (`progress: accountProgress ?? 'CH1INT1'`) maneuver

From these two, I chose the second option because it kept the code clean while being least intrusive.

benalleng: thats so odd, doing some of my own testing, even when setting the useState of account progress to be only a string and defaulted to 'CH1INT1' i still recieve an error on line 46 of `initLocal` with a type error stating that progress cannot be of type `string | undefined` even though it seems to be that progress has type string set everywhere

app/[lang]/chapters/[slug]/[lesson]/page.tsx
```diff
@@ -108,23 +108,11 @@ export default function Page({ params }) {
...
-  const lastUnlockedLessonPath = getLastUnlockedLessonPath(progress!)
+  const lastUnlockedLessonPath = getLastUnlockedLessonPath(progress)
   const currentLessonPath = `/${pathData.pageId}/${pathData.chapterId}/${pathData.lessonId}`
   const isRestrictedFromLesson = isLessonCompleted(
```
Comment (benalleng) [inline app/[lang]/chapters/[slug]/[lesson]/page.tsx:113]: I believe this should be proceeded by a not operand as we are checking that the user has Not completed the specified lesson to determine if they should be redirected.

In testing i am able to type in lesson paths I expect to be locked without being redirected
``` suggestion
  const isRestrictedFromLesson = !isLessonCompleted(
```

benalleng (changes requested):
Think a small bug slipped through

app/[lang]/chapters/[slug]/[lesson]/page.tsx
Comment (shaavan) [reply app/[lang]/chapters/[slug]/[lesson]/page.tsx:113]: Here comes @benalleng with his eagle eyes!! 🦅
Great catch! ✨

Since this change fixes a different bug significantly different from the goal of this PR, and since this change is independent of changes done in this PR, it's best to address this change separately to maintain the atomic nature of PR.

Let me create a separate PR for this change!

providers/ProgressProvider.tsx
shaavan: @benalleng 

That's because the progress in the `initLocal` function is equal to the return value of `getProgressLocal` whose return value can be `<string | undefined>`. That's why the error is showing.

image

It can be taken as one more reason to not change `accountProgress` type, because of the number of changes it requires to implement this PR's change.

app/[lang]/chapters/[slug]/[lesson]/page.tsx
Comment (shaavan) [reply app/[lang]/chapters/[slug]/[lesson]/page.tsx:113]: Okay, so this change does depend on changes in this PR (removing the mandatory signup screen)
Let me include this change with this PR.

Thanks again for catching this! 🙌

providers/ProgressProvider.tsx
benalleng: I guess my question is why should get ProgressLocal() allow for an undefined type? shouldn't it also default to 'CH1INT1'? 
I guess my thinking is that any instance of progress possibly being undefined would be better handled if it defaulted back to 'CH1INT1' as that is equivalent to base/no progress.

shaavan:
Updated from b61b1001da3a6ae75a664201d422e40e1384bb32 to 98ad49d932fc391a37a256fa92acfa24b3856fdf ([pr421.02](https://github.com/shaavan/saving-satoshi/commits/pr421.02) -> [pr421.03](https://github.com/shaavan/saving-satoshi/commits/pr421.03), [diff](https://github.com/shaavan/saving-satoshi/compare/pr421.02..pr421.03))
Addressed @benalleng comment.

Changes:

1. Fix the bug related to preventing access to locked challenges.

app/[lang]/chapters/[slug]/[lesson]/page.tsx
Comment (shaavan) [reply app/[lang]/chapters/[slug]/[lesson]/page.tsx]: Addressed in 98ad49d932fc391a37a256fa92acfa24b3856fdf

benalleng (approved):
tACK for the locked challenges.
local storage progress works as expected.

Testing pathway:
1. Went to "CH1GEN2" via localStorage then logged in, `accountProgress` and `localStorage` was now saved as "CH1GEN2" 
2. proceeded one more lesson to "CH1GEN3" then logged out, local storage remebered "CH1GEN2" and reverted me back one lesson
3. upon logging in with that same private key I was able to return to "CH1GEN3" without needing to use the next button or re-completing the challenge

I am approving as I will defer to @afterburn for the progress and accoutProgress types so that I am not a blocker

providers/ProgressProvider.tsx
shaavan: This seems to be a better approach to me at first glance.

But I noticed that "undefined" is used in the `getProgressLocal` function when setting res fails for some reason.

```js
export default async function getProgressLocal() {
  try {
    const res = localStorage.getItem('SavingSatoshiProgress') ?? 'CH1INT1'

    return res
  } catch (errors) {
    console.error(errors)
    return undefined
  }
}
```

So what should be the approach you would suggest:

1. Should we replace the undefined with "CH1INT1" even when `getProgressLocal` fails for unspecified reasons?

shaavan:
> tACK for the locked challenges.
local storage progress works as expected.

Yay! 🚀

Let's wait for @afterburn's ACK before merging!

afterburn:
Modified the code a bit so that we can initialize the progress provider to `CH1INT1` because this value should never be an empty string, `null` or `undefined`, if no progress can be retrieved for an account we should initialize it to `CH1INT1`. That way, every code that uses `progress` can safely assume that it'll always be a string to run comparisons against.

ecurrencyhodler:
tACK for me. Will wait for convo to resolve them merge.

PR #426, opened 2023-05-05
Remove Demo Info for V1

### LINKED ISSUE #402: Update copy on website to remove mentions of demo
Update copy to remove mentions of this being a demo.

- [ ] Home page
- [ ] Chapters page

**Home Page**
Screenshot 2023-04-21 at 12 45 18 PM

**Home page**
Screenshot 2023-04-21 at 12 45 29 PM

**Suggested copy**
Home page: Start the demo -> Play the game

Chapters page: Blocked by #381

vinayaksh42:
Fixes #402 

Removed demo info from:

- Home Page
- Chapter Page


ecurrencyhodler:
Sizing of the buttons is off. 
Screenshot 2023-05-05 at 1 27 03 PM

app/[lang]/chapters/page.tsx
```diff
@@ -22,13 +22,13 @@ export default async function ChaptersPage({ params }) {
             Chapters
           </h1>
 
-          <p className="pb-8 pl-10 pr-10 pt-4 text-center text-2xl sm:text-3xl lg:text-4xl">
+          <p className="pb-2.5 pl-10 pr-10 pt-4 text-center text-2xl sm:text-3xl lg:text-4xl">
             Explore the mysteries of Satoshi and learn about Bitcoin along the
             way.
```
Comment (benalleng) [inline app/[lang]/chapters/page.tsx:27]: Lets go ahead and bite the bullet on making these translation keys

Comment (vinayaksh42) [reply app/[lang]/chapters/page.tsx:27]: Sure, I will add it. What about the newly added text?
I changed the previous 'demo_Disclaimer' translation keys to only 'Disclaimer' for English but the other languages still have the old keys so do I update the keys or we should only update the keys when the translation within them is being updated?
I think we should formulate a plan for updating the translations.

ecurrencyhodler:
I think the button size is defined by the text. Is there anyway for them to be the same size regardless of text?
Screenshot 2023-05-06 at 9 13 44 AM

app/[lang]/chapters/page.tsx
Comment (vinayaksh42) [reply app/[lang]/chapters/page.tsx:27]: I tried to add the the title and chapter page description as translation keys but because this component is not a client component it is not possible to use the 'useLang' hook. One work around could be adding the text as component

ecurrencyhodler:
This is a tACK for me.
Screenshot 2023-05-07 at 6 56 49 PM

ecurrencyhodler (approved):

components/Hero.tsx
```diff
@@ -29,13 +29,13 @@ export default function Hero() {
         </p>
         <div className="grid grid-cols-1 justify-center pt-5 pb-2 sm:pb-8 md:grid-cols-2 md:space-x-5 md:pt-8">
           <div className="flex items-center justify-center md:justify-end">
-            <Button classes="text-xl w-full" href={routes.chaptersUrl}>
+            <Button classes="text-xl !w-64" href={routes.chaptersUrl}>
```
benalleng, line 32: we still need it to be full width on mobile, is there some error causing us to need the important tag?
``` suggestion
            <Button classes="text-xl w-full md:w-64" href={routes.chaptersUrl}>
```

components/Hero.tsx
```diff
@@ -29,13 +29,13 @@ export default function Hero() {
...
               {t('hero.start_journey')}
             </Button>
           </div>
           <div className="flex justify-center pt-5 md:justify-start md:pt-0">
             <Button
-              classes="text-xl w-full"
+              classes="text-xl !w-64"
```
benalleng, line 38: same here
``` suggestion
              classes="text-xl w-full md:w-64"
```

benalleng (changes requested):
I noticed you are using the important tag, is there a conflict that is causing some error discrepancies for the button width?

ecurrencyhodler:
@sabakhilji I think will be resolved in #407

components/Hero.tsx
vinayaksh42: upon removing the `!` important tag the w-auto(false) or w-full(true) class from https://github.com/saving-satoshi/saving-satoshi/blob/master/shared/Button.tsx#L30 the button component gets added. If I provide a value to the `full` prop it either gives the option for auto or full width and in case of not providing a value to the `full` prop it defaults to `w-auto` due to this ternary operation: 
```
full ? 'w-full' : 'md:w-auto'
```
In the local environment the width works without adding the important tag but in production, it disables the `w-64`class, and the `w-auto` gets enabled, hence the need for an important tag. I have added the w-full for mobile view.

Issue #431, opened 2023-05-10
QA issues


benalleng:
Here I will begin to create a list of issues I found while conducting QA. problems will generally be organized by order of devices where all devices are under all. Will separate devices by comments

benalleng:
desktop:

benalleng:
Android:
[ ] - text on [this](https://savingsatoshi.com/en/chapters/chapter-2/hashing-1) page has weird styling
- Firefox:

- Chrome:

afterburn:
Thanks for collecting these @benalleng, I think its better if next time we create a separate issue for each (however small the issue may be) so that we can address them in separate PRs, (and roll back easier, if needed). This also makes it so that it is easier for new contributors to chime in on smaller/easier issues.

PR #435, opened 2023-05-10
QA UX fixes

benalleng:
This PR addresses the following issues from #431
1. Chapter page styling to fix url fragment scroll behavior
2. InputChallenge responsize styling
3. outro page scroll behavior on mobile
4. navbar unecessary scrolling
5. Intro page consistent padding
6. Scripting challenge scroll fixes for mobile

### DIFF
--- a/components/Navbar/Navbar.tsx
+++ b/components/Navbar/Navbar.tsx
@@ -5,9 +5,9 @@ import NavbarMobile from './NavbarMobile/NavbarMobile'
 
 export default function Navbar({ params }) {
   return (
-    <div className="z-10 flex h-full">
+    <>
       <NavbarDesktop params={params} />
       <NavbarMobile params={params} />
-    </div>
+    </>
   )
 }
--- a/components/Navbar/Navbar.tsx
+++ b/components/Navbar/Navbar.tsx
@@ -5,9 +5,9 @@ import NavbarMobile from './NavbarMobile/NavbarMobile'
 
 export default function Navbar({ params }) {
   return (
-    <>
+    <div className="z-10 flex h-full">
       <NavbarDesktop params={params} />
       <NavbarMobile params={params} />
-    </>
+    </div>
   )
 }
--- a/content/lessons/chapter-1/outro-2.tsx
+++ b/content/lessons/chapter-1/outro-2.tsx
@@ -6,7 +6,7 @@ import { ChapterEnd } from 'ui'
 export const metadata = {
   title: 'chapter_one.outro_one.title',
   image: '/assets/images/chapter-1-finish.jpg',
-  theme: 'bg-transparent',
+  theme: 'bg-transparent/20',
   key: 'CH1OUT2',
 }
 
--- a/ui/chapter/End.tsx
+++ b/ui/chapter/End.tsx
@@ -33,12 +33,15 @@ export default function End({
   }
 
   return (
-    <div>
+    <div className="fixed h-[100vh] w-screen flex-col">
       <Image
         src={image}
         alt={t('chapter_two.title')}
         fill
-        className="relative -z-10 object-cover object-bottom"
+        quality={100}
+        loading="eager"
+        priority
+        className="h-full w-full object-cover"
       />
       <div className="absolute bottom-0 ml-auto w-full bg-gradient-to-b from-transparent via-[#00000040] to-[#00000080] p-4 pb-12 text-left md:p-16">
         <div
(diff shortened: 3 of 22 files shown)


shaavan (commented):
Hi, @benalleng!

Great work on solving the UX issues so far!
I have gone through half the commits, and the changes look solid.

However, I have noticed one issue in the changes, which we can look at; in the meantime, I go through the rest of the PR.

content/lessons/chapter-1/outro-2.tsx
```diff
@@ -6,7 +6,7 @@ import { ChapterEnd } from 'ui'
 export const metadata = {
   title: 'chapter_one.outro_one.title',
   image: '/assets/images/chapter-1-finish.jpg',
-  theme: 'bg-transparent',
+  theme: 'bg-transparent/20',
```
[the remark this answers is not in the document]
benalleng: good point, what do you think we should choose for a bg-color of the menu when dealing with a tansparent navbar?
Just the standard 'bg-back'?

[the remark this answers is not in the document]
shaavan: Standard bg-back will work well here.

1. It has looked good on the pages preceding it.
2. It will maintain the theme consistency with the rest of the chapter.

components/Navbar/Navbar.tsx
```diff
@@ -5,9 +5,9 @@ import NavbarMobile from './NavbarMobile/NavbarMobile'
 
 export default function Navbar({ params }) {
   return (
-    <>
+    <div className="z-10 flex h-full">
```
shaavan, line 8: Hey @benalleng 

I am not able to observe any UX change with this particular change. Would you please elaborate on this? Thanks!

Also, I am still able to scroll the navbar, and I can access the content after scrolling beyond the initial viewport. I don't think this is an optimal behavior.

**Reference:**

https://github.com/saving-satoshi/saving-satoshi/assets/85434418/33206b25-e7bf-4be6-a6de-70b390bb6963

shaavan (commented):
I went through the rest of the commits, and I could rationalize all the changes to be necessary and correct. I have one doubt and potential improvement suggestion that I have added to this comment.

Other than that, the code and UX improvements in this PR are perfect!

components/Navbar/Navbar.tsx
benalleng: You're correct, This was part not correctly describing the problem and part not properly fixing it, my intention is to simply have the mobile menu not add scrolling where it currently does and for it to bee the full length of the screen when scrolling already exists, which i have not fixed.

benalleng: Do you think we should disable scrolling for the navbar menu? I could see how it would be odd on a large copy page if you were to open the menu you would have a large blank space

benalleng: This particular change has to do with the functionality of having the background be an image

shaavan: > Do you think we should disable scrolling for the navbar menu?

This can be a good idea in the short run. But it can be limiting if we have a long list of navbar content and a small viewport to display it.

One solution would be to have the navbar occupy the full height of the page so no chapter content is visible when the dropdown menu is open.

Another solution could be to disable scrolling for all components except the dropdown menu when it is open.

afterburn: The usual way to handle this is to put `overflow-y: hidden` on the `<body>` element when the menu is opened, and to remove it when it is closed so users are not able to scroll when they have the menu open. Scrolling can happen inside the menu element itself, this preserves the users scroll position as well if I am not mistaken.

ui/chapter/End.tsx
```diff
@@ -33,12 +33,15 @@ export default function End({
   }
 
   return (
-    <div>
+    <div className="fixed h-[100vh] w-screen flex-col">
```
afterburn, line 36: Isn't `h-[100vh]` the same as `h-screen`?

ui/chapter/Intro.tsx
```diff
@@ -1,14 +1,26 @@
+import clsx from 'clsx'
+
 import { Title } from 'ui'
 
 export default function Intro({
   children,
   heading,
+  mobileCentered,
```
afterburn, line 8: Not a huge fan of props like these, makes it hard to override any css classes that are being added if this value is `true`. Let's instead just add an optional `className` prop so from the parent we can pass `my-auto my-8` if/when needed.

afterburn (changes requested):
Thanks @benalleng, requested some small changes and left a suggestion for the mobile menu in the convo you had with @shaavan.

shaavan (approved):
LGTM!

afterburn (approved):

PR #436, opened 2023-05-11
Scripting Challenge UX

### LINKED ISSUE #427: Various UX Challenges for Ch.2 Challenge 1 & 2
I did a quick run through of some of the challenges and am leaving some thoughts from a UX perspective of a player who is inexperienced with coding. I'm opening up this issue to start a discussion. Depending on how it goes, I may split this into individual issues. But let's discuss below.

**Ch.2 Challenge 1 Hashing-2**
url: en/chapters/chapter-2/hashing-2

It's not clear to me as a player what the success metric is to move on to the next section. It's not finding a hash with a "0". That's in pt. 2. To solve this, maybe we can say in the success bar at the bottom: "Nicely done! Now that you see how hashes work, continue to learn more about hashing!"

Screenshot 2023-05-05 at 12 45 31 PM

**Ch.2 Challenge 1 Hashing-6**
en/chapters/chapter-2/hashing-6

It took a while to see "need a hint". This should appear faster. Also the word "trigonometry" is hidden. IMO it shouldn't because it's already hidden behind a tool tip. Lastly, I suggest copy changes from "Answer" to "Try typing: trigonometry"
Screenshot 2023-05-05 at 1 12 15 PM

**Ch.2 Challenge 2 Hashing-6**
url: en/chapters/chapter-2/scripting-1

There is a copy mistake. Computem->computer
Screenshot 2023-05-05 at 12 46 45 PM

**Ch.2 Challenge 2 Scripting-2**
url: en/chapters/chapter-2/scripting-2

I don't know if this is because I don't know how to code. But I got stuck on this challenge section because I didn't know what formula to write. Considering how in Ch. 1 we hand held people through decoding op_returns, should we do something similar for this challenge? Perhaps we can add a tooltip with the formula if people don't know how to start?
Screenshot 2023-05-05 at 12 48 24 PM

benalleng:
Closes #427

[Preview](https://saving-satoshi-git-fork-benalleng-scriptingux-savingsatoshi.vercel.app/en/chapters/chapter-2/scripting-2?dev=true)

I altered the copy to hopefully make the lesson clearer, not easier for the user.

Additionally i tried to allow for more readable python error codes with the monaco editor. While it works, this method feels quite hacky, but I couldn't find the documantation for handling errors in the monaco editor


ecurrencyhodler:
tACK! all links to resources work. Text also adjusts well as I toggle between python/JS.

ui/lesson/ScriptingChallenge/Editor.tsx
sabakhilji, line 1: One question I want to ask is why there are 8 digits in the hex code of the background and what is the reason for to change?
//This is not related to the review, just a personal query to understand

benalleng: as compared to a regular characters hex should always be read in block of 2 characters.
The hex code #FFFFFF is signifying (FF) (FF) (FF). You might also notice that rgb colors are denoted in 3 pieces of information as well rgb(255, 255, 255) each number is representing the value of the different colors and hex is no different!
Why are there 8 characters here? well to add opacity rgba(255, 255, 255, 0.5) represents a color at 50% opacity and hex uses a "4th" character but becuase hex is read in blocks of 2 these would be characters 7 and 8

benalleng: In this particular cases I needed it as the monaco editor only accepts color codes, i.e. not "bg-transparent" so i needed hex and #000000 is black and transparent is represented as #00000000

shaavan (commented):
I have noticed one bug in the terminal.

While in **Python**, if I run the script without _digesting_ the `sha-256` in hexadecimal, the script gets stuck. While doing the same thing in **Javascript** causes the site to crash, leading to the loss of the written code.

**Python:**

Screenshot 2023-05-16 at 4 05 15 PM

**Javascript:**

https://github.com/saving-satoshi/saving-satoshi/assets/85434418/28775ff8-2e52-4d60-b679-9fd8de75e2d6

Forgetting to digest the result in write encoding seems like a common mistake to me, and this leading to page crashes can be a common and annoying problem faced by users.

shaavan (commented):
No crashes anymore! Yay!! 🚀

**Javascript:**

image

**Python:**

image

----
The error message in JS is much more helpful compared to Python's message. To maintain consistency and make it beneficial, you can change the error message of the Python terminal to match the JS.

benalleng:
This is something I think we should address in a separate discussion after some user QA, currently we have quite a bit of logic designed to display different TypeErrors and SyntaxErrors so we would need to manually set switches for each possible error or remove the custom error logic entirely. 
The reason the javaScript returns an error in this case is because in js that code technically isn't producing an error, just has a mismatched type to what we are expecting in the result

benalleng:
I noticed that the editor can be changed and thus causes you to lose progress, I am going to go ahead and make success stick for this challenge and since the run button disappears we won't allow the user to continue playing in this challenge. (I think this is actually the preferred behavior as any extended play and the user may realize this editor is not fully functional at all)

ui/lesson/ScriptingChallenge/index.tsx
```diff
@@ -29,29 +29,35 @@ export default function ScriptingChallenge({
   lang,
   config,
   successMessage,
+  languageCopy,
 }: {
   children?: React.ReactNode
   lang: string
   config: EditorConfig
   successMessage: string
+  languageCopy: any
```
afterburn, line 38: Lets use a function instead of `any` here, i.e:
```ts
languageCopy: (language: string) => {}
```

content/lessons/chapter-2/scripting-2.tsx
```diff
@@ -121,11 +125,18 @@ const config: EditorConfig = {
...
+  const handleLanguageCopy = (language) => {
+    setLanguage(language)
+  }
+
   return (
     <ScriptingChallenge
       lang={lang}
       config={config}
       successMessage={t('chapter_two.scripting_two.success')}
+      languageCopy={handleLanguageCopy}
```
afterburn, line 139: Let's rename this to `onCopyLanguage` so you know from the prop's name that it should be a function, this is the way its usually done in React

i18n/locales/en.ts
```diff
@@ -307,9 +307,21 @@ const translations = {
...
+      javascript_paragraph_three:
+        'We are using the crypto library in javascript to help you along in creating this function, you can use the following external resources to help you write this function if needed:',
+      python_list_one:
+        '<Link href="https://docs.python.org/3/library/hashlib.html" className="underline">hashlib documentation</Link>',
+      python_list_two:
+        '<Link href="https://datagy.io/python-sha256/" className="underline">Tutorial python function</Link>',
+      javascript_list_one:
+        '<Link href="https://www.geeksforgeeks.org/node-js-crypto-createhash-method/" className="underline">crypto documentation</Link>',
+      javascript_list_two:
+        '<Link href="https://www.educative.io/answers/what-is-node-cryptocreatehashalgorithm-options" className="underline">Tutorial javascript function</Link>',
```
afterburn, line 324: I think a better approach would be to create an object for each language if you consider that we plan on adding more languages later on, i.e.:
```ts
{
  javascript: {
    paragraph_three: '',
    list_one: '',
    list_two: ''
  },
  python: {
    paragraph_three: '',
    list_one: '',
    list_two: ''
  }
}
```

ui/lesson/ScriptingChallenge/Runner/index.tsx
```diff
@@ -85,15 +85,28 @@ export default function Runner({
...
             break
           }
           case 'result': {
-            setResult(payload)
+            typeof payload.value === 'string' && setResult(payload)
 
             if (payload.error) {
-              setErrors([payload.error])
-              setHasherState(HasherState.Error)
+              if (payload.error.split('Traceback')[1]) {
```
afterburn, line 95: This will become messy quickly, lets instead add language-specific error handling to their respective files i.e. `ui\lesson\ScriptingChallenge\compilers\python.ts`. It currently exports a `compile` function, lets also have it export a `handleError` function or something similar and call the correct one based on the `language` that is passed via props to this component, rather than attempting to split on 'Traceback', which could break when we add more languages.

afterburn (changes requested):
Thanks @benalleng, requested some changes.

benalleng:
@afterburn @ecurrencyhodler just pinging on how we should proceed with this PR.

ecurrencyhodler:
I think we should merge and ignore any errors we find in the console. I like the updated copy and think that should be preserved even if we replace the console with something else.

benalleng:
ok, ill merge with master in the morning

shaavan (approved):
ACK, in line with this comment:

> I think we should merge and ignore any errors we find in the console. I like the updated copy and think that should be preserved even if we replace the console with something else.

PR #438, opened 2023-05-12
Dynamic tooltip theming across chapters

### LINKED ISSUE #422: Fix tooltip styling
Something has happened to the tool tip styling, not sure when. This used to match the designs.

image

Desired look ([Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=2019%3A22776&t=a2nrfSs8Hc1yKO4I-1)).

image

benalleng:
Closes #422 

I added static themes in the localization tooltips and created a new metadata param for tooltips and other small UI as the gradient theme we are using for chapters does not look good on those small elements.

### DIFF
--- a/content/lessons/chapter-1/intro-1.tsx
+++ b/content/lessons/chapter-1/intro-1.tsx
@@ -7,7 +7,6 @@ export const metadata = {
   title: 'chapter_one.intro_one.title',
   image: '/assets/images/chapter-1-intro-1.jpg',
   theme: 'bg-[#2C3A3F]',
-  solidTheme: 'bg-[#2C3A3F]',
   key: 'CH1INT1',
 }
 
--- a/content/lessons/chapter-1/intro-1.tsx
+++ b/content/lessons/chapter-1/intro-1.tsx
@@ -7,6 +7,7 @@ export const metadata = {
   title: 'chapter_one.intro_one.title',
   image: '/assets/images/chapter-1-intro-1.jpg',
   theme: 'bg-[#2C3A3F]',
+  solidTheme: 'bg-[#2C3A3F]',
   key: 'CH1INT1',
 }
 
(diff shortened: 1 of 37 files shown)


content/lessons/chapter-1/intro-1.tsx
```diff
@@ -7,6 +7,7 @@ export const metadata = {
   title: 'chapter_one.intro_one.title',
   image: '/assets/images/chapter-1-intro-1.jpg',
   theme: 'bg-[#2C3A3F]',
+  solidTheme: 'bg-[#2C3A3F]',
```
afterburn, line 10: Not sure I understand why we need a new `solidTheme`, could you elaborate a little @benalleng? What does it accomplish?

benalleng: The purpose of the solidTheme is mainly to solve a problem that arises with the use of tooltips
Our chapter themes are all gradients at the moment and future chapters also seems to be planned with gradient backgrounds, while this looks good with a gradual gradient across the entire background, for smaller elements like tooltipsI find the gradient unsightly

In our last call @GBKS discussed trying to remove any popups for transparent backgrounds but I am also considering trying to move towards image backgrounds instead of next/Image + transparent bg

perhaps we should use "primaryTheme" "secondaryTheme" monikers

GBKS:
Alright, I am not sure if this makes your life any easier (probably not, sorry), but here's an idea of how it could work. Feel free to push back.

So every chapter/page has a theme, which means a background gradient. How about we add a background color in addition? The tooltip could use that. And the nav bar background could transition to that solid color when expanded on mobile. 

Saving Satoshi chapter end transparency mock-up 230519

Please let me know if this meshes at all with how things are set up technically right now.

benalleng:
This should work! only technical thing is to create a check in the navbarMobile so it know when to change color, other than that, this seems to be exactly what i was thinking with the theme, secondaryTheme setup :+1:

benalleng:
@GBKS what do you think a good color for chapter-2 outro would be?

GBKS:
Great that we're thinking along the same lines (I might just be agreeing with what you came up with in the first place 😀). That purple looks good.

afterburn (approved):

Issue #447, opened 2023-05-16
Error page should reload the page instead of returning the user back home


benalleng:
this should be easily handled by changing the button in `error.tsx` from an href to `onClick={() => router.reload()}`
additionally the copy key should be changed from `back_safety` to `reload`

shaavan:
This issue is addressed by #449, and is successfully merged! 🚀

PR #449, opened 2023-05-16
Error page will reload page on button click instead of going to home …

ujwalkumar1995:
Hey @benalleng 
This PR is for issue #447.
I have made the changes suggested by you. Please review.


app/[lang]/error.tsx
```diff
@@ -40,7 +42,7 @@ export default function Error({
         </h2>
 
         <div className=" mt-8 w-full text-center font-cbrush text-2xl md:w-auto">
-          <Button href="/">{t('error.back_safety')}</Button>
+          <Button onClick={() => router.reload()}>{t('error.reload')}</Button>
```
Comment (benalleng) [inline app/[lang]/error.tsx:45]: Looks like you were right about the router.reload() vs router.refresh() problem

though even after changing to router.refresh() i don't get the behavior i was expecting

benalleng (commented):
Seeing ass this method unfortunately produces an error, i wonder which is the proper method to account for refreshing.

benalleng:
To test getting to an error page you can go to `en/chapters/chapter-2/scripting-2?dev=true` then type 
```
const hash = crypto.createHash().update()
return hash 
```
in the terminal and run it

benalleng:
It looks like the next13 docs describe the router.refresh() as being a generally server-side refresh. 

> The client will merge the updated React Server Component payload without losing unaffected client-side React

In this case maybe it would be best to just use `() => window.location.reload()`

app/[lang]/error.tsx
Comment (benalleng) [inline app/[lang]/error.tsx:45]: As well, the translation data 'error.reload` would not return any valid copy as the key needs to be changed to match in `/i18n/locales/en.ts`

ujwalkumar1995:
Hey @benalleng 
Thank you for the inputs. Unfortunately even I was not able to find a better way to do this. I have used window.location.reload() as per your suggestion.

app/[lang]/error.tsx
Comment (ujwalkumar1995) [reply app/[lang]/error.tsx]: Thank you for the heads up. Fixed this.

benalleng (approved):
tACK! works as expected! And yeah, I just needed to read through the next docs more carefully! Thanks for this PR!

shaavan (approved):
ACK
Great work, @ujwalkumar1995! 🚀

shaavan:
A small thing before we merge!

Squashing the commits together will help keep the commit history clean. So it's best if you can do that!
Cheers! 🙌

ujwalkumar1995:
Hey @benalleng 
I have squashed the commits as per suggestion by @shaavan.
Let me know in case anything else needs to be fixed.

afterburn:
@shaavan again, no need to manually squash commits since we use the `squash and merge` button on github

shaavan:
@afterburn 
> no need to manually squash commits since we use the squash and merge button on GitHub

Oh, that's interesting! I didn't know that GitHub merge also squashes the commits together! It's going to save me a lot of time arranging commits! 🙌

Issue #441, opened 2023-05-15
Accessibility issue: Title of Back button and aria label for "need help" in navbar


sabakhilji:
According to its usage, the title of the back button in the navbar can be something else like "back to start" or "restart game". Right now it is back which was confusing for the user.
![Screenshot (100)](https://github.com/saving-satoshi/saving-satoshi/assets/52318459/30ed3217-48b7-44de-bb16-8e1d7a882635)

Additionally, right now "need help" link in the navbar has no aria label. We can use an aria-label like need help with it to make it accessible.

If this sounds good to everyone I would like to fix these issues.

PR #450, opened 2023-05-16
Navbar accessibility

sabakhilji:
Closes issue #441


shaavan (commented):
I agree with the approach that you have taken to making the app more accessible.
Way to go, @sabakhilji! 🚀

I have a small comment to add to make this PR even better.

i18n/locales/en.ts
```diff
@@ -11,6 +11,7 @@ const translations = {
     coming_soon: 'Coming soon',
     start_chapter: 'Start chapter',
     back: 'Back',
+    poweroff: 'back to chapter selection page',
```
shaavan, line 14: Small nit:

Capitalizing the sentence will look better :)

Also, I think we can remove "page" from the phrase to make it concise.

```suggestion
    poweroff: 'Back to chapter selection',
```

components/Navbar/HelpLink.tsx
```diff
@@ -36,7 +36,7 @@ export default function HelpLink({ params }: { params: any }) {
           target="_blank"
           href="https://forms.gle/NVcg9ukPvUBYjw1u7"
         >
-          ?
```
GBKS, line 39: By the way, the "?" should be Nunito and not Caveat.

sabakhilji: Hey, @GBKS I am not sure about that  '?'. The only line code which I added is related to a screen reader. 
Also, I would like to ask about the title of the power-off button. Right now I changed it to 'back to chapter selection page'. Please let me know what's your opinion about it.

GBKS: Yep, Ben is correct. The new title also looks good.

PR #451, opened 2023-05-17
Only render dynamic <head> client-side for lesson page

### LINKED ISSUE #445: Re-rendering the <head> causes hydration errors
In `/app/[lang]/chapters/]slug]/[lesson]/page.tsx` we re-render the head based on a few different switches and this causes hydration errors as we already have rendered the `<head>` and do not have logic to unmount it first

afterburn:
This PR fixes #445 by only rendering the `<head>` client-side. To accomplish this I added a `<Client>` component that we can use to wrap around other elements which should only render on the client-side.

### DIFF
--- a/app/[lang]/chapters/[slug]/[lesson]/page.tsx
+++ b/app/[lang]/chapters/[slug]/[lesson]/page.tsx
@@ -17,7 +17,7 @@ import { useProgressContext } from 'providers/ProgressProvider'
 import { Loader } from 'shared'
 import { LoadingState } from 'types'
 import { notFound } from 'next/navigation'
-import ClientOnly from 'components/ClientOnly'
+import Client from 'components/Client'
 
 export default function Page({ params }) {
   const searchParams = navigation.useSearchParams()
@@ -58,11 +58,11 @@ export default function Page({ params }) {
     const title = lesson ? t(lesson.metadata.title) : 'Page not found'
 
     return (
-      <ClientOnly>
+      <Client>
         <head>
           <title>{`${title} - Saving Satoshi`}</title>
         </head>
-      </ClientOnly>
+      </Client>
     )
   }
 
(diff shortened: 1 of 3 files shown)


shaavan (commented):
The code changes are clean and concise and prevent the issue of delivering child components before the parent is mounted.

One small suggestion. Squashing the commits of this PR together will keep the commit history clean.

The change seems logical. However, I am trying to catch the hydration failures in Master that this PR fixes, but I cannot find them. It will immensely help if you can share a few instances where this PR solves the hydration problems of the Master branch.

benalleng:
> I have found this occurs most when transitioning from the different switches, i.e. going to a url with ?dev=true and then continuing normally or logging in to load your progress after saving some progress to local storage

afterburn:
@shaavan where are you testing this? Both on savingsatoshi.com and the vercel deployments you will not be able to find the hydration errors since we're running a next.js production build there, they'll show up as minified react errors with a link to [this](https://legacy.reactjs.org/docs/error-decoder.html/?invariant=418).

Your best bet is to checkout master and go to a lesson page, you should then see an element for the Next.js error window somewhere on your screen with a text similar to "3 Errors" which you can press to open it up.

benalleng (approved):
LGTM! no longer getting any minified errors when using the different page loading switches

shaavan:
@afterburn 

> You can read more about this [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits). Not sure if you are mentioning it as a reminder or for some other reason.

Well, the merging process in the project I used to work on earlier was slightly different. Hence we use never used GitHub's "Squash and Merge". So we were focused on keeping the commit history super clean and logical from the PR's author side.
These suggestions are just those old habits popping up! 😂

PR #453, opened 2023-05-19
Introducing NumericInput Component

### LINKED ISSUE #443: Create a NumericInput component
For chapter 3 we'll need a Input component, basically a HTML `<input>`, but with some extra functionality. I'd like for this component to be reusable so it will need to support the following functionality:

For users:
1. Users should be able to input numbers only. For text we'll have a `TextInput` component at some point, this approach greatly reduces complexity.
2. Users should be able to increment/decrement the field by the `step` amount using all the different kinds of input that the HTML `<input>` also supports.

For engineers:
1. Engineers should be able to set a `step`, `min`, `max`, `name` and `label` via props.
2. Engineers should be able to use this both as a controlled and uncontrolled input field, if I pass a `defaultValue` it should be an uncontrolled component, if I pass a `value` and an `onChange` handler it should be a controlled component.

### Design

image

In context:

image

shaavan:
Resolves #443

- This PR introduces the `NumericInput` component.
- Allow increment or decrement of the number based on the step size
- Contains all the relevant properties of the `<input type="number">`
- Can be used as both _controlled_ and _uncontrolled_ components based on the provided props.

### DIFF
--- a/api/local/getProgress.ts
+++ b/api/local/getProgress.ts
@@ -2,11 +2,13 @@ import { defaultProgressContext } from 'providers/ProgressProvider'
 
 export default async function getProgressLocal(): Promise<string> {
   try {
-    const res = localStorage.getItem('SavingSatoshiProgress')
+    const progress = localStorage.getItem('SavingSatoshiProgress')
 
-    if (!res) {
+    if (!progress) {
       throw new Error('Could not read progress from LocalStorage')
     }
+
+    const res = progress.replace(/['"]/g, '')
 
     return res
   } catch (errors) {
--- a/app/[lang]/chapters/[slug]/[lesson]/page.tsx
+++ b/app/[lang]/chapters/[slug]/[lesson]/page.tsx
@@ -35,7 +35,7 @@ export default function Page({ params }) {
 
   const [unlocked, setUnlocked] = useState<number>(LoadingState.Idle)
 
-  // const [val, setVal] = useState(0)
+  const [val, setVal] = useState(0)
 
   useEffect(() => {
     if (!isAccountLoading && !isProgressLoading) {
@@ -115,15 +115,19 @@ export default function Page({ params }) {
     return navigation.redirect(lastUnlockedLessonPath)
   }
 
-  // function handleChange(evt) {
-  //   setVal(evt.target.value)
-  // }
+  function handleChange(evt) {
+    setVal(evt.target.value)
+  }
 
   return (
     unlocked && (
       <>
         <Head />
-        <NumericInput label={'string'} value={20} />
+        <NumericInput
+          label={'string'}
+          value={val}
+          onChange={(e) => setVal(e.target.value)}
+        />
         <Lesson lang={params.lang} />
       </>
     )
--- a/app/[lang]/chapters/[slug]/[lesson]/page.tsx
+++ b/app/[lang]/chapters/[slug]/[lesson]/page.tsx
@@ -18,6 +18,7 @@ import { Loader } from 'shared'
 import { LoadingState } from 'types'
 import { notFound } from 'next/navigation'
 import Client from 'components/Client'
+import NumericInput from 'components/Util/NumericInput'
 
 export default function Page({ params }) {
   const searchParams = navigation.useSearchParams()
@@ -33,6 +34,8 @@ export default function Page({ params }) {
   const { progress, isLoading: isProgressLoading } = useProgressContext()
 
   const [unlocked, setUnlocked] = useState<number>(LoadingState.Idle)
+
+  // const [val, setVal] = useState(0)
 
   useEffect(() => {
     if (!isAccountLoading && !isProgressLoading) {
@@ -112,10 +115,15 @@ export default function Page({ params }) {
     return navigation.redirect(lastUnlockedLessonPath)
   }
 
+  // function handleChange(evt) {
+  //   setVal(evt.target.value)
+  // }
+
   return (
     unlocked && (
       <>
         <Head />
+        <NumericInput label={'string'} value={20} />
         <Lesson lang={params.lang} />
       </>
     )
--- a/components/Util/NumericInput.tsx
+++ b/components/Util/NumericInput.tsx
@@ -1,0 +1,31 @@
+export default function NumericInput({
+  label,
+  name,
+  max,
+  min,
+  step,
+  defaultValue,
+  value,
+  onChange,
+}: {
+  label: string
+  name?: string
+  max?: number
+  min?: number
+  step?: number
+  defaultValue?: number
+  value?: number
+  onChange?: (evt: any) => void
+}) {
+  return (
+    <input
+      type="number"
+      name={name}
+      max={max}
+      min={min}
+      step={step}
+      defaultValue={defaultValue}
+      value={value}
+    />
+  )
+}
--- a/components/Util/NumericInput.tsx
+++ b/components/Util/NumericInput.tsx
@@ -17,6 +17,12 @@ export default function NumericInput({
   value?: number
   onChange?: (evt: any) => void
 }) {
+  if (value && defaultValue) {
+    throw new Error(
+      'Both value and defaultValue were given. Please provide one.'
+    )
+  }
+
   return (
     <input
       type="number"
@@ -24,8 +30,11 @@ export default function NumericInput({
       max={max}
       min={min}
       step={step}
-      defaultValue={defaultValue}
-      value={value}
+      //For uncontrolled component
+      defaultValue={defaultValue ?? undefined}
+      //For Controlled Component
+      value={value ?? undefined}
+      onChange={onChange ?? undefined}
     />
   )
 }


shaavan:
Right now, this PR introduces the skeleton of the component without the relevant styling, which will be addressed in a follow-up commit.

Also, this PR has introduced the `NumericInput` bar at the top of lesson content for easier testing by reviewers. This, too, shall be cleaned up in a follow-up commit.

@afterburn 

Please have a look through the code and functionality, and confirm if it satisfies our needs for the `NumericInput` component.

api/local/getProgress.ts
```diff
@@ -2,12 +2,14 @@ import { defaultProgressContext } from 'providers/ProgressProvider'
...
   try {
-    const res = localStorage.getItem('SavingSatoshiProgress')
+    const progress = localStorage.getItem('SavingSatoshiProgress')
 
-    if (!res) {
+    if (!progress) {
       throw new Error('Could not read progress from LocalStorage')
     }
 
+    const res = progress.replace(/['"]/g, '')
```
afterburn, line 11: Wasn't this removed in a previous PR?

components/Util/NumericInput.tsx
```diff
@@ -0,0 +1,40 @@
...
+  onChange,
+}: {
+  label: string
+  name?: string
+  max?: number
+  min?: number
+  step?: number
+  defaultValue?: number
+  value?: number
+  onChange?: (evt: any) => void
```
afterburn, line 18: Let's make it so this component returns a `number`. To me it makes sense to handle the event itself inside this component and only pass the resulting number up to the parent.  If I choose to use a `NumericInput` I expect the value that comes from it to be a number.

app/[lang]/chapters/[slug]/[lesson]/page.tsx
```diff
@@ -112,10 +115,19 @@ export default function Page({ params }) {
...
+  }
+
   return (
     unlocked && (
       <>
         <Head />
+        <NumericInput
+          label={'string'}
+          value={val}
+          onChange={(e) => setVal(e.target.value)}
```
Comment (afterburn) [inline app/[lang]/chapters/[slug]/[lesson]/page.tsx:129]: I think `e.target.value` will actually be a string in some browsers, we need to make sure it returns a number (see prev comments)

api/local/getProgress.ts
shaavan: Yes, this is fixed in #452.

This commit was a temporary fix to avoid crashing PR due to an unrelated bug.
I will remove this with the next update of PR.

components/Util/NumericInput.tsx
shaavan: Thanks for the review @afterburn.

I have some doubts to make about your comment.

> To me it makes sense to handle the event itself inside this component

Wouldn't handling the event inside the component make it an _uncontrolled component_? By definition, _controlled component_ states are governed by their parents.
I have allowed the possibility of both based on the description in #443 (For engineers, point 2).

components/Util/NumericInput.tsx
```diff
@@ -0,0 +1,40 @@
...
+      type="number"
+      name={name}
+      max={max}
+      min={min}
+      step={step}
+      //For uncontrolled component
+      defaultValue={defaultValue ?? undefined}
+      //For Controlled Component
+      value={value ?? undefined}
+      onChange={onChange ?? undefined}
```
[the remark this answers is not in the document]
shaavan: Out of all events, `onChange` seems the most logical one to handle first.
The other events can be handled later, as they will become relevant for the `NumericInput` component.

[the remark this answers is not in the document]
afterburn: I feel we also want to handle `onInput` at least

components/Util/NumericInput.tsx
afterburn: What I suggest is that we handle the event inside the component, this does not make it an uncontrolled component because as a developer you can still choose to manage the state for the input in the parent or have the `NumericInput` track state which is what the difference is between a controlled and uncontrolled component, where we handle the event has nothing to do with that.

Something like this:
```tsx
function NumericInput() {
  const [value, setValue] = useState(value)

  const handleChange = e => {
    const newValue = Number(e.target.value)
    
    // Based on whether its a controlled or uncontrolled component:
    if (controlled) {
       onChange(newValue)
    } else {
      setValue(newValue)
    }
  }

  return <input onChange={handleChange} />
}
```

This makes the component easier to use, which is the point:

```
function Parent() {
  const handleChange = (val: number) => {
    console.log(val) // Outputs number
  }
  
  return <NumericInput onChange={handleChange} />
}
```

afterburn: Full example of what I mean [here](https://codesandbox.io/s/numeric-input-xgpd6t)

shaavan: Thanks, @afterburn, for the example. This helps clears up my doubts!

Let me take references from it to create the final component!

components/Util/NumericInput.tsx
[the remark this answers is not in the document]
shaavan: According to [w3schools](https://www.w3schools.com/jsref/event_oninput.asp#:~:text=The%20oninput%20event%20is%20similar,when%20the%20element%20loses%20focus.)

> The oninput event is similar to the [onchange](https://www.w3schools.com/jsref/event_onchange.asp) event.
The difference is that the oninput event occurs immediately after the content has been changed, while onchange occurs when the element loses focus.

Since they are so similar in behavior, tackling both might unnecessarily increase complexity.

api/local/getProgress.ts
shaavan: Addressed in 1c7cd6cb913ce1305fa02df9af0a4e5b9c955001

Thanks!

components/Util/NumericInput.tsx
shaavan: Addressed in 1c7cd6cb913ce1305fa02df9af0a4e5b9c955001
I have taken reference from your example to create the final component.
Thanks for the detailed example!

app/[lang]/chapters/[slug]/[lesson]/page.tsx
Comment (shaavan) [reply app/[lang]/chapters/[slug]/[lesson]/page.tsx]: Addressed in 1c7cd6cb913ce1305fa02df9af0a4e5b9c955001
The latest implementation passes the setValue function in the NumericInput.

vinayaksh42:
I am unable to see the `Uncontrolled` NumericInput component in Vercel deployment link, although it is visible when I am running this PR locally. 
Screenshot 2023-05-29 at 2 18 17 PM

components/Util/NumericInput.tsx
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  useEffect(() => {
+    if (isControlled) {
+      setInternelValue(value)
+    }
+  }, [isControlled, value])
+
+  return (
+    <input
```
vinayaksh42, line 52: I think we should add onkeydown for the enter key, the same way we have in ui/lesson/Terminal.tsx.

vinayaksh42 (dismissed):

components/Util/NumericInput.tsx
shaavan: Seems reasonable to me.

Since we don't have any place to accept the numericInput, let me create a dummy handler that logs the value in the terminal when "Enter" is pressed.

shaavan: Also @vinayaksh42 

> I am unable to see the Uncontrolled NumericInput component in Vercel deployment link

I opened the Vercel preview [link](https://saving-satoshi-h2m4c54u1-savingsatoshi.vercel.app/) (Mac, Chrome) in both new and incognito windows and can see both components.

image

---

Can it be browser specific problem? Which browser are you working with?

shaavan:
Updated from 1c7cd6cb913ce1305fa02df9af0a4e5b9c955001 to 704c28130b7e87852932dc490720f50780f9e2a2 ([pr453.02](https://github.com/shaavan/saving-satoshi/commits/pr453.02) -> [pr453.03](https://github.com/shaavan/saving-satoshi/commits/pr453.03), [diff](https://github.com/shaavan/saving-satoshi/compare/pr453.02..pr453.03))
Addressed @vinayaksh42 comment

- Introduced onKeyDown event in the `NumericInput` component.

components/Util/NumericInput.tsx
```diff
@@ -0,0 +1,76 @@
...
+  // The logical operation here simulates XOR operation
+  // Only either of them should be TRUE at one time.
+  // This operation ensures that
+  if ((value && onChange) === defaultValue) {
+    throw new Error('Provide only either of value or defaultValue.')
+  }
+
+  const isControlled = value !== undefined && onChange !== undefined
+
+  const [internalValue, setInternelValue] = useState(
```
afterburn, line 33: internel -> internal

components/Util/NumericInput.tsx
```diff
@@ -0,0 +1,76 @@
...
+    if (isControlled) {
+      onChange(newValue)
+    } else {
+      setInternelValue(newValue)
+    }
+  }
+
+  function handleKeyDown(evt) {
+    if (evt.key == 'Enter') {
+      if (isControlled && onKeyDown) {
```
afterburn, line 49: I should also be able to pass an `onKeyDown` to an uncontrolled component in case I want to hook into that event, same goes for `onChange` the only difference between the controlled and uncontrolled variants is how we handle the internal state.

shaavan:
Updated from cc34c09e6c252ca8d298e44167934c24bd670b70 to 4022862094a473147c9092f6fb5753def89be919  ([pr453.04](https://github.com/shaavan/saving-satoshi/commits/pr453.04) -> [pr453.05](https://github.com/shaavan/saving-satoshi/commits/pr453.05), [diff](https://github.com/shaavan/saving-satoshi/compare/pr453.04..pr453.05))
Addressed @afterburn comment

Update

- Removed the test code for NumericInput from the page.tsx

components/Util/NumericInput.tsx
shaavan: Addressed in cc34c09e6c252ca8d298e44167934c24bd670b70
Thanks!

components/Util/NumericInput.tsx
shaavan: Addressed in cc34c09e6c252ca8d298e44167934c24bd670b70
Thanks!

ecurrencyhodler (approved):

PR #459, opened 2023-05-22
Change progress logic to only save progress on advancement

### LINKED ISSUE #456: Not saving account progress if I go back to Ch. 1
I completed Ch.1 and unlocked Ch. 2. I also completed Ch. 2. You can see this in the beginning of the video where I scroll down to Ch. 2 on the "Chapter overview" page. 

But if I go back into Ch. 1 and get to the "outro" stage, my account is reverted. The video skips to the end where I land on the outro page. I choose not to complete it and go back to the "Chapter overview" page. When I do that I see that I am now locked out of Ch. 2 until I finish the outro in Ch. 1.

https://github.com/saving-satoshi/saving-satoshi/assets/32662508/7c7f704b-1561-4b36-8258-15550c965c72

The ideal solution here is that my account progress is saved and I can freely move across previous chapters I completed. This will also be really helpful when doing QA.

benalleng:
Closes #456

This adds a check to make sure that the user is only saving their progress on new progress steps and not if they go back through the lessons.

### DIFF
--- a/providers/ProgressProvider.tsx
+++ b/providers/ProgressProvider.tsx
@@ -5,6 +5,7 @@ import { ProgressContextType } from 'types'
 import { getProgress, setProgress } from 'api/progress'
 import { getProgressLocal, setProgressLocal } from 'api/local'
 import { useAuthContext } from './AuthProvider'
+import { keys } from 'lib/progress'
 
 export const defaultProgressContext = {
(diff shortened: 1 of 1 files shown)


ecurrencyhodler (approved):
tACK! I went to the Ch.1 Outro and I am no longer locked out of Ch.2!

PR #472, opened 2023-05-24
Add fragment url from start page when the user has completed the first chapter

benalleng:
This will auto scroll from the start page if the user is already logged in and completed the first challenge taking them directly to their latest progress.

### DIFF
--- a/components/Hero.tsx
+++ b/components/Hero.tsx
@@ -3,25 +3,15 @@
 import { Button } from 'shared'
 import { useLang, useLocalizedRoutes, useTranslations } from 'hooks'
 import Image from 'next/image'
-import { useAuthContext } from 'providers/AuthProvider'
 import { useProgressContext } from 'providers/ProgressProvider'
-import { keys } from 'lib/progress'
-
+import { getChapterProgressUrl } from 'lib/progress'
 export default function Hero() {
   const routes = useLocalizedRoutes()
   const lang = useLang()
   const t = useTranslations(lang)
-  const { account } = useAuthContext()
   const { progress } = useProgressContext()
 
-  const progressHrefFragement =
-    progress === keys[keys.length - 1]
-      ? (parseInt(progress.substring(2, 3)) + 1).toString()
-      : progress.substring(2, 3)
-  const chapterHref =
-    account && progressHrefFragement !== '1'
-      ? `${routes.chaptersUrl}#chapter-${progressHrefFragement}`
-      : routes.chaptersUrl
+  const progressFragment = getChapterProgressUrl(progress)
 
   return (
     <div className="flex h-full grow items-end justify-center">
@@ -43,7 +33,10 @@ export default function Hero() {
         </p>
         <div className="grid grid-cols-1 justify-center pt-5 pb-2 sm:pb-8 md:grid-cols-2 md:space-x-5 md:pt-8">
           <div className="flex items-center justify-center md:justify-end">
-            <Button classes="text-xl md:!w-64 w-full" href={chapterHref}>
+            <Button
+              classes="text-xl md:!w-64 w-full"
+              href={`${routes.chaptersUrl}${progressFragment}`}
+            >
               {t('hero.start_journey')}
             </Button>
           </div>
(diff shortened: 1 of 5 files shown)


components/Hero.tsx
```diff
@@ -3,11 +3,25 @@
...
 
 export default function Hero() {
   const routes = useLocalizedRoutes()
   const lang = useLang()
   const t = useTranslations(lang)
+  const { account } = useAuthContext()
+  const { progress } = useProgressContext()
+
+  const progressHrefFragement =
+    progress === keys[keys.length - 1]
```
shaavan, line 18: This logic works correctly for now since the last element in the keys array is currently equal to 'CH2OUT1'.
However, this line of code will not work properly when we add further chapters.

For example,
```
progress = 'CH2OUT1'
keys[keys.length - 1] = 'CH3OUT1'
```

So `progressHrefFragement` according to this logic will be `2`. Whereas it should be `3`.

A more robust approach would be to check if the user progress is equal to last lesson of current chapter. The following code achieves that.

```suggestion
    progress === keys.filter(ele => ele.includes(progress.substring(0,3))).pop()
```

benalleng: So this ternary function is supposed to produce two results
1. When the user is at the very end of the game, i.e. their progress === keys.[keys.length -1] they are sent to the chapter Id +1 so that they can see that the next lesson is 'coming soon'
1. All other progress will direct you to the chapter based on your current progress

shaavan: Okay, this makes sense.

So what I understand is that,
Say I am on `CH2OUT1`, and instead of pressing next, I go back to the home screen.
So when I reopen chapter selection, I will be directed to _Chapter-2_, instead of _Chapter-3_.

benalleng: Correct. As I said in my comment in your PR #477 I like the idea of automatically being able to skip the end screen progress, but we need to add that logic where the progress is tracked and not through any sub-logic that feeds from the progress as It will inevitably cause some headache down the line

benalleng: With my progress at `CH1OUT2` I will still need to progress forward before the next chapter is unlocked, we can add some logic so that the outro screen progress is automatically skipped but we also need to consider in the first chapter we are expecting/requiring the user to log in so this logic would need to check if the user is logged in first

shaavan: Got it!
So skipping the outro page will be unreasonable since we want the user to signup up by the outro of Chapter-1.

shaavan: So logic-wise, the code is great!
So we can separate these changes into a `common` function, and we shall be good to go!

benalleng:
@shaavan I went ahead and created a shared function in lib/progress (not sure if this is the best place for it)

As well I implemented it in the power button and our `useSaveandReturn()` function so that the user is returned to the chapter they were last in when pressing the power button and the `useSaveandReturn()` takes them to the next chapter fragment

afterburn (dismissed):
Let's not introduce a new word 'fragment' for progress keys, let's just keep calling them keys.

lib/progress.ts
```diff
@@ -160,3 +160,21 @@ export const getLessonKey = (chapterId, lessonId) => {
...
+      ? (parseInt(userProgressKey.substring(2, 3)) + 1).toString()
+      : userProgressKey.substring(2, 3)
+  if (
+    !userProgressKey ||
+    (userProgressKey.startsWith('CH1') && !forceFragment)
+  ) {
+    return ''
+  } else {
+    return `#chapter-${progressHrefFragement}`
+  }
```
shaavan, line 179: We can simplify this logic.

1. In the function definition, we explicitly define `userProgressKey` to be a string. So to again check for it being defined is redundant.
2. The name 'forceFragment' can be changed to 'getFragment' to sound closer to its usage.

```suggestion
if (getFragment || !userProgressKey.startsWith('CH1')) {
  return `#chapter-${progressHrefFragement}`
}
return ''
```

benalleng: I simplified the logic such that I do not think a comment is necessary to describe is function.

shaavan (approved):
ACK

Great Simplification of the getChapterKey function. Also, overall the PR looks good!

ecurrencyhodler (approved):

PR #474, opened 2023-05-25
Fix: scroll to top for Chapter.2 Intro-1

### LINKED ISSUE #462: Ch.2 iPhone/Firefox: Start of Ch.2 takes me to text and I don't see the image.
I started Ch. 2 on iPhone/Firefox and it takes me straight to the text and I don't see the image. Starting Ch.2 should take me to the top of the page, not the middle of it I would think.

https://github.com/saving-satoshi/saving-satoshi/assets/32662508/c739e32c-94ff-483f-8a04-213240d0a10a

vinayaksh42:
Fixes #462

### DIFF
--- a/app/[lang]/chapters/[slug]/[lesson]/page.tsx
+++ b/app/[lang]/chapters/[slug]/[lesson]/page.tsx
@@ -18,7 +18,7 @@ import { useAuthContext } from 'providers/AuthProvider'
 import { useProgressContext } from 'providers/ProgressProvider'
 import { Loader } from 'shared'
 import { LoadingState } from 'types'
-import { notFound } from 'next/navigation'
+import { notFound, usePathname } from 'next/navigation'
 import Client from 'components/Client'
 
 export default function Page({ params }) {
@@ -35,6 +35,9 @@ export default function Page({ params }) {
   const { progress, isLoading: isProgressLoading } = useProgressContext()
 
   const [unlocked, setUnlocked] = useState<number>(LoadingState.Idle)
+
+  const pathname = usePathname()
+  useEffect(() => window.scroll(0, 0), [pathname])
 
   useEffect(() => {
     if (!isAccountLoading && !isProgressLoading) {
--- a/app/[lang]/chapters/[slug]/[lesson]/page.tsx
+++ b/app/[lang]/chapters/[slug]/[lesson]/page.tsx
@@ -18,7 +18,7 @@ import { useAuthContext } from 'providers/AuthProvider'
 import { useProgressContext } from 'providers/ProgressProvider'
 import { Loader } from 'shared'
 import { LoadingState } from 'types'
-import { notFound, usePathname } from 'next/navigation'
+import { notFound } from 'next/navigation'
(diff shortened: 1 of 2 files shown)


ecurrencyhodler:
tACK!

benalleng:
Is this a problem on any other page?

ecurrencyhodler:
Only noticed in from chapters overview page to ch. 2 intro-1.

benalleng:
@vinayaksh42 I wonder if we can put this logic in a higher parent component like in `[slug]/[lesson]/page.tsx` to ensure we don't have to repeat this logic if it crops up in the future?
Outside of that, finding the logic that explains why it might happen only on this page?

vinayaksh42:
> @vinayaksh42 I wonder if we can put this logic in a higher parent component like in `[slug]/[lesson]/page.tsx` to ensure we don't have to repeat this logic if it crops up in the future?
> Outside of that, finding the logic that explains why it might happen only on this page?

Sure, we can have this logic at a deeper level.

This is actually only happening in the Firefox browser, the issue here is that the scroll done in order to reach the chapter 2 button on the chapter overview page persists to the intro page of chapter-2.

benalleng:
Ok well in that case it sounds like it might be possible to happen on other pages should the copy of the previous page be long enough, in that case lets go ahead and move it to the page.tsx whereby once any lesson page is rendered we can ensure it starts at the top of the page so th user doesn't miss any content

afterburn:
I agree with @benalleng here, other lessons might not use the `TextImageLayout` component but could be suffering from the same issue (for whatever reason), simplest way to ensure users are always sent to the top of the page on mobile is to just always scroll to `0, 0` which can be easily achieved by placing this effect in `[slug]/[lesson]/page.tsx` instead.

afterburn (dismissed):
I agree with @benalleng here, other lessons might not use the `TextImageLayout` component but could be suffering from the same issue (for whatever reason), simplest way to ensure users are always sent to the top of the page on mobile is to just always scroll to `0, 0` which can be easily achieved by placing this effect in `[slug]/[lesson]/page.tsx` instead.

vinayaksh42:
@ecurrencyhodler Can you check if the fix still works for iPhone/firefox? 
I did try it out and majority of the time it worked properly but a few times (1/10) the fix didn't really work.
I think it has something to do with when the actual page loads. Having the logic in page.tsx initiates the useEffect at an early stage, were as when it was in the TextImageLayout.tsx it was being initiated when the component actually got mounted. 

Found this discussion:  https://github.com/vercel/next.js/issues/28778
I have tried out most of the solutions mentioned in the above discussion but nothing really works as reliably as having the useEffect scrollTo within TextImageLayout.tsx

ecurrencyhodler:
Let's go back to your initial fix of bring it back down to the page rather than page.tsx

ecurrencyhodler:
tACK!

benalleng (approved):
LGTM! Like we said in our dev call today we can circle back to this if it becomes an issue in the future.

PR #477, opened 2023-05-26
Introduce Continue Button

shaavan:
This PR introduces the continue button.

The button will direct the user to their current progress for the current chapter.

- This PR introduces a new `common` function to make this button functional.


shaavan:
@benalleng 

Finally did it! 🚀

Please have a look at my implementation.

ui/chapter/Chapter.tsx
```diff
@@ -123,15 +131,17 @@ export default function Chapter({ children, metadata, lang }) {
                       (chapter.metadata.lessons.length === 0 && null)}
                     <div className="flex pt-8 md:w-full">
                       <Button
-                        href={`${routes.chaptersUrl}/${chapter.metadata.slug}/${chapter.metadata.intros[0]}`}
+                        href={`${routes.chaptersUrl}/${chapter.metadata.slug}/${linkFrag}`}
```
benalleng, line 134: using something along the lines of
``` suggestion
                        href={`${keysMeta.progress.path}`}
```

or something similar might simplify the logic

**I noticed that the `keysMeta` already has the `chapters` in the url what do you think about removing it in a future PR as we already have this `routes.chapters` logic built in

benalleng (changes requested):
Looking great! 
Still have a few things to workout at the moment. All buttons are tracking whether you are in the middle of the lesson so we get some confusing copy to the user


Additionally in the middle of chapter-2 i ran into an issue whereby the lesson was no longer being tracked 
`https://saving-satoshi-git-fork-shaavan-continue-btn-savingsatoshi.vercel.app/en/chapters/chapter-2/undefined`

Lastly at the outro it takes me to a combination of the two outro pages in the url
`https://saving-satoshi-git-fork-shaavan-continue-btn-savingsatoshi.vercel.app/en/chapters/chapter-1/outro-1,%20outro-2`

I think some of the logic might be better suited to simply use the `keysMeta` object in `/lib/progress.ts` appended to the end of the path.

ui/chapter/Chapter.tsx
shaavan: > I noticed that the keysMeta already has the chapters in the url what do you think about removing it in a future PR as we already have this routes.chapters logic built in

This would be a fair simplification while also introducing consistency in the manner we use hrefs in the codebase.

shaavan:
Updated from 17e27d332600dc8623b3593a68300fb1159e2dff to fa8222836f73ee29c5380cf5c3bba791e8641420 ([pr477.01](https://github.com/shaavan/saving-satoshi/commits/pr477.01) -> [pr477.02](https://github.com/shaavan/saving-satoshi/commits/pr477.02), [diff](https://github.com/shaavan/saving-satoshi/compare/pr477.01..pr477.02))
Addressed @benalleng comment

- Used `keysMeta[progress].path` to set the href.
- Updated `isBetweenChapter` logic to be true only for the current chapter.
- Removed the redundant `progressChapterMetaMapper.tsx` file.

ui/chapter/Chapter.tsx
```diff
@@ -47,7 +48,13 @@ export default function Chapter({ children, metadata, lang }) {
   const chapter = chapters[metadata.slug]
   const position = metadata.position + 1
   const isEven = position % 2 == 0
-
+  const chapterLessons = keys.filter((ele) =>
+    ele.includes(progress.substring(0, 3))
+  )
+  const isBetweenChapter =
+    progress !== chapterLessons[0] &&
+    progress !== chapterLessons.pop() &&
```
benalleng, line 56: With the chapterLessons.pop() being equal to outro-2 if i quit the game on the end screen without continuing you progress remains at `CH1OUT2` but now i am prompted to restart the chapter

image

benalleng: I think you should remove the logic where `progress !== chapterLessons.pop()` until we add some logic that allows the outro screen to be automatically saved / skipped

benalleng (changes requested):
I like the idea of having the progress automatically skip past the outro-2 screen whereby i can quit the game, i.e. not having to press continue but it still shows me as having completed the chapter. However, we would need to include some logic to allow quitting the game to save progress at the end screen. Same issue as you brought up in #472 

> Say I am on CH2OUT1, and instead of pressing next, I go back to the home screen.
So when I reopen chapter selection, I will be directed to Chapter-2, instead of Chapter-3.

technically speaking their progress is still in CH2 `CH2OUT1` so we either need to
1. any action on the end screen saves to the next chapter, quitting, saving continuing
2. outro progress is not tracked, I am still directed to the page but the progress goes as such, `CH2MIN1` -> `CH3INT1`

ui/chapter/Chapter.tsx
benalleng: If we want the logic to allow me to quit at the outro screen we need to add some logic that allows quitting at the outro screen to also save or skip progress to the next lesson. However this should only be acheiveable if the user is logged in as they could jump past the logging in logic created for future lessons. Until then, I do not think that `progress !== chapterLessons.pop()` will work as expected

benalleng: With my progress at `CH1OUT2` I will still need to progress forward before the next chapter is unlocked, we can add some logic so that the outro screen progress is automatically skipped but we also need to consider in the first chapter we are expecting/requiring the user to log in so this logic would need to check if the user is logged in first

shaavan: > With the chapterLessons.pop() being outro-2 and my progress at CH1OUT2 if i quit the game on the end screen without continuing you progress remains at CH1OUT2 but now i am prompted to restart the chapter

You are right! That's an unintended behavior. Well, now I know what happens when you split development between two separate days 😅

> Until then, I do not think that progress !== chapterLessons.pop() will work as expected

I agree with your reasoning. Let me go ahead and remove this part of the code with the next update of PR.

shaavan:
> However, we would need to include some logic to allow quitting the game to save progress at the end screen.

This is one hurdle we need to figure out before allowing skipping outros.
Only after saving progress at the CH1OUT2, we shall logically be allowed to progress to CH2INT1. But skipping outros will create contradictory logic here.

> technically speaking their progress is still in CH2 CH2OUT1 so we either need to
> 
> Allow any action on the end screen saves to the next chapter; quitting, saving, and continuing
*This is most likely the easiest solution and the most straightforward, circumstantially it will also address the comment i quoted above
make sure outro progress is not tracked, I am still directed to the page but the progress goes as such, CH2MIN1 -> CH3INT1

Both seem reasonable approaches to go forward with. However, though solution-1 might be easier to implement, I prefer solution-2 for its logical simplicity.
As far as I can plan logic in my head, always skipping the last lesson of every chapter while tracking progress will not create any logical overhead for the rest of the codebase.

shaavan:
Updated from fa8222836f73ee29c5380cf5c3bba791e8641420 to bac8dd5e6e4f29eb6dff0ef11089a84dcec48383 ([pr477.02](https://github.com/shaavan/saving-satoshi/commits/pr477.02) -> [pr477.03](https://github.com/shaavan/saving-satoshi/commits/pr477.03), [diff](https://github.com/shaavan/saving-satoshi/compare/pr477.02..pr477.03))
Addressed @benalleng comment

Update:

- Removed the following logic.

```
isBetweenChapter = ... && progress !== chapterLessons.pop() && ...
```

ui/chapter/Chapter.tsx
benalleng: I opened a PR #478 so we remember

ui/chapter/Chapter.tsx
```diff
@@ -47,7 +48,12 @@ export default function Chapter({ children, metadata, lang }) {
   const chapter = chapters[metadata.slug]
   const position = metadata.position + 1
   const isEven = position % 2 == 0
-
+  const chapterLessons = keys.filter((ele) =>
+    ele.includes(progress.substring(0, 3))
+  )
+  const isBetweenChapter =
+    progress !== chapterLessons[0] &&
```
benalleng, line 55: We could add some logic to remove the continue button when the user is at the very Last lesson as there is nothing really to continue to. While the previous logic caused some progression issues this shouldn't have the same problem.

``` suggestion
progress !== keys[keys.length - 1]
```

benalleng (approved):
LGTM! I left a comment about an optional change based on the continue button on the very last lesson but I think that is an optional add.

shaavan:
Updated from bac8dd5e6e4f29eb6dff0ef11089a84dcec48383 to 74a06b5660ba978939db00ba84a08f027d820bd0 ([pr477.03](https://github.com/shaavan/saving-satoshi/commits/pr477.03) -> [pr477.04](https://github.com/shaavan/saving-satoshi/commits/pr477.04), [diff](https://github.com/shaavan/saving-satoshi/compare/pr477.03..pr477.04))
Addressed @benalleng comment

Changes:

- Updated `isBetweenChapter` logic to be false when the player is at the very last lesson in the game.

@benalleng 
> ...continue button on the very last lesson but I think that is an optional add.

Well if an optional add makes this PR better, I will always be glad to include it.
Thanks a lot, @benalleng, for your thorough review! 🙌

ecurrencyhodler:
tACK! I tried going back and from from chapters overview page to various challenges and event intro/outro. Worked like a charm. Great work!

Let's address the continue button on last challenge separately because I believe that is being discussed #480.

PR #479, opened 2023-05-27
Scroll position is retained on body when modal is opened

### LINKED ISSUE #284: Modal allows vertical scrolling in mobile
With the horizontal scrolling for the avatars there also can be some odd snappy vertical scrolling that looks like below

It has to do with the lower address bar in the mobile screen on firefox, so I'm not sure how to fix it ATM

Tested on Firefox on Android 

https://user-images.githubusercontent.com/108441023/228583943-76155fa3-d4c7-4506-b5d7-4ffb54801989.mp4

benalleng:
Closes #284 

This will prevent the body, not the modal from being scrolled in the background while the modal is open

This also makes the onClose() param of the Modal context remove that class whenever it is called


afterburn (changes requested):
Instead of repeating this code for every modal:
```ts
if (state.open) {
  document.body.classList.add('overflow-y-hidden')
}
```

We should add this to the `<Modal>` component so we have a single place to modify the behavior. This also makes it easier for people who will implement future modals.

afterburn (approved):

PR #482, opened 2023-05-29
Delete .github/workflows directory

ecurrencyhodler:
Github recently added the ability to auto-add issues to github projects natively through projects. We should move to using this tool so that we can avoid updating expiring PAT's.

I'm also deleting the workflow directory as a whole because the only workflow we are using was to add an issue to a github project through github actions.


ecurrencyhodler:
Also we should keep the workflow dir in the repo for the backend because the workflow in projects only allow the ability to pull in one repo not multiple.

PR #483, opened 2023-05-30
Adjust profile pic padding and sizing and fix chapter 2 outro screen

### LINKED ISSUE #481: Blank profile pic is too large and end chapter text on mobile needs to be adjusted
- The blank avatar icon is too big. You can find the dimensions [here in figma.](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=3634%3A37469&t=FVX8DBftixa21WXr-1)
- The layout (and some text details) of the chapter end screen is not quite right. The text covers the whole image, you can't really see what's going on. The text should be adjusted so it is lower on mobile as you can see in the 3rd screen.

benalleng:
Closes #481


ecurrencyhodler:
I think the text is low enough the ch2 outro on mobile but want to double check with @GBKS 

image

components/Navbar/UserButton.tsx
```diff
@@ -35,9 +35,9 @@ export default function UserButton() {
               <button
                 onClick={() => handleClick(Modal.SignIn)}
                 aria-label="profile"
-                className="text-grey-300 h-10 cursor-pointer"
+                className="text-grey-300 flex h-full w-full cursor-pointer items-center justify-center hover:bg-black/25"
               >
-                <Icon icon="avatar" />
+                <Icon icon="avatar" className="h-[30px] w-[30px]" />
```
afterburn, line 40: If I read this correctly the button could now potentially be bigger than the icon (since button uses `h-full` and `w-full` whereas the icon is a fixed `30px`). I think its better if we make sure that the button is a fixed `30px` instead, and have the icon adapt its sized based on the button so we can ensure that wherever the user clicks on the icon, the event handler fires.

components/Navbar/NavbarDesktop/NavbarDesktop.tsx
```diff
@@ -41,7 +41,7 @@ export default function NavbarDesktop({ params }) {
         <nav className="flex items-center">
           <TabGroup params={params} />
           <HelpLink params={params} />
-          <span className="flex h-full items-center">
+          <span className="h-[69px] w-16">
```
afterburn, line 44: Is there no way to use built-in tailwind classes instead of magic numbers?

components/Navbar/NavbarMobile/NavbarMobile.tsx
```diff
@@ -84,7 +84,9 @@ export default function NavbarMobile({ params }) {
         </div>
         <div className="flex items-center">
           <HelpLink params={params} />
-          <UserButton />
+          <span className="h-[62px] w-12">
```
afterburn, line 87: Here too.

components/Navbar/UserButton.tsx
GBKS: This HTML structure is getting a bit out of hand for the user button. 

image

You should be able to do this without any `div` and `span`. Main issue I see is that there's a mix of flexbox and non-flexbox layout techniques. If you use flexbox, the container can define how children should resize so you can avoid all the `w-full` and `h-full`. I'd really recommend looking into making the markup more lean. The best flex box guide I've found is the one by [CSS-Tricks](https://css-tricks.com/snippets/css/a-guide-to-flexbox/), been pulling that one up 1,000 times. But let me know if there are good reasons for all this nesting.

benalleng: > If I read this correctly the button could now potentially be bigger than the icon (since button uses h-full and w-full whereas the icon is a fixed 30px).

I went for this approach due to the user button being used both on the homepage and the lesson pages where the space for the user icon is much larger. I keeping with the buttons on the navbar in the lesson pages i tried to also apply the interactivity to the full space of the icon in the navbar to act as the button

afterburn:
@benalleng you have some conflicts

afterburn (approved):

PR #485, opened 2023-05-31
Nostr share on chapter completion

benalleng:
This is a test of sharing progress through nostr
It checks if you have a nostr extension and then adds a button to share via desktop
Mobile sharing is always visible

A few things to think about
- I chose the blastr relay `wss://nostr.mutinywallet.com` because we are a post-only entity and we want to maximize the places that this post gets seen. If the user wants we could give them an option to set their own relay somewhere down the line.

--------
### Nostr Content

image
Code your own way through the mysteries of bitcoin @savingsatoshi

--------

[Preview](https://saving-satoshi-git-fork-benalleng-nostr-support-savingsatoshi.vercel.app/en/chapters/chapter-1/outro-1?dev=true) (If you don't have a nostr extension, nothing will appear on outro-2 on desktop)


benalleng:
Receiving this build error:
`Type error: Property 'nostr' does not exist on type 'Window & typeof globalThis`
from this code
```
const handleShare = async () => {
  42 |     if (
> 43 |      typeof window.nostr !== 'undefined' &&
     |                    ^
  44 |       !shared &&
  45 |       progress === (keys[10] || keys[keys.length - 1])
  46 |     ) {

```
I understand that the nostr object exists in the window type because of the extension but how can i make sure that typescript is happy in this situation that i am respecting type safety?
I tried extending the widnow interface but I may not have done it correctly as that didn't seem to work correctly.

afterburn:
Hey @benalleng I retargeted your branch to `dev`, please resolve conflicts when you come back from your well deserved vacation 🌴

GBKS:
Here are some visuals users can share along with the copy. Just keeping it simple and re-using the chapter success images. A link to the site can be in the text that goes with the image.

nost-share-screens-230717

We should have a special message for the last one, when you complete the whole experience.

As far as the interaction, I don't think we need a modal. Let's just add a button and users can decide if they want to share (I really dislike annoying share pop-ups and hope we don't get into that habit with Nostr).

public/assets/images/nostr-share-chapter-1.jpg
GBKS, line 1: This image is 1MB. It's also a PNG, but with JPG as filename (wonder how that happened). If you change it to JPG and compress via something like [ImageOptim](https://imageoptim.com), you can get it down to 180KB without visible loss of quality. Nostr clients and users will thank you.

GBKS: Reason to take care of compression ourselves here is because (I think) Nostr clients will load those directly and React will not optimize. But let me know if I'm wrong on this.

benalleng: > This image is 1MB. It's also a PNG, but with JPG as filename (wonder how that happened).

Lol that's on me

benalleng:
@GBKS do you generally like the post format where it is "chapter-conpletion-image" => "text" => "savingsatoshi npub"

GBKS:
You're right, it's getting a bit big. But not sure that's a huge issue. Ideal would be to have a custom link with a specific preview image, could that be an option?

GBKS:
I like it. It does not work for me on mobile though. I see the button on both Android and iOS. On iOS, I get an error message and the button label changes. On Android, nothing happens when tapping. I assume it's trying to use some URI scheme and I don't have a matching app installed, right? How can we fix that?

Regarding the relays, an alternative to blasting would be to use the [window.nostr.getRelays()](https://github.com/nostr-protocol/nips/blob/master/07.md). I don't think all the extensions support this, so maybe just a thing to look into later?

I saw the image URL is set to share a 1920px wide image at 100% quality. Maybe change to 1200px and 75% quality for faster load times?

If we can fix the mobile issue, it'd be great to put this live soon.

benalleng:
Trying to think about how to check if a signer like Amber (Android) exists to only show it for that user just like on desktop

benalleng:
> Regarding the relays, an alternative to blasting would be to use the [window.nostr.getRelays()](https://github.com/nostr-protocol/nips/blob/master/07.md). I don't think all the extensions support this, so maybe just a thing to look into later?

definitely easy enough to do a fallback like `windwo.nostr.getRelays() || blastr relay`

benalleng:
Uploading signal-2023-11-28-161837_002.mp4…

GBKS:
For the general share button, you could also use the [webshare API](https://caniuse.com/web-share) for browsers that support it. It pulls up an OS-native share sheet that has contextual options based on the users contacts, installed apps, etc.

benalleng:
> [webshare API](https://caniuse.com/web-share) for browsers that support it

I dont have a reasonable way to implement and test this with my current setup

benalleng:
> could we say that the phone does not support Nostr sharing and point them towards a way to get that set up

Now that some browsers support mobile extensions this error handles both checking if the nostr.inent is undefined, i.e. the user does not have any nostr stuff on their device but also an error occurs if they simply reject the event whether on purpose or accidentally.

GBKS:
OK. Looking good. I'd make the X logo slightly smaller and change the button label to "Share via X". Seems more consistent with the other button. But then maybe let's merge this and see what happens?

benalleng:
You can visit 
https://saving-satoshi-git-fork-benalleng-nostr-support-savingsatoshi.vercel.app/_next/image?url=%2Fassets%2Fimages%2Fnostr%2Fnostr-share-chapter-1.jpg&w=1200&q=75 
to confirm that these links function with the correct hostname
or view it below noting the url 
image

components/Modals/SignIn.tsx
0tuedon, line 1: This change should'nt affect any other file I suppose?

0tuedon (approved):
Approved this only worried about the change of file name.

components/Modals/SignIn.tsx
benalleng: No, unlike an import these modals are controlled via the Modal context which already has this Signin name so changing this file doesn't change anything just gets the naming inline of the existing context

PR #486, opened 2023-05-31
Create Profile component

### LINKED ISSUE #444: Create a Profile component
For Chapter 3 we'll need a component that acts as a visual representation of a user. This component will be used for the coop mining challenge so it will need to support the following functionality:

For engineers:
1. Engineers need to be able to pass an avatar (string), a name (string), a description (string), and react children.
2. For the react children engineers should be able to both render the component as a standalone component (`<Profile />`) as well as wrap some other elements to support the variants listed in the design, i.e.:
```jsx
<Profile>
  // Children here
</Profile>
```

### Design
image

Variants:

image
image
image

sabakhilji:
Resolves #444

Created a profile component and tested it on the sign-up modal.

I think it needs more styling will do it after review.
Attached Below is the screenshot of the profile component rendered on the signup modal:

Screenshot (103)


sabakhilji:
@afterburn please review.

ecurrencyhodler:
closes #444

components/Modals/SignUp.tsx
```diff
@@ -12,6 +12,9 @@ import { generateKeypair } from 'lib/crypto'
 import { register } from 'api/auth'
 import { Input } from 'shared'
 import Modal from './Modal'
+{
```
vinayaksh42, line 15: @sabakhilji can you uncomment the code in SignUp.tsx so that the component is visible for testing in deployment link

shaavan:
@sabakhilji 

I looked through the code, and the structure of the code matches the requirements mentioned in the description of #444. Though it still needs styling, which can be addressed further down the line. 

However, the code might need further improvement, which Kevin will be the best person to point out.

A gentle ping to @afterburn. Could you please take a look at the code? Thanks!

components/utils/Profile.tsx
```diff
@@ -0,0 +1,24 @@
...
+  name: string
+  description?: string
+  children?: React.ReactNode
+}) {
+  return (
+    <div className="flex w-full flex-col items-center p-4">
+      <Avatar avatar={avatar} />
+      <div className="my-5 flex flex-col gap-4">
+        <span className="text-xl font-bold text-white">{name}</span>
+        <span className="text-xl font-bold text-white">{description}</span>
```
afterburn, line 19: You have marked the `description` prop as optional by using `?`, you are correct it should be optional. However you are _always_ rendering a `span` for it here. Instead we should only render this `span` if `description` has been set:

```jsx
{description && <span className="text-xl font-bold text-white">{description}</span>}
```

components/utils/Profile.tsx
```diff
@@ -0,0 +1,24 @@
...
+      </div>
+      <div className="w-full">{children}</div>
```
afterburn, line 21: The same applies to `children`.

afterburn (changes requested):
Good work @sabakhilji, left some comments.

sabakhilji:
> Good work @sabakhilji, left some comments.

Thanks!! I did the required changes.

components/Modals/SignUp.tsx
```diff
@@ -169,6 +172,8 @@ export default function LoginModal({ onClose, state }) {
           {!loading && t('modal_signup.confirm')}
         </button>
       </div>
+     
+      <Profile name='friend' description="good friend"></Profile>
```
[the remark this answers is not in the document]
sabakhilji: done!

afterburn (approved):

PR #490, opened 2023-06-01
Feedback button for end screen

### LINKED ISSUE #487: Share feedback link at the end of the game
We decided that the share feedback link to our google form should be only at the end of the game.

At the moment it is a text link but it would probably work best as @ecurrencyhodler suggested [here](https://github.com/saving-satoshi/saving-satoshi/issues/480#issuecomment-1570486317).

A button would allow us to more easily repeat it without changing much copy on the chapter end screens

benalleng:
Closes #487

Feedback link now a button and is only present on the final outro screen tracking the key array

### DIFF
--- a/ui/chapter/End.tsx
+++ b/ui/chapter/End.tsx
@@ -1,9 +1,11 @@
 'use client'
 
 import { Button } from 'shared'
-import { useSaveAndReturn, useTranslations } from 'hooks'
+import { useSaveAndReturn, useTranslations, usePathData } from 'hooks'
 import { Modal, useModalContext } from 'providers/ModalProvider'
 import { useAuthContext } from 'providers/AuthProvider'
+import { keys } from 'lib/progress'
+import { lessons } from 'content'
 import clsx from 'clsx'
 import Image from 'next/image'
 
@@ -20,9 +22,13 @@ export default function End({
 }) {
   const t = useTranslations(lang)
   const modals = useModalContext()
-  const saveAndReturn = useSaveAndReturn()
-
   const { account } = useAuthContext()
+  const saveAndReturn = useSaveAndReturn()
+  const { chapterId, lessonId } = usePathData()
+
+  const chapterLessons = lessons?.[chapterId]
+  const lesson = chapterLessons?.[lessonId]?.metadata ?? null
+  const currentLessonKey = lesson?.key ?? 'CH1INT1'
 
   const handleClick = () => {
     if (!account) {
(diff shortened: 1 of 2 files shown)


ecurrencyhodler:
On Chrome/iPhone, the text is smushed to the very top. But this should get resolved with #483 I think.

IMG_2716

But i do like how it is displayed in desktop
Screenshot 2023-06-01 at 8 58 59 AM

If the mobile issue will get resolved in #483 then it's a tACK for me.

ui/chapter/End.tsx
```diff
@@ -20,9 +22,13 @@ export default function End({
...
   const t = useTranslations(lang)
   const modals = useModalContext()
+  const { account } = useAuthContext()
   const saveAndReturn = useSaveAndReturn()
+  const { chapterId, lessonId } = usePathData()
 
-  const { account } = useAuthContext()
+  const chapterLessons = lessons?.[chapterId]
+  const lesson = chapterLessons?.[lessonId]?.metadata ?? null
+  const currentLessonKey = lesson?.key ?? 'CH1INT1'
```
vinayaksh42, line 31: shouldn't we change the CH1INT1 to CH1OUT2 so that if we are unable to fetch the current lesson data the feedback form is at least displayed in chapter 1?

Although upon testing I was unable to find a scenario in which the currentLessonKey was using 'CH1INT1'.

benalleng: this would occur if we decided to put the `end.tsx` component into `intro-1`. While this would never happen perhaps there is a better way to handle lesson being undefined

ecurrencyhodler (approved):
Ben says text issue should be resolved with #483 so approving it from a QA perspective.

ui/chapter/End.tsx
shaavan: The caller of the End.tsx component must always be a lesson with a defined lessonKey.

So if the currentLessonKey is `undefined` in some cases, it's best to pass a console error to report the discrepancy.

shaavan: I was thinking about,

What if the caller component passes its lesson key to the **End.tsx** component?
This IMO will be a better approach because:

1. The parent will directly inform the child component about its lesson key identity. This eliminates the need for checking the `lessonKey` identity by the child component.
2. It will make the **End.tsx** component more particular in its purpose.
3. It will provide a single source of truth for critical information such as lessonKey.

This shall be a better approach, in general, for all the child components.

benalleng: Ok! I feel like that change may be a bit outside the scope of this PR but I think it is a good possible issue to make into a new PR

afterburn (approved):
We can resolve the issue about the lesson key in another PR/issue.

PR #494, opened 2023-06-03
Replace REPL solution

### LINKED ISSUE #74: Investigation: Embed repl.it
We'd like to explore the possibility of using repl.it and need to investigate more. For this ticket, let's try embedding a REPL into this code base: https://docs.replit.com/hosting/embedding-repls

Here's a sample replit that can be used: https://replit.com/@satsie/Saving-Satoshi-Chapter-3-Challenge-2#main.py

Some questions a proof of concept would help answer:

- What does a REPL look and feel like when in the site?
- Do we have control over styling?
- Does it feel natural or alien in the context of our other challenges and UI design?
- Does it work on mobile?

Note that they also have an API so if we can't control the styling, we can still use the engine.

afterburn:
This PR closes #74 and removes our client-side REPL approach in favor of our new REPL API.

The design has not been approved by @GBKS yet but we can address that in a follow up PR.


shaavan (commented):
I experimented with the code, and I noticed a few points of improvement:

<h3> 1. The unusual placement of compiler options </h3>
Though JS is displayed on the right, the challenge first defaults to Python. We should move the Python option to the left or set Javascript as the default.

<h3>2. Unhelpful messages in the Python console.</h3>

Where the JS console is quite helpful in pointing out the error, the Python console is not very helpful in doing so.

**Reference:**
   
Screenshot 2023-06-05 at 5 04 02 PM
Screenshot 2023-06-05 at 5 04 07 PM

Whereas in [programiz](https://www.programiz.com/python-programming/online-compiler) compiler 

Screenshot 2023-06-05 at 5 04 18 PM

<h3> 3. Incorrect marking of wrong code </h3>

The scripting challenge is marked correct after some time, even when the code is wrong. This problem is present in both Python and JS.

**Reference:**

https://github.com/saving-satoshi/saving-satoshi/assets/85434418/5673426b-bbb6-4477-b078-7f95ccf1158a

<h3> 4. Improper functioning of UI elements </h3>

The Run script button does not reset back to the play button when the code fails due to a syntax error in Python.

**Reference:**

https://github.com/saving-satoshi/saving-satoshi/assets/85434418/a6ca6063-d737-42e4-8e3e-906b6c26e135

afterburn:
Thanks for reporting these @shaavan.
1. You're right, I will make JS the default for now.
2. I am aware, Python tracebacks are empty for some reason on the AWS instance, while locally they show up correctly. I've tried some fixes but none seem to work. Will address this in separate PR. This is the reason why I'm defaulting to JS in 1.
3. I've added some logic that will force a success after max_tries has been reached, seems like it needs some work still, will address in this PR.
4. Not sure what is going wrong here, will look into it.

afterburn:
@shaavan fixed 1 and 3, will address 2 and 4 in a separate PR dedicated to python, the issues are related.

shaavan (commented):
1. The JS is now set as the default compiler
2. Let’s address 2 and 4 in the separate PR.

<h3> Update on Number 3 </h3>

- I can still write the incorrect code and get marked correctly after a finite time.

**Reference:**

https://github.com/saving-satoshi/saving-satoshi/assets/85434418/d05e82ba-e746-4281-82bd-380bf40a47a6

- I can write elaborate code to never get a correct answer, yet I am marked correct. Instead of a maximum run limit, we should apply a maximum time limit. This automatically solves all the wrong code-related problems

**Reference:**

https://github.com/saving-satoshi/saving-satoshi/assets/85434418/9946992b-bc65-4368-a6ff-db78af7b3789

- Running a function that really finds a nonce with a hash starting with five zeros is quite doable by our compiler. This provides another reason for replacing the step limit with the time limit.

**Reference:**

https://github.com/saving-satoshi/saving-satoshi/assets/85434418/a4840f69-44b8-403f-9619-14f641dec483


New findings:

1. Renaming the function leads to an error. This seems to be the intended behavior. Still, I favored mentioning it.

---

I forgot to say so yesterday, but thank you very much, @afterburn, for your excellent work! Let the real coding begins!

afterburn:
@shaavan pushed some fixed, please review

afterburn:
@shaavan well let's unpack this a little bit, the challenge here is that the user needs to write a script that produces a correct sha256 hex to teach them how these hashes are being generated. The system actually verifies whether you are producing correct hashes in various ways (checking for type of string, checking for a length of 64 chars, etc), so in that sense it works.

You're right that at some point a solution will be forced because we've already verified that the user is producing valid hashes at that time and it becomes of question of letting them use valuable compute to keep hashing, or just force a solution. I opted for the latter.

shaavan:
@afterburn 

> It becomes a of question of should we let users keep hashing, which uses valuable compute, even though the lesson has already been learned? Or do we just force a solution if the produced strings match the criteria of a sha256 hash?

I agree with your logic and understand the reasoning behind taking this approach.
There's only one question in my mind. How much "hacky" are we expecting our users to be, and how robust do we want our system to be portrayed?

afterburn:
@shaavan I think the code you wrote in the last example is going into "I'm trying to break this on purpose" territory where as I'd say the examples before that are in "the system should be able to handle this properly" territory. Hope that answers your question.

shaavan:
> I think the code you wrote in the last example is going into "I'm trying to break this on purpose" territory where as I'd say the examples before that are in "the system should be able to handle this properly" territory. Hope that answers your question.

Yes, that answers my question. I think I went a little overboard on my "hacky" side 😅

The PR, in its current state, is at a good balance between efficient use of resources and robustness. Let me do one final code review before ACKing the PR!

As always, thanks for your fantastic work, @afterburn!

ecurrencyhodler:
Do you have recommended code I can input for these challenges? I'm at a loss at how to pass the scripting challenge.

ecurrencyhodler:
Let's have @adamjonas go try the scripting challenge in Ch. 2. If he approves we can merge.

adamjonas:
Python experience for me. This code worked in the demo that afterburn had showed me:

https://github.com/saving-satoshi/saving-satoshi/assets/755825/e3a30896-10c3-42bb-a189-00e9ffda7411

For reference the python code I used is:

```python
import hashlib

def find_hash(nonce):
    while True:
        # Combine the nonce with the current iteration number
        input_data = f"{nonce}{find_hash.counter}".encode()

        # Compute the SHA-256 hash of the input data
        hash_object = hashlib.sha256(input_data)
        hash_hex = hash_object.hexdigest()

        # Check if the hash starts with five zeroes
        if hash_hex.startswith("00000"):
            return hash_hex, find_hash.counter
        else:
            find_hash.counter += 1

# Initialize a counter variable as an attribute of the find_hash function
find_hash.counter = 0

# Test the find_hash function with a nonce
nonce = "my_random_nonce"
hash_result, iteration = find_hash(nonce)

print(f"Hash found: {hash_result}")
print(f"Required iterations: {iteration}")
```

JavaScript is a infinite loop as well:
Screenshot 2023-06-12 at 8 02 27 PM

js code:

```js
const crypto = require('crypto');

function findHash(nonce) {
  let counter = 0;
  
  while (true) {
    const input_data = `${nonce}${counter}`;
    
    const hash = crypto.createHash('sha256');
    hash.update(input_data);
    const hash_hex = hash.digest('hex');
    
    if (hash_hex.startsWith('00000')) {
      return {hash: hash_hex, iterations: counter};
    } else {
      counter++;
    }
  }
}

// Test the findHash function with a nonce
const nonce = "my_random_nonce";
const result = findHash(nonce);

console.log(`Hash found: ${result.hash}`);
console.log(`Required iterations: ${result.iterations}`);
```

ecurrencyhodler:
Awesome. Thanks Jonas!

afterburn:
@adamjonas I think the confusion comes from the fact that there is some code invisible to the user which actually calls the `find_hash`/`findHash` function (in a loop), so all the user needs to do is write the function body for `find_hash`/`findHash`.

You dont need to write a loop yourself, in fact, if you do, it probably wont work. I think this is not a problem of the REPL solution but more so a problem of how we communicate the goal with the user.

afterburn:
@adamjonas @ecurrencyhodler @GBKS here is a video of the current experience with working code for JS.

https://github.com/saving-satoshi/saving-satoshi/assets/10757768/36a284b1-47c7-4224-b5bc-a10a8a39d9b6

Some thoughts:
1. It's not immediately clear when I press run that the system first needs to spin up a docker container, so there is a delay before stuff starts appearing in the "Hasher" tab.
2. We should communicate more clearly to the user that they only have to write the function body, not an actual script that performs a loop. This is what the current script looks like:

To clarify @adamjonas, this is the actual script that runs:
```js
/* USER CODE */

function findHash(nonce) {
  return crypto.createHash('sha256').update(nonce.toString()).digest('hex')
}

/* INVISIBLE SYSTEM CODE */
const timeLimit = 3000;
const difficulty = 5;
let hashFound = false;
let nonce = 0;
let previousHash = ''
let previousHashMatchCount = 0

async function _runScript() {
  let hash = findHash(nonce)
  if (typeof hash !== 'string') {
    console.log('Error: findHash does not return a string')
    return console.log('KILL')
  }

  if (hash.length !== 64) {
    console.log('Error: findHash should return a string of 64 characters')
    return console.log('KILL')
  }

  function forceSolution() {
    hash = findHash(nonce + 1);
    hash = new Array(difficulty + 1).join("0") + hash.slice(0, hash.length - difficulty);
    console.log(hash);
    hashFound = true
  }

  setTimeout(() => {
    if (!hashFound) {
      forceSolution()
    }
  }, timeLimit)

  while (!hashFound) {
    hash = findHash(nonce);

    if (hash === previousHash) {
      previousHashMatchCount += 1
      if (previousHashMatchCount === 3) {
        console.log('Error: Your script keeps generating the same hash, make sure you use the nonce')
        break
      }
    }

    console.log(hash);
    previousHash = hash
    nonce++;
    await _sleep(50);
  }

  console.log('KILL')
}

function _sleep(t) {
  return new Promise((resolve) => {
    const timeout = setTimeout(() => {
      resolve();
      clearTimeout(timeout);
    }, t);
  });
}

_runScript();
```

afterburn:
Created a new issue for the above https://github.com/saving-satoshi/saving-satoshi/issues/513

ecurrencyhodler:
Apologies jonas. Let's raise this on the call today.

PR #499, opened 2023-06-08
Modal accesibility

### LINKED ISSUE #498: Modal avatar accessibility
Presently sign-up modal avatar pictures are not accessible via the narrator.
As per discussion with @GBKS and @afterburn wrapping element inside button.

sabakhilji:
Closes #498


ecurrencyhodler:
@sabakhilji your build is failing. Do you need help investigating?

sabakhilji:
Yes, @ecurrencyhodler!! I don't know the reason for the build fail, this PR addressed a very small change though.

components/Modals/SignUp.tsx
```diff
@@ -99,16 +99,23 @@ export default function LoginModal({ onClose, state }) {
...
-              key={i}
-              avatar={`/assets/avatars/${i}.png`}
-              size={80}
-              onClick={() => setAvatar(i)}
-              className={clsx('inline-block h-20 w-20 rounded-full border-2', {
-                'border-white': avatar === i,
-                'border-transparent': avatar !== i,
-              })}
-            />
+            <div>
```
GBKS, line 102: You'll need a few more tweaks here:
- Move the `onClick` event to the button. Otherwise, this is a button that does not do anything
- Add an [aria-pressed](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed) state that reflects whether this button is selected
- Add an [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) with something like "Select avatar 1" to let users know what happens when the button is pressed
- Remove the wrapping `div`, it is unnecessary

Did you check whether the button has any default styling applied, like a background or border? If so, you may need to remove those as well.

afterburn:
image

components/Modals/SignUp.tsx
sabakhilji: Thanks, @GBKS!!! to review, I am addressing the answers here according to the changes :

- I checked onClick worked perfectly, as we are passing it in the avatar component explicitly.
- I tried aria presses false and it worked perfectly but as HorizontalScrollView is a single component and we are wrapping the avatar and button inside it. Also, there is an array map in HorizontalScroll View, so every time when I am moving from one avatar to another even if I have already selected one of the images, the narrator declares button is not selected. So I tried the aria-label.
- I used aria-label="select one of the five avatar images" with the button please LMK if this is useful.(Please note narrator declares this 5 times).
- I removed the div element.

components/Modals/SignUp.tsx
```diff
@@ -99,16 +99,24 @@ export default function LoginModal({ onClose, state }) {
...
-              key={i}
-              avatar={`/assets/avatars/${i}.png`}
-              size={80}
-              onClick={() => setAvatar(i)}
-              className={clsx('inline-block h-20 w-20 rounded-full border-2', {
-                'border-white': avatar === i,
-                'border-transparent': avatar !== i,
-              })}
-            />
+            <button
```
benalleng, line 102: In case Kevin's comment wasn't clear this button needs a key as it is within the .map() function

components/Modals/SignUp.tsx
```diff
@@ -99,16 +99,24 @@ export default function LoginModal({ onClose, state }) {
...
+              aria-label="select one out of the five avatar"
```
benalleng, line 103: I think that this is perhaps a bit redundant for the aria label, perhaps using the key `i` or even setting a name for each individual avatar

benalleng (changes requested):

components/Modals/SignUp.tsx
GBKS: Yes, this should state something like "Select avatar 1", "Select avatar 2"....

Imagine a blind user selects these buttons one after the other. Whenever a button is selected, they need to know what pressing it does. So each button needs to indicate which one will be selected on click.

components/Modals/SignUp.tsx
```diff
@@ -99,16 +99,24 @@ export default function LoginModal({ onClose, state }) {
 
         <HorizontalScrollView>
           {[1, 2, 3, 4, 5].map((i) => (
-            <Avatar
-              key={i}
-              avatar={`/assets/avatars/${i}.png`}
-              size={80}
-              onClick={() => setAvatar(i)}
```
GBKS, line 106: The `onClick` function needs to go on the button. With a screen reader, the user can select the button, but pressing it doesn't do anything.

components/Modals/SignUp.tsx
GBKS: Thanks for addressing things, just left a few more specific comments.

Did you test with a screen reader? I just did here and while the avatars were able to get focus, I could not actually select them (meaning they would get the white outline).

ecurrencyhodler:
tACK! Great work getting this displayed! I think we need another code review b4 merging.

components/Modals/SignUp.tsx
```diff
@@ -102,16 +102,25 @@ export default function LoginModal({ onClose, state }) {
...
               key={i}
-              avatar={`/assets/avatars/${i}.png`}
-              size={80}
-              onClick={() => setAvatar(i)}
-              className={clsx('inline-block h-20 w-20 rounded-full border-2', {
-                'border-white': avatar === i,
-                'border-transparent': avatar !== i,
-              })}
-            />
+              aria-label={`select avatar${i}`}
```
GBKS, line 107:
```suggestion
              aria-label={`Select avatar ${i}`}
```
Capitalization and spacing.

components/Modals/SignUp.tsx
```diff
@@ -102,16 +102,25 @@ export default function LoginModal({ onClose, state }) {
...
+              aria-pressed="true"
```
GBKS, line 108:
```suggestion
              aria-pressed={avatar === i}
```
Only make it true when this specific avatar is selected.

content/lessons/chapter-2/mining-1.tsx
```diff
@@ -282,10 +282,10 @@ export default function Mining1({ lang }) {
                 {t('chapter_two.mining_one.paragraph_seven')}
               </div>
               <div className="mt-2 break-all border border-dashed border-white p-2.5 font-space-mono text-[15px] font-normal">
-                0000 0000 00
+                0000 2f22 2f
                 <span className="text-white text-opacity-50">
-                  72 947e 2f22 250f ac0d dd88<br></br> 2fcb f37c f6e2 340a 4112
-                  9b6r 23a2 823a
+                  cb 4112 0000 250f f37c 9b6r<br></br> 0072 ac0d f6e2 23a2 947e
+                  dd88 340a 823a
```
benalleng, line 288: I think this code is leftover from the `formatHash` issue. Lets go ahead and revert this code back to master. _copying and pasting the code from the left side should to the trick

benalleng (changes requested):
Thanks!, some leftover code needs to be removed as well as some small adjustments to the aria options as GBKS pointed out

benalleng (approved):
LGTM! nice work!

GBKS:
Sorry, this was not quite ready. The `onClick` function still has to be moved to the button, as I pointed out multiple times. It cannot be on the image.

When I use VoiceOver right now, I can highlight the different avatars. From a technical perspective, the button element is highlighted (not the image). But pressing does not do anything. That's because the `onClick` is not on the button. 

`key={i}` can also be removed from the Avatar, since the mapping happens on the button element.

ecurrencyhodler:
Hey Christoph. I created another issue based on your comment. Apologies for merging prematurely. Saba you can try to address it again if you want while you wait for #509 to be unblocked. If not, we can backlog this to the next cycle or wait for someone else to pick it up.

benalleng:
That's my bad, I thought it was resolved. I can pick it up.

PR #514, opened 2023-06-13
Fix: mining stall bug

### LINKED ISSUE #471: Ch2 Challenge 3 Mining Stall
In ch. 2 Challenge 3, I go to the module where I turn the miners back on. After a moment I see the 10x button flashing. But I didn't press it right away cuz I got distracted. When I did, the challenge stalled and I was unable to continue because the miners stopped. I believe this is a result of the challenge being time based.

https://github.com/saving-satoshi/saving-satoshi/assets/32662508/49091794-fe41-4b01-885f-8d424955594d

vinayaksh42:
Fixes #471


ecurrencyhodler:
I tested this. I waited for over a minute before pressing the 10x button. I also went to different tabs. The challenge still worked. I'm giving this a tACK although I'm still not 100% confident that the bug won't reappear because I did similar tests in prod and it also seemed to work.

PR #519, opened 2023-06-13
Update: Change 1000 block to 100 block

### LINKED ISSUE #517: Change # of blocks in Ch.2 mining to 100 blocks
Let's update the # of blocks to 100 instead of 1,000. 100 is more realistic.

vinayaksh42:
Fixes #517


ecurrencyhodler:
tACK! Checked block #'s too.

shaavan (commented):
The block number is updated correctly.

However, since the blocks number are reduced by a factor of 10, the motion of the progress bar does not feel as smooth as with 1000 blocks.

https://github.com/saving-satoshi/saving-satoshi/assets/85434418/5b3ff7e1-858f-4846-bfa5-632951575d94

The start-stop motion of the progress bar sticks out from the rest of the animations on the page.

@vinayaksh42 
What do you think about going with a smoother motion for the progress bar

Also, a gentle ping to @GBKS. What do you think about this point?

GBKS:
Looks really nice. All the extra subtle transitions bring this to life. A few notes:

- Transitions are missing when reaching 100 blocks. The text instantly switches out and the highlight borders instantly appear, etc. Can we make all of these consistently smooth?
- Is it possible to randomize the time in between finding blocks? That would be more realistic. That would also resolve the point by @shaavan.

I wonder if we should update the initial copy. We state that we assume a consistent 3,500 transactions per block. Should we add that we also assume consistent transaction fees?

vinayaksh42:
> Looks really nice. All the extra subtle transitions bring this to life. A few notes:
> 
> * Transitions are missing when reaching 100 blocks. The text instantly switches out and the highlight borders instantly appear, etc. Can we make all of these consistently smooth?
> * Is it possible to randomize the time in between finding blocks? That would be more realistic. That would also resolve the point by @shaavan.
> 
> I wonder if we should update the initial copy. We state that we assume a consistent 3,500 transactions per block. Should we add that we also assume consistent transaction fees?

I have fixed the above mentioned issues. While fixing the stalling bug the fade in fade out for step 4 was accidentally removed, I have added it back. The time taken to find a block is also random now.

ecurrencyhodler:
tACK! WRT to Christoph's comment, I agree we should make a small mention of making fees consistent. Can we update the copy to the following?

"For this simple simulation, we’ll just assume that every block contains 3,500 transactions and 0.061 BTC in fees."

Then let's merge.

Screenshot 2023-06-14 at 8 21 28 AM

GBKS:
@ecurrencyhodler shouldn't the copy read "rewards and fees" instead of just "fees"?

I just tested a few times and it looks good overall. There are some very minor glitches in there. In one of my tests, the 10x button never lit up. In other tests, the transitions did not smoothly complete. Looked like the text faded out, but the transition got cut off half-way through, the text instantly disappeared, and the new one showed up. Maybe something around mismatched intervals and transition durations?

GBKS:
Just tested again and saw the flickering in Mac/Safari. I think it's because the CSS fade is set to 300ms, and the code interval that triggers the fade out, text change, and fade in, is also set to 300. Browsers are not always that precise, so the slightest mismatch can lead to these not lining up. Better to give the browser a little more time to wrap up the animation, and then trigger the next code change (sometimes computers need to be spoonfed). For example, try setting the code intervals to 325ms.

Another small thing is that React throws errors in the console because of `stroke-width` properties in `Cross.tsx` and `GreenCheck.tsx`. Could you please rename those to `strokeWidth`?

There's also an issue with the lesson page `head` tag being rendered inside the page content and not as a top-level tag (`[lang]/chapters/[slug]/[lesson]/page.tsx` line 66). I assume there's some central way in React to change the page title rather than rendering another title tag somewhere in the body (which is not valid HTML). That can be looked at in a different PR and is not high priority, IMHO, it just throws an ugly error.

vinayaksh42:
> 

Thanks a lot for all the suggestions. I have changed the interval timing to 325ms, I am unable to see the flickering in chrome and safari. Changed the `stroke-width` to `strokeWidth` for `GreenCheck.tsx` and `Cross.tsx`.

GBKS:
Here's [an example recording](https://www.dropbox.com/s/ojqh3prmqxrsnjc/saving-satoshi-flicker-230620.mp4?dl=0) of the flickering.

GBKS:
I still see the flickering. After taking a closer look I realized it's due to the use of animation. How animations work is that once they have played, they reset to the beginning. Let's say you have a fade out that takes 300ms, and it starts at opacity 1 and animates to opacity 0. After 300ms, it goes back to opacity 1. Because Javascript and CSS timing is not 100% in-sync, we sometimes get this situation.

Solution is to use the [fill-mode](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-fill-mode) `forwards` in your animation declaration, for example:

`animation: fadeIn 0.3s ease-in-out forwards;`

Result is that after 300ms, it doesn't revert, but stays on the end state of the animation.

Also, you need to revert the change in globals.scss from `0.3s` to `0.4ms`. It has to be `0.3s`. Otherwise the animation is longer than the JS timeout you defined and gets cut off again.

ecurrencyhodler:
I tested all of challenge 3 and didn't see any flickering. Here's recording of the beginning part of the challenge. 

Will let christoph review before merging.

https://github.com/saving-satoshi/saving-satoshi/assets/32662508/da410345-5b5d-4917-80c7-ffbd18471a94

GBKS:
Flickering is still there. You still need the Javascript delay to be longer than the CSS animation as previously mentioned. You can make it 325ms.

Small thing on mobile layouts is that the check icon and the x icon next to the small hash have different widths, so there's a glitchy effect happening when the check briefly appears. Should be straightforward to make the widths the same.

vinayaksh42:
> Flickering is still there. You still need the Javascript delay to be longer than the CSS animation as previously mentioned. You can make it 325ms.
> 
> Small thing on mobile layouts is that the check icon and the x icon next to the small hash have different widths, so there's a glitchy effect happening when the check briefly appears. Should be straightforward to make the widths the same.

Changed 300ms to 325ms, unable to notice any flickering on chrome/safari/firefox on macOS.
The icon issue was actually related to the different width size of hash for mobile view. The glitchy effect won't happen now.

ecurrencyhodler:
I'm looking to merge this tomorrow. If there still is a flickering issue, we can resolve it in a separate pr.

PR #522, opened 2023-06-14
Improve scripting challenge UX #513

### LINKED ISSUE #513: Improve Scripting Challenge UX
1. It's not immediately clear when I press `Run script` that the system first needs to spin up a Docker container (system monitor tab), so there is a delay before stuff starts appearing in the "Hasher" tab.
2. We should communicate more clearly to the user that they only have to write the function body, not an actual script that performs a loop. This is because behind the scenes there is already a program written that calls the `findHash` function, so users just need to implement the function. (full code below video)
3. If there are errors we should automatically jump to the Console tab (which logs the errors).

Video reference:

https://github.com/saving-satoshi/saving-satoshi/assets/10757768/0f53ae42-aef1-46cb-a950-7536f0d60f03

```js
/* USER CODE */

function findHash(nonce) {
  return crypto.createHash('sha256').update(nonce.toString()).digest('hex')
}

/* INVISIBLE SYSTEM CODE */
const timeLimit = 3000;
const difficulty = 5;
let hashFound = false;
let nonce = 0;
let previousHash = ''
let previousHashMatchCount = 0

async function _runScript() {
  let hash = findHash(nonce)
  if (typeof hash !== 'string') {
    console.log('Error: findHash does not return a string')
    return console.log('KILL')
  }

  if (hash.length !== 64) {
    console.log('Error: findHash should return a string of 64 characters')
    return console.log('KILL')
  }

  function forceSolution() {
    hash = findHash(nonce + 1);
    hash = new Array(difficulty + 1).join("0") + hash.slice(0, hash.length - difficulty);
    console.log(hash);
    hashFound = true
  }

  setTimeout(() => {
    if (!hashFound) {
      forceSolution()
    }
  }, timeLimit)

  while (!hashFound) {
    hash = findHash(nonce);

    if (hash === previousHash) {
      previousHashMatchCount += 1
      if (previousHashMatchCount === 3) {
        console.log('Error: Your script keeps generating the same hash, make sure you use the nonce')
        break
      }
    }

    console.log(hash);
    previousHash = hash
    nonce++;
    await _sleep(50);
  }

  console.log('KILL')
}

function _sleep(t) {
  return new Promise((resolve) => {
    const timeout = setTimeout(() => {
      resolve();
      clearTimeout(timeout);
    }, t);
  });
}

_runScript();
```

afterburn:
This PR closes #513

### DIFF
--- a/app/[lang]/chapters/[slug]/[lesson]/page.tsx
+++ b/app/[lang]/chapters/[slug]/[lesson]/page.tsx
@@ -1,6 +1,7 @@
 'use client'
 
 import { useEffect, useState } from 'react'
+import { createPortal } from 'react-dom'
 
 import { chapters, lessons } from 'content'
 import { usePathData, useTranslations, useLocalizedRoutes } from 'hooks'
@@ -19,7 +20,18 @@ import { useProgressContext } from 'providers/ProgressProvider'
 import { Loader } from 'shared'
 import { LoadingState } from 'types'
 import { notFound } from 'next/navigation'
-import Client from 'components/Client'
+
+const Portal = ({ children, id }) => {
+  const [mounted, setMounted] = useState(false)
+
+  useEffect(() => {
+    setMounted(true)
+
+    return () => setMounted(false)
+  }, [])
+
+  return mounted ? createPortal(children, document.getElementById(id)) : null
+}
 
 export default function Page({ params }) {
   const searchParams = navigation.useSearchParams()
@@ -62,11 +74,9 @@ export default function Page({ params }) {
     const title = lesson ? t(lesson.metadata.title) : 'Page not found'
 
     return (
-      <Client>
-        <head>
-          <title>{`${title} - Saving Satoshi`}</title>
-        </head>
-      </Client>
+      <Portal id="head">
+        <title>{`${title} - Saving Satoshi`}</title>
+      </Portal>
     )
   }
 
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -41,12 +41,12 @@ export default function RootLayout({
 }) {
   return (
     <html lang="en">
-      <head>
+      <head id="head">
         <script
           type="application/ld+json"
           dangerouslySetInnerHTML={schema()}
           key="product-jsonld"
-        />
+        ></script>
       </head>
       <body className="font-cbrush">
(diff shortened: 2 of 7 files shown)


afterburn:
Scripts for testing:

```js

let nonce = 0
let found = false

while(!found) {
    const hash = crypto.createHash('sha256').update(nonce.toString()).digest('hex')
    console.log(hash)

    if (hash.startsWith('000')) {
        found = true
    }

    nonce++
}
```

```python
from hashlib import sha256

nonce = 0
found = False

while not found:
    hash = sha256(str(nonce).encode('utf-8')).hexdigest()
    print(hash)

    if hash.startswith('000'):
        found = True
    
    nonce += 1
```

ecurrencyhodler:
I tested using the JS code 6 times. 3x's it finds a hash instantly. 3x's times it stalls like this in the video. I'm not sure if that's intended.

https://github.com/saving-satoshi/saving-satoshi/assets/32662508/7144f306-e206-41d8-9863-7d1fc6a3fa27

In the off chance it takes longer than 15 seconds to find a solution, I like [Kevin's suggestion](https://github.com/saving-satoshi/saving-satoshi/issues/513#issuecomment-1591387614) of a button to skip through the challenge. If the code runs, it means they've successfully completed the challenge. 

But if we're surfacing a skip button at the 15 second mark, what would the difference between that and forcing the UI to just say they found a block?

afterburn:
This is still a draft so please dont test yet, I have some more stuff to do first. I dont think we should be allowing scripts to run endlessly, there should be a timeout at some point to free up compute for other users.

IMO 15 seconds is too long, not scalable, but this obviously poses some problems with this challenge since it can take a while to find a hash starting with x zeroes. This is why I dropped from 5 to 3 zeroes, but even then in some cases it doesnt find a hash in the alotted time (currently set to 5 seconds).

Needs some more fine tuning to get right.

ecurrencyhodler:
Tested this on chrome and firefox on Mac and works great. Did it both for python and JS. Finding a solution was fairly quick across 3 tries per language. Tested it 6 times total.
https://github.com/saving-satoshi/saving-satoshi/assets/32662508/5d006459-8dd2-41a9-a7f9-eb09b98e598f

PR #527, opened 2023-06-16
New miningStatisxtics variants

### LINKED ISSUE #448: Expand MiningStatistic component functionality
In Chapter 3 we'll want to use the same `MiningStatistic` component that @vinayaksh42 created in  #349 but expand its functionality so it supports the variants we need.

### Design
image

Variants:
image
image
image
image

benalleng:
Closes #448 

Currently creates several new variants to work with in future challenges for both the hash power and blocks found

### DIFF
--- a/content/lessons/chapter-2/mining-1.tsx
+++ b/content/lessons/chapter-2/mining-1.tsx
@@ -188,23 +188,11 @@ export default function Mining1({ lang }) {
     hydrated && (
       <div className="grid grid-cols-1 justify-center justify-items-center md:my-auto md:flex md:flex-row">
         <div className="fade-in grid w-full grid-cols-1 items-center px-[15px] py-[25px] md:order-last md:my-0 md:mx-[30px] md:w-[405px] md:p-[25px]">
-          <div
-            className={clsx(
-              'relative mb-2.5 font-nunito text-lg font-semibold',
-              {
-                'fade-in text-white': blocks !== 0,
-                'text-black/50': blocks === 0,
-              }
-            )}
-          >
-            <span className={clsx({ 'fade-in text-[#EDA081]': blocks !== 0 })}>
-              {t('chapter_two.mining_one.progress_bar_title')}
-            </span>{' '}
-            <span className="absolute right-0">
-              {Intl.NumberFormat().format(blocks)} of 1,000
-            </span>
-          </div>
-          <ProgressBar progress={blocks / 10} />
+          <ProgressBar
+            total={1000}
+            title={t('chapter_two.mining_one.progress_bar_title')}
+            progress={blocks}
+          />
           <MiningStatisticNonce
             title={t('chapter_two.mining_one.progress_bar_one')}
             content={nonce}
--- a/content/lessons/chapter-2/mining-1.tsx
+++ b/content/lessons/chapter-2/mining-1.tsx
@@ -8,6 +8,8 @@ import {
   MiningStatisticHash,
   ProgressBar,
 } from 'ui'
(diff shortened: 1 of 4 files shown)


afterburn:
Not sure if its just the screenshots but the text looks kind of dark compared to whats in figma:

image
image

ecurrencyhodler:
@afterburn can you take a look at this PR to review?

benalleng:
the disabled text's color can now be chosen though it remains fixed at 25 opacity

content/lessons/chapter-2/mining-1.tsx
```diff
@@ -188,45 +190,43 @@ export default function Mining1({ lang }) {
...
-            </span>{' '}
-            <span className="absolute right-0">
-              {Intl.NumberFormat().format(blocks)} of 1,000
-            </span>
-          </div>
-          <ProgressBar progress={blocks / 10} />
+          <BlockProgress
+            variant={BlockProgressVariant.TotalBar}
+            fontSize="text-lg"
+            disabled={blocks === 0 && 'text-black'}
```
afterburn, line 196: It doesn't make a lot of sense to pass a string to the `disabled` prop, it should be a `boolean`, this is how this prop is commonly being used when creating core components. If you want to apply different css classes based on the disabled state you should do something like this instead:

```tsx
<BlockProgress
  className={clsx({ 'text-black': blocks === 0 })}
  disabled={blocks === 0}
  ...
/>
```

afterburn: Same applies to the other places where you do this.

ui/common/BlockProgress.tsx
```diff
@@ -0,0 +1,96 @@
...
+        )}
+      >
+        <span
+          className={clsx('flex-grow', {
+            'fade-in text-[#EDA081]': progress !== 0,
+          })}
+        >
+          {title}
+        </span>{' '}
+        {variant === BlockProgressVariant.Percentage && percentage && (
```
afterburn, line 55: What happens if `percentage` is `0`, since `0` is considered a falsy value?

ui/common/BlockProgress.tsx
```diff
@@ -0,0 +1,96 @@
...
+                'top-0 left-0 h-full origin-left transform-gpu rounded-[5px] bg-[#FBEBC6] transition-transform',
+                {
+                  '!bg-[#5DBC59]': progress === total,
+                }
+              )}
+              style={{
+                transform: `translate3d(-${
+                  progress <= 5 && progress > 0
+                    ? (total - 5) / 10
+                    : (total - progress) / 10
```
afterburn, line 87: I don't understand what is happening here. `progress` should take a value between either `0` and `1` or `0` and `100` for it to make sense to people who arent familiar with this code. I'm not sure why you are subtracting `5` and dividing by `10`.

benalleng: This was leftover from when the total was 1000, we added this to make the first increment of the progress bar to be noticeable with 1000 total increments. I think its ok for us to remove

afterburn (changes requested):
Hey @benalleng, thank you for the effort. I don't think this is the right approach though.
By introducing all these kinds of variants we lose flexibility/composability. I was thinking we should do something more along these lines:

https://codesandbox.io/s/core-components-3tjy3n

benalleng:
Ok! just having a blank slate instead is a much simpler approach

benalleng:
I added a new MiningCard that allows us to create endless statistics cards

benalleng:
Closing in favor of #535

PR #530, opened 2023-06-22
Introduce Chip Component

### LINKED ISSUE #508: Create Chip component
In Chapter 3 we need to create a `Chip` component we can use to display some data on top of user avatars:
Lets create a component that can support the variants shown below:

image
image

[Link to Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=4228-70031&t=eCkKSeUA1EDezOwi-4)

shaavan:
Resolves #508

I have temporarily added the Chip component to the lessons page for easier component testing.


benalleng:
Any desire to add a pointer-cursor to indicate this component is clickable?

shaavan:
Updated from 8150bded74e99abf3966cfef9c5171506ec57b68 to 387639dcf4aff634b4c319771be747434f26c49c ([pr530.01](https://github.com/shaavan/saving-satoshi/commits/pr530.01) -> [pr530.02](https://github.com/shaavan/saving-satoshi/commits/pr530.02), [diff](https://github.com/shaavan/saving-satoshi/compare/pr530.01..pr530.02))
Addressed @benalleng comment

- Changed the cursor to cursor pointer when hovering over the component.

app/[lang]/chapters/[slug]/[lesson]/layout.tsx
```diff
@@ -14,6 +15,11 @@ export default function Layout({ children, params }) {
     <div className={`${theme} flex flex-col`}>
       <div className="fix-grow-issue flex min-h-screen flex-col overflow-hidden">
         <Navbar params={params} />
+        <Chip
+          position={1}
+          coins={1.49}
+          img={'/assets/images/chapter-1-holocat.jpg'}
+        />
```
Comment (afterburn) [inline app/[lang]/chapters/[slug]/[lesson]/layout.tsx:22]: Let's leave it up to the Chip's parent to determine what it renders:
```jsx
<Chip img={...}>
  <span>#1</span>
</Chip>
```

Where both `img` and children should be optional.

afterburn (changes requested):
Thanks @shaavan left a comment.

app/[lang]/chapters/[slug]/[lesson]/layout.tsx
Comment (shaavan) [reply app/[lang]/chapters/[slug]/[lesson]/layout.tsx:22]: Sounds like a good idea to make the component more generalized.

However, the component, as per our primary use case, needs to shuffle between two states, with different designs for the states.

https://github.com/saving-satoshi/saving-satoshi/assets/85434418/1d1f3089-2b58-4dd0-936f-4b56ebc47dee

So in making the component more generalized, it might become difficult to keep track of which child span of the Chip component should be displayed and with which color and state combination.

shaavan:
@afterburn, @GBKS
A gentle ping!

app/[lang]/chapters/[slug]/[lesson]/layout.tsx
Comment (afterburn) [reply app/[lang]/chapters/[slug]/[lesson]/layout.tsx:22]: I understand what you are saying but in that case lets create a separate component for that while the Chip component just does what I mentioned. This separate component would then use the Chip component, so:

```jsx
export default function Chip({ children }: { children?: React.ReactNode }) {
  return <div className="chip-classes">{children}</div>
}

export default function ProfileChip({ ranking, balance, img }: { ranking: number; balance: number; img: string }) { 
  // Do what you are currently doing in your Chip.tsx here.
  return <Chip>
    ...
  </Chip>
}
```

This will allow us to reuse the Chip component for other stuff.

afterburn (changes requested):
Hey @shaavan, replied on your comment

app/[lang]/chapters/[slug]/[lesson]/layout.tsx
Comment (shaavan) [reply app/[lang]/chapters/[slug]/[lesson]/layout.tsx:22]: Seems like a logical approach to go with.
Let me implement it!

shaavan:
Updated from 387639dcf4aff634b4c319771be747434f26c49c to 8d7a510fc33b68387f4b194c64a6cec96d4181f0 ([pr530.02](https://github.com/shaavan/saving-satoshi/commits/pr530.02) -> [pr530.03](https://github.com/shaavan/saving-satoshi/commits/pr530.03), [diff](https://github.com/shaavan/saving-satoshi/compare/pr530.02..pr530.03))
Address @afterburn comment

- Made the Chip component generalized and introduced the ProfileChip component for specific use cases.

afterburn (approved):

PR #534, opened 2023-06-27
ContributionBar Component

### LINKED ISSUE #446: Create "ContributionBar" component
For Chapter 3 we'll need to create a progress bar component to visually represent how much each user has contributed to the overall hashpower of the mining pool, a "contribution bar". Additionally there needs to be a way to specify a group of opposing miners (bitrey). We need to support the following functionality:

For users:
1. Users should be able to hover over the bars and a tooltip should appear which displays content that can be passed from the parent component, in the case of the mining challenge this will show the amount of blocks 'mined'.

For engineers:
1. Engineers should be able to pass an array of objects to the component where each object contains a user (object), a color (string) and a percentage (number) for both the protagonists (user + friends/npcs) and antagonists (bitrey). The user object itself consists of a username (string) and an avatar (string). The antagonists array should be optional.
2. Engineers should be able to pass a render function for the tooltip mentioned above, in the case of the mining challenge this tooltip will be showing the amount of blocks per user so the render function should take the `percentage` as an argument so we can get the amount of blocks by taking a `percentage` of total blocks specified in the parent.
3. We should render the progress bars using css transforms exclusively, we should not use css attributes like `width` for animation and instead try to do as much as possible on the GPU. See comments [here](https://github.com/saving-satoshi/saving-satoshi/pull/349#discussion_r1174134949) and [here](https://github.com/saving-satoshi/saving-satoshi/pull/349#discussion_r1193190852).

### Design
image
image

vinayaksh42:
Fixes #446

(Prop details) ContributionInfo:
- username (string): The username of player/npc/antagonists(bitrey)
- avatar (string): The avatar identifier associated with the user.
- color (string): The color of progress bar associated with the user.
- percentage (number): The percentage value associated with the progress of the user
- side (number): The side indicator associated with the user, where 0 represents right side (bitrey) and 1 represents the left side (player/npc).

example:
```js
contributionInfo={[
              {
                username: 'player',
                avatar: '1',
                color: 'EDA081',
                percentage: blocks / 10,
                side: 1,
              },
              {
                username: 'enemy',
                avatar: '3',
                color: '5DBC59',
                percentage: blocks / 10,
                side: 0,
              },
            ]}
```

To-Do:
- [ ] Add support for multiple NPCs on left side of the bar. Having some issues regarding this, one method could be to give an offset to the team member at the first position and adjusting the z-index accordingly.
- [ ] Render Function for tooltip

A few things to think about:
- [ ] Error handling for overlapping progress bars. Do we have to manage it at the component level or do we have to take care of it when we are providing the values to be displayed in the ContributionBar.


vinayaksh42:
preview: https://saving-satoshi-git-fork-vinayaksh42-issue446-savingsatoshi.vercel.app/en/chapters/chapter-2/mining-1?dev=true

it is currently linked to the blocks found of mining challenge for demonstration.

GBKS:
Looking good. I think it will look a little better if the line at the center is dotted and not dashed (due to the lack of control of dash spacing in CSS as well as unclean anti-aliasing).

As far as the 150ms transition duration for when blocks change width, it does look a bit jittery right now. This should be fine-tuned with the interval at which the component gets updates from the parents for a smoother effect. If there are constant updated, a duration may not even be necessary.

ui/common/ContributionBar.tsx
```diff
@@ -0,0 +1,72 @@
...
+              className={`absolute h-full w-full overflow-hidden rounded-[5px] `}
+            >
+              <div
+                className={`absolute inset-0 flex transform-gpu border-r-2 border-black transition-transform`}
+                style={{
+                  transform: `translate3d(${
+                    info.side === 0 ? 100 - info.percentage : -(100 - value)
+                  }%,0,0)`,
+                  backgroundColor: `#${info.color}`,
+                  zIndex: info.zindex,
```
benalleng, line 39: Not critical but could this just be handled by checking the index inherent to the map like `zIndex: {i * 10}`

benalleng (changes requested):
Looking good, two questions, one about the zindex needing to be an argument and then with the initial load I see the three avatars having unusual starting positions.



Is this due to their percentage arguments in the ContributionInfo object or something else?

ui/common/ContributionBar.tsx
```diff
@@ -0,0 +1,72 @@
...
+  let value = 0
+  return (
+    <div className="relative h-[30px] w-full rounded-[5px] bg-black/20">
+      <div className="absolute z-50 flex h-full w-1/2 border-r-2 border-dotted border-white/50"></div>
+      {contributionInfo.map((info, i) => {
+        value += info.percentage
+        return (
+          <>
+            <div
+              className={`absolute h-full w-full overflow-hidden rounded-[5px] `}
```
afterburn, line 30: These can just be strings instead of js objects

ui/common/ContributionBar.tsx
```diff
@@ -0,0 +1,72 @@
...
+              className="absolute inset-0 z-50 flex transform-gpu items-center justify-center transition-transform"
+              style={{
+                transform: `translate3d(${
+                  info.side === 0
+                    ? (100 - info.percentage) / 2
+                    : -(
+                        (100 -
+                          (value - info.percentage + info.percentage / 2) * 2) /
+                        2
+                      )
```
afterburn, line 53: This seems odd, could you explain what it does exactly?

ui/common/ContributionBar.tsx
vinayaksh42: Yes, we can do that. The unusual balance of avatar position is due to the value provided by the parent component. It can adjusted.

ui/common/ContributionBar.tsx
vinayaksh42: This is for determining the position at the which the avatar icon needs to be at with respect to the position of progress bar.
The default expression is:
`(100 - value) / 2`
let's say we have the `value = 20` which means 20 % of the progress bar.
The expression will give the value:
`(100 - 20) / 2 = 40`
Now, the avatar will be positioned on the centre after leaving an offset of 40 towards the right side.
Screenshot 2023-07-02 at 5 47 25 PM

The above expression works for a single avatar. Now, for multiple progress bar the expression becomes:
`[ 100 - {(PreviousValue + CurrentValue / 2) * 2}] / 2`
Basically the `value` is replaced by `{(PreviousValue + CurrentValue / 2) * 2}` 
Here is how the math for this one works:
Screenshot 2023-07-02 at 5 52 28 PM

afterburn: Thank's for the thorough explanation, while this works I think the following approach is a little cleaner and easier to understand: https://codesandbox.io/s/contribution-bar-5jy252?file=/src/ContributionBar.js I also took the liberty to rename some variables. It could use some further clean up (currently there's 2 loops doing basically the same thing) but I'll leave that up to you.

afterburn (changes requested):
Left a comment here: https://github.com/saving-satoshi/saving-satoshi/pull/534#discussion_r1249551613

ui/common/ContributionBar.tsx
GBKS: Looks a bit overly complicated. You could just use a flex layout and set a `flex-basis` percentage on each box, with the last element having `margin-left: auto`, which pushes it to the right. Give the box `position:relative` and center the image with `position:absolute; left: 50%; top: 50%; transform: translate(-50%, -50%); z-index: 2;`. My 2 cents, you do what works best for you.

ecurrencyhodler:
Hey so Vinayak is feeling pretty sick. 

@afterburn, Vinayak and I think we should merge this pr and create an issue to [simplify the logic](https://github.com/saving-satoshi/saving-satoshi/pull/534#discussion_r1249551613) in a separate PR. Does that sound good to you?

Vinayak go ahead and clean up the conflicts so we can get this merged.

afterburn:
> Hey so Vinayak is feeling pretty sick.
> 
> @afterburn, Vinayak and I think we should merge this pr and create an issue to [simplify the logic](https://github.com/saving-satoshi/saving-satoshi/pull/534#discussion_r1249551613) in a separate PR. Does that sound good to you?
> 
> Vinayak go ahead and clean up the conflicts so we can get this merged.

That's fine by me, lets resolve the conflicts, then I'll merge.

Issue #524, opened 2023-06-15
Ch.2 Title text overflow


ecurrencyhodler:
Looks like the title text "hashing out a plan" overflows out of the box on mobile. 
![image](https://github.com/saving-satoshi/saving-satoshi/assets/32662508/ba1e249f-82bd-46f2-8f62-993d7e605f6b)

ecurrencyhodler:
Is the solution here to change the font? @GBKS

benalleng:
I think a simpler solution would be to add some dynamic text size on mobile, just making it a little smaller should do the trick

GBKS:
Making the text smaller only works if the user has not chosen to increase the font size on their device via accessibility settings, or in their browser tab (small `aA` button in the bottom left on iOS Safari). A more flexible solution is to restrict it to one line of text using [text-overflow: ellipsis](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow).

benalleng:
Unfortunately It looks like our solution doesn't work site-wide. lets re-open

PR #538, opened 2023-06-27
Navbar text overflow

vinayaksh42:
Fixes: #524

### DIFF
--- a/components/Navbar/Address.tsx
+++ b/components/Navbar/Address.tsx
@@ -55,7 +55,7 @@ export default function Address() {
                 )}
               </p>
             )}
-            <p className="px-0.5 text-2xl leading-none text-white">
+            <p className="line-clamp-1 overflow-hidden text-ellipsis px-0.5 text-2xl leading-none text-white">
               {t(
                 isChapterCompletePage
                   ? 'navbar.chapter_complete'


ecurrencyhodler:
Tested on Firefox and chrome for iOS. It's a tACK for me. 

image
image

benalleng (approved):
LGTM!

benalleng:
Looks like we were a bit too hasty, this still seems to be an issue with that challenges themselves, and I am having trouble getting it to work locally.

vinayaksh42:
> Looks like we were a bit too hasty, this still seems to be an issue with that challenges themselves, and I am having trouble getting it to work locally.

Can you share screenshot of it as well as tell about the scenario in which it is not working.

benalleng:
Challenges don't seem to be working and locally it doesn't work at all
### on dev.savingsatoshi
image

### locally
image

benalleng:
Some of the problem might have to do with the complexity of how the Chapter#, Challenge# is rendered

vinayaksh42:
> Challenges don't seem to be working and locally it doesn't work at all
> 
> ### on dev.savingsatoshi
> image
> 
> ### locally
> image

Try running these commands 
```
git fetch upstream dev
git rebase upstream/dev
```

Your local branch is most probably not in sync with the dev branch as the changes are working perfectly for me locally as well as in production.

The issue with challenges-navbar in mobile view is more related to how we handle it from a design perspective. The aim of this PR was not to fix it as it was not mentioned in the scope of the issue #524 

Thanks for reporting it @benalleng, @GBKS do you have any suggestions on how we can handle the overflowing text in challenges-navbar?

ecurrencyhodler:
Yea I'll create a separate issue. The navbar for challenges will require design to give feedback on it.

PR #537, opened 2023-06-27
Add new ProfilesContainer component for HashrateChallenge variants

benalleng:
prelminary work for #503 

creates new Scrolling Profile container and 2 new lesson templates to demonstrate the functionality, does not add the complete HashrateChallenge

[Solo Preview](https://saving-satoshi-git-fork-benalleng-issue503-savingsatoshi.vercel.app/en/chapters/chapter-3/solo-1?dev=true)

[Pool Preview](https://saving-satoshi-git-fork-benalleng-issue503-savingsatoshi.vercel.app/en/chapters/chapter-3/pool-1?dev=true)

PR #542, opened 2023-06-30
Add HashrateChallenge

### LINKED ISSUE #503: HashrateChallenge component
Chapter 3 primarily consists of hashrate battles between 2 or more parties. To simplify the creation of the different challenges for Chapter 3 we should create a reusable `<HashrateChallenge>` component. This is what it would look like when rendered:

#### 1 vs. 1
image

#### 4 vs. 1 (Idle)
image

#### 4 vs. 1 (In progress)
image

#### 4 vs. 1 (Completed)
image

As you can see from the designs the component should:
- Support a dynamic amount of participants.
- Have different states (Idle, In progress, Completed).

When building this component we should reuse the components we created in:
https://github.com/saving-satoshi/saving-satoshi/issues/444
https://github.com/saving-satoshi/saving-satoshi/issues/448
https://github.com/saving-satoshi/saving-satoshi/issues/446

benalleng:
Builds off the Profile component in #537 

closes #503

Adds the hashrateChallenge and begins to add steps of the challenge

[Preview](https://saving-satoshi-git-fork-benalleng-finalize-6f5b4c-savingsatoshi.vercel.app/en/chapters/chapter-3/solo-1?dev=true)

To note, I am unsure how to diplay your teammates on ththe contribution bar without having to give them each 1 block each. this may be an enhancement we make to the contribution bar.


ecurrencyhodler:
Hey Ben. This should unblocked now with this commit by Kevin: https://github.com/saving-satoshi/saving-satoshi/commit/2036a08f8cb3d97d6a3546795a62aba2426464d4

ecurrencyhodler:
Just kidding. it's going to be addressed in https://github.com/saving-satoshi/saving-satoshi/pull/548

ecurrencyhodler:
Okay it's been merged! G2G Ben.

benalleng:
Going to go through today and see how quickly I can implement it!

ui/common/StartButton.tsx
```diff
@@ -91,20 +97,27 @@ export default function Card({
...
-            'border-black/25 text-black/25': powerUp === false,
-            ' animate-duration-500 animate-pulse border-[#FBEBC6] bg-[#FBEBC6] text-[#571A1F] shadow-[0_0px_15px_rgba(251,235,198,0.75)]':
-              powerUp === true,
-          }
-        )}
-        onClick={handleClick}
-      >
-        10x
-      </div>
+    <div
```
GBKS, line 100: Let's use the `<button>` for buttons (or anchor if it's for navigation). Better for accessibility and you get button-like behavior for free (just need to remove the default styling).

GBKS:
Cool to see this taking shape.

content/lessons/chapter-3/pool-1.tsx
```diff
@@ -1,105 +1,186 @@
...
+                  {
+                    'text-opacity-25': step === 0,
+                  }
+                )}
+              >
+                {i === 2 || i === 4
+                  ? i === 2
+                    ? PROTAGONISTS[2].value
+                    : ANTAGONISTS[0].value
+                  : 0}
```
afterburn, line 179: This is a bit confusing, I assume you have to do this index checking because `<HashrateChallenge>` takes a profiles prop, I think you can maybe get around having to do these conditionals if instead you make it take a protagonists and antagonists array instead.

content/lessons/chapter-3/pool-1.tsx
```diff
@@ -1,105 +1,186 @@
...
 }
 
-const TOTAL_BLOCKS = 100
+export default function Pool1({ lang }) {
+  const t = useTranslations(lang)
+  const [transactionsConfirmed, setTransactionsConfirmed] = useState(0)
+  const [bitcoinMined, setBitcoinMined] = useState(0)
+  const [step, setStep] = useState(0)
+  const [blocks1, setBlocks1] = useState(0)
+  const [blocks2, setBlocks2] = useState(0)
```
afterburn, line 24: Let's rename theset to `protagonistsBlockAmount` and `antagonistsBlockAmount` or something similar, blocks1 and blocks2 is not very descriptive

content/lessons/chapter-3/pool-1.tsx
```diff
@@ -110,24 +191,74 @@ export default function POL1({ lang }) {
...
+                    'text-white/25': step === 0,
+                    'fade-in text-white': step !== 0,
+                  })}
+                >
+                  {step === 0
+                    ? 0
+                    : (
+                        profile.hashpower /
+                        10 ** (profile.hashpower.toString().length - 2)
+                      ).toFixed(2)}
```
afterburn, line 214: Not sure what is happening here, why are we converting hashpower to a string, then taking the length of the string and subtracing 2. Whenever we're doing magic like this (which could very well be necessary) its good to leave some comments so others know what the magic is doing :smile:

afterburn (changes requested):
Thanks for the work @benalleng, left a couple comments. I also feel there is quite a bit of logic that currently resides in `solo-1.tsx` and  `pool-1.tsx` to the point where I'm wondering whether it would make more sense to move as much to the HashrateChallenge component itself (if possible).

I also think the `ContributionBar` can use some cleaning up, but I'll address that in a separate PR.

ecurrencyhodler:
So I did the challenges. It looks like the sizing of the design changes when the text gets displayed which feels a bit jarring. Also the screen with the extra text doesn't fit within the standard window. The design fix here I think is that the width of the text box should wider. 

This is true for solo-1 and pool-1. I don't think we have challenge 3 yet.

https://github.com/saving-satoshi/saving-satoshi/assets/32662508/fc95e0b2-a1d0-4464-a953-82b810cbaa5a

If the components work functionally, I saw let's merge and fix the design issues in the next sprint.

benalleng:
I moved the logic from the content pages into the shared <hashRateChallenge /> component additionally I went through to try and make the component generally easier to understand. In some places it might still be a doozy but i have added a few comments where I feel like it is absoulely neccessary.  Lastly for some of the design fixes I think we should hold off and fix those in a future PR, with @ecurrencyhodler's comment I think that can be fixed with some more generalized widths for the text when in that format and for @GBKS I would need to see exactly which of each of those elements have mistakes as I import quite a few components to build this PR.

Issue #506, opened 2023-06-11
Chapter 3.3 - Pool vs Bitrey (Win)


afterburn:
Using the `HashrateChallenge` component we created in https://github.com/saving-satoshi/saving-satoshi/issues/503 we should create the Pool vs. Bitrey challenge, in this challenge multiple users will form a pool and go up against Bitrey, ultimately the pool will win due to their combined hashrate + coordination outperforming Bitrey's hashrate.

This challenge consists of:
1. A 'matchmaking' page, here 4 people can form a group to go up against bitrey. These should be NPCs for the time being (in the future these might be real people). [Link to Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=4102-48879&t=VizOk0U0qeP1YnUA-4)
2. A page that uses the `HashrateChallenge` component to show the actual mining process, NPCs from previous page should persist to this page. [Link to Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=4117-54423&t=VizOk0U0qeP1YnUA-4)

Issue #505, opened 2023-06-11
Chapter 3.2 - Coop vs Bitrey (Lose)


afterburn:
Using the `HashrateChallenge` component we created in https://github.com/saving-satoshi/saving-satoshi/issues/503 we should create the Coop vs. Bitrey challenge, in this challenge multiple users will go up against Bitrey but there will be no coordination between the users which results in Bitrey winning.

This challenge consists of:
1. A 'matchmaking' page, here 4 people can form a group to go up against bitrey. These should be NPCs for the time being (in the future these might be real people). [Link to Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=4102-48845&t=LZXGkrObhw4Swxjv-4)
2. A page that uses the `HashrateChallenge` component to show the actual mining process, NPCs from previous page should persist to this page. [Link to Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=4102-48618&t=eCkKSeUA1EDezOwi-4)

Issue #504, opened 2023-06-11
Chapter 3.1 - Solo vs Bitrey lesson (Lose)


afterburn:
Using the `HashrateChallenge` component we created in https://github.com/saving-satoshi/saving-satoshi/issues/503 implement the Co-op vs. Bitrey (Lose) challenge. In this challenge the player and cooperators will go up against Bitrey. They lose this round because they didn't split the nonces when working together.

[Link to Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=4102-52875&t=LZXGkrObhw4Swxjv-4)

ecurrencyhodler:
Updated the issue title and description to match story line in figma.

PR #550, opened 2023-07-12
Contribution bar improvements

afterburn:
Follow-up PR for #534

image

Further improvements I think we can make in a separate PR:
- Render the avatars on a different place in the markup so that the `overflow-hidden` doesn't cut them off when the block amount is low.

### DIFF
--- a/content/lessons/chapter-3/pool-1.tsx
+++ b/content/lessons/chapter-3/pool-1.tsx
@@ -2,7 +2,7 @@
 
 import { useTranslations } from 'hooks'
 import { ProfileWithHashPower } from 'types'
-import { Card, HashFrequency, Text } from 'ui'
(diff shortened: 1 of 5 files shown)


ecurrencyhodler:
Tested on desktop and mobile on chrome.

There seems to be some issues with dynamic spacing on text. But I think that is beyond the scope of the PR. The contribution bar seems to well sized though.
Screenshot 2023-07-12 at 1 00 29 PM

For mobile, the sizing seems off as well but the contribution bar seems to work well. 
IMG_2916

afterburn:
@ecurrencyhodler yeah those issues are not really related to the `ContributionBar` itself like you'd assumed, they will be addressed when we start working on #504, #505, and #506.

ecurrencyhodler (approved):
tACK

PR #555, opened 2023-07-17
Close websocket after passing the scripting challenge

### LINKED ISSUE #554: After successfully passing CH2SCR2, the websocket stays open
It is possible to get into a neverending state of success messages using the following script:

```ts

let found = false
let nonce = 0

while(!found) {
    const hash = crypto.createHash('sha256').update(nonce.toString()).digest('hex')
    if (hash.startsWith('00000')) {
        console.log(hash)
    }
    nonce++
}
```
Notice `found` never gets set to `true`.

This in turn causes the challenge to time out after which the user fails the challenge.

To solve this, make sure the websocket closes after a successful completion has been detected.

afterburn:
Resolves #554


benalleng (approved):
tack! no longer gets trapped in a loop until timeout! challenge stays completed

PR #570, opened 2023-07-25
Contribution bar improvements

afterburn:


afterburn:
Found a small bug where you see the progress bar jump to 0% state on initial load, marking as ready for review after resolving that issue.

ecurrencyhodler (commented):
Okay. Since this is more design focused I'd like @GBKS to take a look as well. 

I will say the animations are definitely smoother which was nice to see. 

Question: In this section, it looks like only the blue miner is doing work and progressing. Is that intentional? Should we try to split it up a bit more?

https://github.com/saving-satoshi/saving-satoshi/assets/32662508/023d45d4-ca7e-47fd-9bd3-4a7a1a80fb1b

afterburn:
Regarding your question @ecurrencyhodler I also noticed that issue but that's not within the scope of this PR, that would belong to #505, which unfortunately has been closed. So I would recommend opening a new issue.

ecurrencyhodler:
Understood. Thanks for clarifying. I can create an issue for this thanks!

Looking at #579 we don't run into the same issue so looks like it's just for the co-op vs. bitrey (lose) challenge.

afterburn:
Yeah its a bug in the Coop lesson, updated my prev comment right as you replied.

ecurrencyhodler:
In that case let's just merge this. Great work kevin.

benalleng (approved):
LGTM! looks nice and smooth! We can address #581 and remove the fixed blocks in another PR.

PR #571, opened 2023-07-25
change challenge order

### LINKED ISSUE #572: Fix chapter 3 lesson ordering
After intro-2 im being sent to coop-1, this should be solo-1 instead

benalleng:
Fixes the challenge order for chapter 3

Resolves #572


ecurrencyhodler:
tACK! Order is now correct. Challenge 3 is a repeat of challenge 2 but I I believe it's just a place older until someone works on ch. 3.

afterburn (approved):

PR #578, opened 2023-07-25
Constrained monaco editor

### LINKED ISSUE #577: Investigate monaco plugin for disabling editing for certain parts of user code
https://github.com/Pranomvignesh/constrained-editor-plugin

afterburn:
Resolves #577, made it so that we can set constraints on a per-lesson basis (see scripting-2.tsx).

### DIFF
--- a/content/lessons/chapter-2/scripting-2.tsx
+++ b/content/lessons/chapter-2/scripting-2.tsx
@@ -22,6 +22,8 @@ const javascript = {
 
 // Create a program that finds a sha256 hash starting with 5 zeroes.
 // To submit your answer, log it to the terminal using console.log().
+
+// Type your code here
 `,
   validate: async (answer) => {
     if (!answer.startsWith('00000')) {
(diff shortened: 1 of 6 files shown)


vinayaksh42 (approved):
LGTM!
Tried out the [/scripting-2](https://saving-satoshi-git-feat-constrained-editor-savingsatoshi.vercel.app/en/chapters/chapter-2/scripting-2?dev=true) challenge. Unable to remove the code in both JS and Python.

Just one minor thing I noticed is that if the user has written down a certain amount of code and then decided to ctrl+A (select all) and delete, it won't let the user delete the code as the constrained part is also selected but I don't think this is a blocker.

ecurrencyhodler:
Tested this and I am unable to delete text in terminal for Ch.1 Challenge 2. 

I also tested Ch. 2 Scripting and was unable to delete the instructions which is great! 

>Just one minor thing I noticed is that if the user has written down a certain amount of code and then decided to ctrl+A (select all) and delete, it won't let the user delete the code as the constrained part is also selected but I don't think this is a blocker.

Yeah that's because it includes the text that has been restricted. If they highlight the code they wrote,  the text can be edited/deleted.

Merging this. Great find Vinayak and thanks kevin for researching and submitting this PR.

PR #579, opened 2023-07-25
Chapter 3 Challenge 3 (Pool-Win)

### LINKED ISSUE #506: Chapter 3.3 - Pool vs Bitrey (Win)

benalleng:
Closes #506 

[Preview](https://saving-satoshi-git-fork-benalleng-issue506-savingsatoshi.vercel.app/en/chapters/chapter-3/pool-1?dev=true)

### TODO
- get the correct avatar images imported

### DIFF
--- a/content/lessons/chapter-3/coop-2.tsx
+++ b/content/lessons/chapter-3/coop-2.tsx
@@ -102,6 +102,7 @@ export default function Coop2({ lang }) {
     setAntagonistsBlockAmount(newBlock)
   }
 
+  //This code below is for a potential fix to the miners having 1 blocks when they should have 0
   /*function findIndexOfHighestHashpower(players, key) {
     let highestScore = -Infinity;
     let indexOfHighestScore = 0;
--- a/content/lessons/chapter-3/coop-2.tsx
+++ b/content/lessons/chapter-3/coop-2.tsx
@@ -12,7 +12,6 @@ import { useAuthContext } from 'providers/AuthProvider'
 
 export const metadata = {
   title: 'chapter_three.coop_vs_bitrey_loose.title',
-  image: '/assets/images/chapter-2-intro-1.jpg',
   theme: 'solo-1-theme',
   secondaryTheme: 'solo-1-secondary-theme',
   key: 'CH3COO2',
--- a/content/lessons/chapter-3/coop-2.tsx
+++ b/content/lessons/chapter-3/coop-2.tsx
@@ -9,6 +9,7 @@ import { sleep } from 'utils'
 import Profile from 'ui/common/Profile'
 import clsx from 'clsx'
 import { useAuthContext } from 'providers/AuthProvider'
+import { cssVarThemeChange } from 'lib/themeSelector'
 
 export const metadata = {
   title: 'chapter_three.coop_two.title',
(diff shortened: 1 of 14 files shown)


ecurrencyhodler:
Beautifully done. I imagine with #570 this will looks smooth as butter. tACK.

ecurrencyhodler:
Ben: "This is ready for review! I added one additional enhancement to the contributionBar as well, the avatars now fade in once the step !== 0 so there are no avatars visible on first page load. All the copy for pool-1 and pool-2 should be correct but I did to some typo adjustments so a proofread is in order. Lastly I made the pool-1 contributor box quite differently than the figma in terms of its html / css so a design review would be handy there"

Leaving this comment here for now. Will follow up later.

content/lessons/chapter-3/coop-2.tsx
```diff
@@ -35,19 +35,19 @@ export default function Coop2({ lang }) {
   const PROTAGONISTS = [
     {
       username: 'You',
-      avatar: '/assets/avatars/1.png',
+      avatar: account?.avatar,
```
afterburn, line 38: Do we need a fallback here in case avatar is not available, or is that handled elsewhere?

content/lessons/chapter-3/pool-2.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export default function Pool2({ lang }) {
+  const saveAndProceed = useSaveAndProceed()
+  const t = useTranslations(lang)
+
+  useEffect(() => {
+    const rootStyles = document.documentElement.style
+    rootStyles.setProperty('--CH3SOL1-bg', '#411e4f')
+    rootStyles.setProperty('--CH3SOL1-gradient-start', '#3C1843')
+    rootStyles.setProperty('--CH3SOL1-gradient-stop', '#45235a')
```
afterburn, line 22: We seem to be doing this a lot, perhaps we can create a util function for it?

afterburn (changes requested):
Thanks @benalleng, see comments above.

ecurrencyhodler:
Wow it's so smooothhhhh.

@afterburn I think this is what you're referring to, but it defaults to a silhouette. I don't think we have to worry about this edgecase because a profile is required to get to this stage in the game.
Screenshot 2023-07-27 at 1 48 52 PM


@benalleng One copy correction:
"Splitting the extra nonce into two parts hass several advantages:"

Hass has two s's. Should be one. 

In general I think we can tighten up the copy but we can address that later.

content/lessons/chapter-3/coop-2.tsx
benalleng: the fallback for these are handled in the Profile and contributionBar components respectively

PR #582, opened 2023-07-27
Removed Ch. 1 error message in Transacting Challenge

### LINKED ISSUE #574: Remove Ch. 1 error message in Transacting Challenge
During a user interview, we discovered that the user found this message to be confusing when they tried to set the "var". This sent him down the wrong path to solving the challenge. We should stick to the standard error message "Sorry that's quite not right". 

The "Almost! Remember..." message is only helpful if they actually are close to writing out the full command, not when they're just setting the var.
Screenshot 2023-07-25 at 8 14 31 AM

harshita130602:
Fixes #574


ecurrencyhodler:
Hello @harshita130602! Thank you so much for this PR. 🙏

I tested it and there is no longer the "almost" message so it's a tACK for me! Let's get a code review on this and then we can merge. 
test

afterburn (approved):

PR #586, opened 2023-07-31
Remove blocks from other miners in CH3COO2

### LINKED ISSUE #585: Remove other miner's blocks in Ch3 co-op vs bitrey (lose)
In CH3COO2 only one protagonist miner is supposed to dinner blocks but it would still be nice to show a small sliver of progress for the other miners so their avatars are not all overlapping.

One approach is to give them blocks in the hashrate component but hide their numbers when rendering the page to the user

benalleng:
Closes #585 

in CH3COO2 only one miner is supposed to find blocks based on the nonce overlap. I created a semi-hacky way of leaving the other miners to be visible while still giving the correct number fo blocks to the protagonists / antagonists.

### DIFF
--- a/content/lessons/chapter-3/coop-2.tsx
+++ b/content/lessons/chapter-3/coop-2.tsx
@@ -80,6 +80,11 @@ export default function Coop2({ lang }) {
 
   const PROFILES: ProfileWithHashPower[] = [...PROTAGONISTS, ...ANTAGONISTS]
 
+  const highestHashpowerProtagonist = findIndexOfHighestHashpower(
+    PROTAGONISTS,
+    'hashpower'
+  )
+
   const handleStepUpdate = async (newStep: number) => {
     setShowText(false)
     await sleep(325)
@@ -99,6 +104,18 @@ export default function Coop2({ lang }) {
 
   const handleAntagonsitBlockUpdate = (newBlock: number) => {
     setAntagonistsBlockAmount(newBlock)
+  }
+
+  function findIndexOfHighestHashpower(players, key) {
+    let highestScore = -Infinity
+    let indexOfHighestScore = 0
+    for (let i = 0; i < players.length; i++) {
+      if (players[i][key] > highestScore) {
+        highestScore = players[i][key]
+        indexOfHighestScore = i
+      }
+    }
+    return indexOfHighestScore
   }
 
   useEffect(() => {
@@ -175,7 +192,11 @@ export default function Coop2({ lang }) {
                   }
                 )}
               >
-                {PROFILES[i].value}
+                {i === PROFILES.length - 1
+                  ? profile.value
+                  : i === highestHashpowerProtagonist
+                  ? profile.value + 3
+                  : '0'}
               </span>
             </Card>
             <Card className="flex gap-4">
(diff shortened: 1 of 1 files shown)


content/lessons/chapter-3/coop-2.tsx
```diff
@@ -101,6 +106,18 @@ export default function Coop2({ lang }) {
...
+    let indexOfHighestScore = 0
+    for (let i = 0; i < players.length; i++) {
+      if (players[i][key] > highestScore) {
+        highestScore = players[i][key]
+        indexOfHighestScore = i
+      }
+    }
+    return indexOfHighestScore
+  }
+
```
afterburn, line 120: If you are passing all the data this function needs via arguments we can move this function outside of the hook.
Also I think you can use a `.reduce` for this:
```ts
const indexOfHighestScore = players.reduce((highestIndex, current, index) => 
  current.value > players[highestIndex].value ? index : highestIndex,
0)
```

content/lessons/chapter-3/coop-2.tsx
```diff
@@ -175,7 +192,11 @@ export default function Coop2({ lang }) {
                   }
                 )}
               >
-                {PROFILES[i].value}
+                {i === PROFILES.length - 1
+                  ? profile.value
+                  : i === highestHashpowerProtagonist
+                  ? profile.value + 3
+                  : '0'}
```
afterburn, line 199: Not sure what is happening here, but I would recommend you do something like this instead:

```tsx
{i === PROFILES.length - 1 && profile.value}
{i === highestHashpowerProtagonist && profile.value + 3}
{i !== PROFILES.length - 1 && i !== highestHashpowerProtagnist && '0'}
```

While this probably takes more compute, I think its neglible if it makes it easier to parse than nested ternary operators.

afterburn (changes requested):
Thanks @benalleng, I requested some changes

ecurrencyhodler:
CH3COO2 from a design and usability stand points works for me.

GBKS:
Not sure if related to this PR, but there seems to be an issue with how the values are displayed. See below, where the miner on the right has a higher hash rate, but ends up with a lower amount of hashes checked.

image

From watching the numbers a few times, it seems like the exponent is not increased at the right time. 

[Preview link I used for testing](https://saving-satoshi-git-fork-benalleng-issue585-savingsatoshi.vercel.app/en/chapters/chapter-3/coop-2?dev=true)

I'm also curious why the numbers end up differently every time I run it? Don't we have some fixed math in place? Or are the some random factors in there?

afterburn (approved):

afterburn:
@ecurrencyhodler code looks good, feel free to merge after issue flagged by @GBKS has been resolved or if we decide to address it in a separate issue

GBKS:
This should be a separate issue as it's not directly related to the code touched here. The problem is in [line 246 of coop-2.tsx](https://github.com/saving-satoshi/saving-satoshi/blob/8664cb03702b0db7ab198c44570d82aedf2d9217/content/lessons/chapter-3/coop-2.tsx#L246). For some reason, the current nonce is multiplied with the hash power when calculating the exponent. It should be just the nonce. There is a comment next to that code around wrapping this functionality into a separate component or helper, which seems like a good idea.

PR #592, opened 2023-08-07
Scientific notation number overflow

### LINKED ISSUE #591: Exponent incorrectly tracks the number in <HashrateChallenge />
The exponent does not properly track the number as it counts up during the mining step of the challenge.

It changes arbitrarily compared to the counting numbers

Originally mentioned here https://github.com/saving-satoshi/saving-satoshi/pull/586#issuecomment-1665778337

### LINKED ISSUE #594: Ch. 3 Just nonce should be multiplied
The problem is in [line 246 of coop-2.tsx](https://github.com/saving-satoshi/saving-satoshi/blob/8664cb03702b0db7ab198c44570d82aedf2d9217/content/lessons/chapter-3/coop-2.tsx#L246). For some reason, the current nonce is multiplied with the hash power when calculating the exponent. It should be just the nonce. There is a comment next to that code around wrapping this functionality into a separate component or helper, which seems like a good idea.

258460987-e22a52ce-db18-4623-9452-1f7f13ffe4a8
From watching the numbers a few times, it seems like the exponent is not increased at the right time.

_Originally posted by @GBKS in https://github.com/saving-satoshi/saving-satoshi/issues/586#issuecomment-1668511258_

benalleng:
Closes #591, Closes #594

numbers should now move up a exponent only when they have actually increased the number of decimals

[Preview](https://saving-satoshi-git-fork-benalleng-issue591-savingsatoshi.vercel.app/en/chapters/chapter-3/solo-1?dev=true)


content/lessons/chapter-3/coop-2.tsx
```diff
@@ -236,15 +236,14 @@ export default function Coop2({ lang }) {
                   {step === 0
                     ? 0
                     : (
-                        profile.nonce! /
-                        10 ** (profile.nonce!.toString().length - 2)
+                        profile.hashes! /
+                        10 ** (profile.hashes!.toString().length - 2)
```
afterburn, line 240: Rather than telling the typescript compiler were absolutely certain `profile.hashes` exists, I think it makes sense to add a type or interface instead that represents the protagonists/antagonists. 

Something like:
```ts
interface Player {
      username: string;
      avatar: string;
      hashpower: number;
      hashes: number;
      color: string;
      value: number;
}
```

afterburn (changes requested):
Thanks @benalleng, requested a small change.

benalleng:
drafting as I'm going to go ahead and make this into a helper function as we mentioned in the code comments

benalleng:
@ecurrencyhodler No they are not hardcoded, only a range within what we expect them to fall, I made a decision to add some variance early on. Ultimately for this particular issue I think changing their hashPower values would fix it as there is only <5% difference between them in hashPower. Happy to have the randomness vs hard-coding discussion with everyone as I think @GBKS also asked about it in another comment.

ecurrencyhodler:
I actually kinda like that it's variable now that i think about it. Makes it feel more like "mining".

ecurrencyhodler:
Okay let's not treat my comment as a blocker. I think we are good to merge once we get an okay from kevin.

afterburn (approved):

PR #601, opened 2023-08-15
Bypass feature flags using dev query param

### LINKED ISSUE #600: Use dev query parameter to bypass feature flag system
In order to be able to test unreleased lesson content in our vercel previews we need to be able to bypass our feature flag system since we currently only have one environment/database. Alternatively we could spin up a backend for our `dev` environment but this is much more work, though we should do this eventually.

afterburn:
This PR resolves #600 so we can use `?dev=true` to bypass feature flags on the `/en/chapters` page, and it makes sure that the `?dev=true` query param will be passed along to lessons.

### DIFF
--- a/hooks/useSaveAndProceed.ts
+++ b/hooks/useSaveAndProceed.ts
@@ -10,22 +10,25 @@ import { useProgressContext } from 'providers/ProgressProvider'
 import { usePathData, useLocalizedRoutes } from 'hooks'
 import { lessons } from 'content'
 import { useAuthContext } from 'providers/AuthProvider'
+import useEnvironment from './useEnvironment'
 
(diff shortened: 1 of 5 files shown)


ecurrencyhodler:
tACK! I'm able to enter into Ch. 3.

hooks/useSaveAndProceed.ts
```diff
@@ -34,6 +37,7 @@ export default function useSaveAndProceed() {
         await saveProgressLocal(nextLessonKey)
       }
     }
+    console.log(route)
```
benalleng, line 40: we should probably delete this, or at least hide it ony when in dev

benalleng (approved):
LGTM! works as expected.

Issue #590, opened 2023-08-02
Fix width of miners displayed in Ch. 3 Challenge 2 and 3.


ecurrencyhodler:
Right now, I have to scroll to see all the miners in Ch. 3 challenges 2 and 3. This is the default state. We should enable scrolling but only if the screen width is significantly decreased. 

Here is a screen recording:

https://github.com/saving-satoshi/saving-satoshi/assets/32662508/e658bc73-55e3-42e5-9b90-93d36efa67b5

Here is the design in figma:
<img width="449" alt="Screenshot 2023-08-01 at 7 37 01 PM" src="https://github.com/saving-satoshi/saving-satoshi/assets/32662508/d1a5b661-9402-4b5b-abaa-c7cf64a143de">

ecurrencyhodler:
Closed by https://github.com/saving-satoshi/saving-satoshi/pull/604

PR #604, opened 2023-08-18
Fix width of miners displayed in Ch. 3 Challenge 2 and 3.

0tuedon:
This PR should solve the issue of scrolling on 1290px screens and above, i had to reduce the gap and size of the containers on certain breakpoints  #590

### DIFF
--- a/content/lessons/chapter-3/coop-2.tsx
+++ b/content/lessons/chapter-3/coop-2.tsx
@@ -166,7 +166,7 @@ export default function Coop2({ lang }) {
   }, [])
 
   return (
-    <div className="fade-in flex flex-col items-center  gap-[20px] px-[15px] py-[30px] md:pt-[40px] 2xl:gap-[30px] 2xl:px-[75px]">
+    <div className="fade-in flex flex-col items-center  gap-5 px-[15px] py-8 md:pt-10 2xl:gap-8 2xl:px-[75px]">
       <HashrateChallenge
         totalBlocks={TOTAL_BLOCKS}
         blockRatio={BLOCK_RATIO}
--- a/content/lessons/chapter-3/coop-2.tsx
+++ b/content/lessons/chapter-3/coop-2.tsx
@@ -166,7 +166,7 @@ export default function Coop2({ lang }) {
   }, [])
 
   return (
-    <div className="fade-in flex flex-col items-center gap-[20px] px-[15px] py-[30px] md:px-[75px] md:pt-[40px]">
+    <div className="fade-in flex flex-col items-center  gap-[20px] px-[15px] py-[30px] md:pt-[40px] 2xl:gap-[30px] 2xl:px-[75px]">
       <HashrateChallenge
         totalBlocks={TOTAL_BLOCKS}
         blockRatio={BLOCK_RATIO}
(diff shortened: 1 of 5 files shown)


ecurrencyhodler:
hihi! Is this ready for review? If so, would you mind putting this PR into open status?

0tuedon:
> hihi! Is this ready for review? If so, would you mind putting this PR into open status?

not yet i need to do some final check

0tuedon:
@ecurrencyhodler, this is ready for review . i would also need your feedback on the reduction of the gap on lesser screens

content/lessons/chapter-3/coop-2.tsx
```diff
@@ -166,7 +166,7 @@ export default function Coop2({ lang }) {
   }, [])
 
   return (
-    <div className="fade-in flex flex-col items-center gap-[20px] px-[15px] py-[30px] md:px-[75px] md:pt-[40px]">
+    <div className="fade-in flex flex-col items-center  gap-[20px] px-[15px] py-[30px] md:pt-[40px] 2xl:gap-[30px] 2xl:px-[75px]">
```
afterburn, line 169: Try to refrain from using arbitrary pixel values as much as possible, `pt-[40px]` could instead be `pt-10` instead, and unless absolutely necessary rather than doing `py-[30px]` use `py-8`.

afterburn (changes requested):
Thanks @0tuedon, requested a small change in the comment above.

ecurrencyhodler:
@0tuedon Okay so it looks like the white scrolling bar also disappeared. I think we need that when the width becomes narrow triggering scrolling to the right to see other hashers. Otherwise I won't know to scroll to the right to see other players. 

Here's a screen of what it looks like without the white bar. There is actually BitRey to the right but there is not visual cue to scroll to see them. 
Screenshot 2023-08-21 at 9 32 22 AM

Here's the size of the screen if that helps.
Screenshot 2023-08-21 at 9 32 31 AM

content/lessons/chapter-3/coop-2.tsx
0tuedon: oh okay but py-8 is 32px is that fine?

0tuedon:
> @0tuedon Okay so it looks like the white scrolling bar also disappeared. I think we need that when the width becomes narrow triggering scrolling to the right to see other hashers. Otherwise I won't know to scroll to the right to see other players.
> 
> Here's a screen of what it looks like without the white bar. There is actually BitRey to the right but there is not visual cue to scroll to see them. Screenshot 2023-08-21 at 9 32 22 AM
> 
> Here's the size of the screen if that helps. Screenshot 2023-08-21 at 9 32 31 AM

okay but are we fine with the width of each players box?

ecurrencyhodler:
Yeah. Looked good to me on full screen.

benalleng:
tack! looks to function as expected

ecurrencyhodler:
Hi @0tuedon. 

Can we turn the scroll bar white?
Screenshot 2023-08-24 at 8 21 19 AM

ecurrencyhodler:
Yes! ty.

afterburn (approved):
Code looks fine, @ecurrencyhodler feel free to merge

PR #606, opened 2023-08-21
Chapter 3 Challenge 4 (Splitting Rewards)

### LINKED ISSUE #507: Chapter 3.4 - Splitting rewards
After beating Bitrey users will learn about how pool rewards are being distributed amongst participants.

This challenge consists of:
1. An explainer page. [Link to Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=4102-48939&t=eCkKSeUA1EDezOwi-4)
2. Page using `HashrateChallenge` component, on the bottom there should be text explaining each step. [Link to Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=4228-69579&t=VizOk0U0qeP1YnUA-4)

benalleng:
Closes #507 

~~Still not implemented with the backend, currently uses a fixed JSON string to feed the data in.~~

```
  const poolThreeOutcome = JSON.parse(
    '{"protagonists": [15, 9, 29, 4], "antagonists": [43]}'
  )
```

~~The backend needs to be implemented but this is the basic format for the JSON in this lesson~~

Decided to go with using fixed rewards on `pool-3` and as a result `split-3` so there is not too much overhead on subsequent chapters

[Preview](https://saving-satoshi-git-fork-benalleng-issue507-savingsatoshi.vercel.app/en/chapters/chapter-3/pool-3?dev=true)


ecurrencyhodler:
Thanks Ben!

Can we make the text width wider? I think it unnecessarily makes the user scroll down when there's some text. 

https://github.com/saving-satoshi/saving-satoshi/assets/32662508/bbb2f2c2-b113-43d7-9b32-f4c67acf7328

ui/common/BoxDisplayer.tsx
```diff
@@ -12,7 +12,7 @@ export default function Card({
   return (
     <div
       className={clsx(
-        'rounded-[5px] px-2 py-1.5 font-space-mono text-sm',
+        'rounded-[5px] py-[4px] px-[7px] font-space-mono text-sm leading-none',
```
0tuedon, line 15: we could use some tailwind classes here  like py-1?

ui/introduction/Introduction.tsx
```diff
@@ -6,7 +6,15 @@ import { redirect } from 'next/navigation'
 import { TextImage } from 'ui'
 import { useLocalizedRoutes, usePathData } from 'hooks'
 
-export default function Introduction({ children, lang }) {
+export default function Introduction({
+  children,
+  lang,
+  imagePosition,
+}: {
+  children: any
```
0tuedon, line 14: we could use ReactNode as a children type instead of any?
also what types are lang?

0tuedon (changes requested):

0tuedon:
@benalleng  Nice work, i left some not so major comments.

ui/introduction/Introduction.tsx
benalleng: all langs are strings, but they're everywhere so we can address that in another PR, made the changes here though

ecurrencyhodler:
tACK!

Issue #597, opened 2023-08-08
Update copy in ch. 2 challenge 2


ecurrencyhodler:
During a user interview, the user faced some issues which leads to several questions:
1. He was stuck on the word "function" and wasn't sure if he had to write out a function to complete the challenge. Should we change that?
1. Hashes need to be printed to the console in order to complete the challenge but this isn't explicitly stated in the instructions. Is this something that we need to do?
1. Should we also provide the function code inside the terminal like Kevin showed us on the call today?

Here is the [full interview](https://drive.google.com/file/d/1eL7HxhHN0zN7bZFWMmaY_gwmvDXJHOoj/view) for reference.

PR #607, opened 2023-08-21
fix: update chapter 2 challenge 2 copy in English

pinheadmz:
closes https://github.com/saving-satoshi/saving-satoshi/issues/597

Well, at least the first two bullet points. I can't recall what this refers to:
> Should we also provide the function code inside the terminal like Kevin showed us on the call today?

### DIFF
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -311,20 +311,18 @@ const translations = {
     scripting_two: {
       title: 'Automation',
       paragraph_one:
-        'Alright, time to write and run your own code. Write a function that returns a sha256 hash. The code should calculate and return a hash based on the nonce argument.',
-      paragraph_two:
-        'When you press "Run the script", the function you wrote will be called over and over with an ever-increasing nonce (which is just a regular number). Write code to calculate and return a hash based on the nonce. We are trying to run this function until we find a hash that starts with five zeroes (“00000”).',
+        'Alright, time to write and run your own code. Write a script that generates a sha256 hash that begins with five zeroes ("00000..."). Your code should repeatedly call the sha256 function with different input until the output satisfies this requirement. You should try incrementing an integer inside a loop to get different inputs. In cryptography this number may referred to as a "nonce" or "number used once".',
       python: {
-        paragraph_three:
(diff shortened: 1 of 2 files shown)


i18n/locales/en.ts
```diff
@@ -311,20 +311,18 @@ const translations = {
...
+        'Alright, time to write and run your own code. Write a script that generates a sha256 hash that begins with five zeroes ("00000..."). Your code should repeatedly call the sha256 function with different input until the output satisfies this requirement. You should try incrementing an integer inside a loop to get different inputs. In cryptography this number may referred to as a "nonce" or "number used once".',
       python: {
-        paragraph_three:
-          'We are using the hashlib library in python to help you along in creating this function, you can use the following external resources to help you write this function if needed:',
+        paragraph_two:
+          "When you find a nonce with a hash that begins with five zeroes, use Python's `print()` function to return the hash to the console. We are using the hashlib library in python to help you along in creating this function, you can use the following external resources to help you write this function if needed:",
```
pinheadmz, line 317: hm I had written this with single quotes and an escaped `Python\'s` but i think a git hook linter in the project changed it?

ecurrencyhodler:
Hey Matthew. Thanks for submitting this PR. Let's discuss this on our call tomorrow.

afterburn (approved):
lgtm

i18n/locales/en.ts
afterburn: @pinheadmz We do have a pre commit hook that runs the `lint-staged` command you can find in `package.json`, which runs prettier to format this doc. But IIRC we don't have any rule that would do this. Could it be your own editor settings? 

In any case this should work: `"When you find a nonce with a hash that begins with five zeroes, use Python's 'print()' function to return the hash to the console. We are using the hashlib library in python to help you along in creating this function, you can use the following external resources to help you write this function if needed:"` as you can see in 7a18e5a160ecb1d4955a86adf1dbdb6681284108

benalleng:
I think it's fine as is but we do have the ability to add styles to text with an inline `<span>` to get some "code text" effects
image

PR #608, opened 2023-08-22
Change blocks found amount found for coop-2, made player the on to find blocks

### LINKED ISSUE #581: Change progress of characters in Ch. 3 Co-op vs Bitrey (Lose)
In this section, it looks like only the blue miner is doing work and progressing. We should split this up more evenly across contributors. Let's reference the % distribution in #579.

https://github.com/saving-satoshi/saving-satoshi/assets/32662508/023d45d4-ca7e-47fd-9bd3-4a7a1a80fb1b

_Originally posted by @ecurrencyhodler in https://github.com/saving-satoshi/saving-satoshi/pull/570#pullrequestreview-1546364979_

afterburn:
Resolves #581, set the blocks found amount to 0 for 3 co-op players, set the Player to be the one to find blocks.

### DIFF
--- a/content/lessons/chapter-3/coop-2.tsx
+++ b/content/lessons/chapter-3/coop-2.tsx
@@ -46,7 +46,7 @@ export default function Coop2({ lang }) {
       hashpower: 4395,
       hashes: protagonistHash[0],
       color: '#F3AB29',
-      value: step == 0 ? 0 : 1,
+      value: protagonistsBlockAmount,
     },
     {
       username: 'Mining Maniacs',
@@ -54,7 +54,7 @@ export default function Coop2({ lang }) {
       hashpower: 4054,
       hashes: protagonistHash[1],
       color: '#FE5329',
-      value: step == 0 ? 0 : 1,
+      value: 0,
     },
     {
       username: 'Hash Hoppers',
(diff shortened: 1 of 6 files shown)


afterburn:
@ecurrencyhodler I see, so basically what it was like before? https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=4102-48738&mode=design&t=IMDBZhgluqy77skG-4

If the intention is to be able to have the remaining 3 players have 0 blocks found, but still be visible on the ProgressBar, I can make this fix in the ProgressBar component itself. So when we pass it 0 blocks found, it will still render something for the player. Please confirm if this is the intended goal.

ecurrencyhodler:
Yeah the main thrust of the issue is to reorder the characters in Ch. 3 Co-op vs Bitrey in the progress bar. Hash hoppers should be the last ones on the right winning blocks. The other hashers are visible and are on the left with 0 blocks.

Thanks.

benalleng:
This is code from me, I added the other miners getting individual blocks and editing the numbers for the user so it appeared on the contribution bar but did not show them as having blocks, this was the purpose of the ternaries `value: step == 0 ? 0 : 1`.  When I made this I wasn't sure how the contribution bar functioned and I decided to go with the approach to edit the profiles instead.

If you want to adjust the `contributionBar` to work this way that would definitely work too and probably make more sense on the content page!

As for the intention of the issue, I think it was just to rearrange the `PROTAGONISTS[]` like @ecurrencyhodler mentioned

afterburn:
@ecurrencyhodler @benalleng ready for another review

ecurrencyhodler:
tACK! Great work kevin. Thanks.
Screenshot 2023-08-23 at 3 34 25 PM

ecurrencyhodler (approved):

PR #609, opened 2023-08-22
Fix og tags

### LINKED ISSUE #312: Preview image still only visible on /[lang] url
I noticed when posting a link to my friends to test the demo that the preview url is still only visible on the `/[lang]` url but looking at the app folder i could not see immediately why this was the case.
@vinayaksh42 any ideas? I know you helped in removing the `/[lang]` url on the homepage and hoped you might have some insight
nothing in `/app/head.tsx` mentions the /[lang] url

afterburn:
Resolves #312, changed the url to use a relative path rather than an absolute one, this should make it so that Next will resolve it for the request.


benalleng (changes requested):
Unfortunately, upon testing there doesn't seem to be a change.

I think i figured out why it isn't working.

upon inspecting the `<head>` of this deployment I noticed the description did not match the description from `app/head.tsx` but instead was coming from `app/page.tsx`.

image

 There, there is another og:image that I think is overriding the metadata here

```
openGraph: {
    title: 'Code your way through the mysteries of Bitcoin - Saving Satoshi',
    description:
      'A fun learning resource for developers to learn about how to build on bitcoin.',
    type: 'website',
    image:
      'https://imagedelivery.net/wyrwp3c-j0gDDUWgnE7lig/6818cd1b-ea0d-49a4-ea14-d34c1bbae100/public',
    url: 'https://savingsatoshi.com/',
  },
```

if we change the preview image here we might get to see some results

afterburn:
Thanks @benalleng, you're right, `page.tsx > metadata` will always override whatever we put in `app/head.tsx`. I now used a different approach that should allow us to very easily add opengraph/twitter tags to our pages by importing `defaultMetadata`, then engineers will have the ability to override those defaults, like I have done with `title` in some of the files.

afterburn:
Noticed this doesn't actually fix #312, although it does fix a metatag problem we had, so now moving on to fixing the actual bug.

afterburn:
@benalleng @ecurrencyhodler found an issue with our middleware, should be ready for review now.

Screenshot 2023-08-23 at 11 39 32

ecurrencyhodler:
Thanks Kevin! Tested in discord but the preview seems to be smaller than anticipated. This is what I see when I paste into discord: https://saving-satoshi-b6hjyv9m0-savingsatoshi.vercel.app/
Screenshot 2023-08-23 at 8 35 42 AM

I saw one version where @benalleng was able to generate one in discord with a bigger image:

Screenshot 2023-08-23 at 8 35 56 AM

 Do you think we could make the preview image larger like this one? No worries if not.
 
 The good news is that I checked using the linkedin url and twitter and both are able to generate a preview image:
Screenshot 2023-08-23 at 8 41 21 AM
Screenshot 2023-08-23 at 8 38 45 AM

config/metadata/twitter.ts
```diff
@@ -0,0 +1,10 @@
+const twitter = {
+  twitter: {
+    card: 'summary',
```
benalleng, line 3: just a quick change for @ecurrencyhodler 's suggestion

```suggestion
    card: 'summary_large_image',
```

benalleng (approved):
LGTM! We can just change the twitter card type to see the larger image

PR #611, opened 2023-08-22
Test openGraph metadata

benalleng:


benalleng:
current problem, a blank pathname, `/` results in this from metadata scrapers

image


this is with `/en`

image

Issue #587, opened 2023-08-01
Chapter 3 Challenge 2 [Co-op vs BitRey loose] styling related feedback (Christoph)


vinayaksh42:
- not use a fixed width at smaller sizes, but a max-width. That allows the boxes to shrink a little further if needed, like for iPhone SE. (for the waiting screen avatar boxes). 
- centering the copy on mobile
- the “Waiting…” text would slightly fade in and out to indicate that there’s activity. And then if the images faded in nicely.

0tuedon:
Hey @vinayaksh42  for the second issue i don't know what copy should me center aligned here (LGTM)
screenshot below
<img width="777" alt="Screenshot 2023-08-19 at 22 48 36" src="https://github.com/saving-satoshi/saving-satoshi/assets/90271995/58d58ebd-09f3-4b2e-a584-990ad258d54b">

benalleng:
I believe the issue there is mobile, I'm not sure though as it does seem inline with other pages.

![Screenshot_20230819-190017.png](https://github.com/saving-satoshi/saving-satoshi/assets/108441023/226a8bd6-0290-41a3-8cba-231668c5bbbb)

ecurrencyhodler:
Closed by #613

PR #613, opened 2023-08-23
Fix: max- width instead of width and fade in and out on waiting text and images

0tuedon:
This Pr references #587  and solves  on the styling  of chapter 3 challenge 2.

### DIFF
--- a/content/lessons/chapter-3/coop-1.tsx
+++ b/content/lessons/chapter-3/coop-1.tsx
@@ -75,11 +75,11 @@ export default function Coop1({ lang }) {
 
   return (
     <div className="grid grid-cols-1 justify-center justify-items-center md:my-auto md:flex md:flex-row">
-      <div className="fade-in my-[30px] grid grid-cols-2 gap-[15px] md:order-last md:mx-[30px] md:h-[400px] md:w-[410px] md:gap-[30px]">
+      <div className="fade-in my-[30px] grid w-full max-w-[300px] grid-cols-2 justify-center justify-items-center gap-[15px] md:order-last md:mx-[30px] md:min-h-[400px] md:min-w-[410px] md:gap-[30px]">
         {players.map((profile, i) => (
           <div
             key={i}
-            className="h-auto w-full max-w-[160px] border-2 border-dotted border-white/25 p-[15px] md:h-[185px] md:w-[190px] md:max-w-none"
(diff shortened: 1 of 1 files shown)


ecurrencyhodler:
Thanks for submitting this pr. Let's put a pause on this until @GBKS can clarify on a couple issues. This isn't a blocker for launch.

I can confirm the waiting fades in and out which is nice. I'm not sure how the centered text should look though.

GBKS:
Thanks for working on this. I still see some layout issues and left-aligned text. Screenshots are via the Responsive design mode on Safari on Desktop.

image

image

Generally, it is not a good idea to use fixed width or height for content containers (especially height). Text will render at different lengths across translations, people change text size, etc.

0tuedon:
> Thanks for working on this. I still see some layout issues and left-aligned text. Screenshots are via the Responsive design mode on Safari on Desktop.
> 
> image image
> Generally, it is not a good idea to use fixed width or height for content containers (especially height). Text will render at different lengths across translations, people change text size, etc.

hey @GBKS  before i commit my changes i do what to know if this  loading style works

https://github.com/saving-satoshi/saving-satoshi/assets/90271995/90704101-840d-411a-aea6-380566542467

0tuedon:
@GBKS  can i get your feedback on the new fix

GBKS:
Thanks for the tweaks. I still see layout issues on mobile.

image

Before any of the other 3 miners are loaded, the grid nicely fits into the screen. As soon as one of them is loaded, it expands to the screen edges (it should keep the first state with the padding). This is likely due to the boxes responding to the width of the labels. The layout logic should be reverse where the box defines the width, and the label resizes its width accordingly.

Again, it is not a good idea to use fixed heights. Using a min-height is a better idea, because it allows the container to resize if the content happens to be longer (localization to another language, for example, might cause this).

0tuedon:
> Thanks for the tweaks. I still see layout issues on mobile.
> 
> image
> 
> Before any of the other 3 miners are loaded, the grid nicely fits into the screen. As soon as one of them is loaded, it expands to the screen edges (it should keep the first state with the padding). This is likely due to the boxes responding to the width of the labels. The layout logic should be reverse where the box defines the width, and the label resizes its width accordingly.
> 
> Again, it is not a good idea to use fixed heights. Using a min-height is a better idea, because it allows the container to resize if the content happens to be longer (localization to another language, for example, might cause this).

Alright, i have been able to restraint from using fixed width, and i have been able to fix the layout issues mentioned, you could review

GBKS (approved):
Looking good, thank you for addressing my feedback.

PR #649, opened 2023-09-12
add .nvmrc and setup file

### LINKED ISSUE #642: Add .nvmrc file to repo
A request from Christoph is to add a ".nvmrc" file to the repo. This allows the repo to auto-switch to an appropriate node version when installing dependencies.

sabakhilji:
closes #642


ecurrencyhodler:
Thanks for submitting this pr Saba! Can you change the name of the PR to what you have in the PR description? add .nvmrc and setup shell file

Then in the PR description you can say that it closes issue 649.

Thanks!

sabakhilji:
> Thanks for submitting this pr Saba! Can you change the name of the PR to what you have in the PR description? add .nvmrc and setup shell file
> 
> Then in the PR description you can say that it closes issue 649.
> 
> Thanks!

Oh, I am sorry I was in a hurry :) I updated it now. Please check. Also, I need to update the readme file after someone reviewed this.

ecurrencyhodler:
Awesome thanks for updating the name Saba!

.nvmrc
```diff
@@ -0,0 +1 @@
+16.20.0
```
benalleng, line 1: Sorry! I was over zealous with the necessary version, the max needed is `16.13.0`

setup.sh
benalleng, line 1: Seems good to me, I am not completely sure this file is totally necessary for onboarding new devs but It looks good otherwise!

benalleng (changes requested):
Small change for the version needed, doesn't actually need to be that high but otherwise it looks good!

.nvmrc
sabakhilji: done !!

setup.sh
sabakhilji: So should we include this?

benalleng: can't hurt!

setup.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
benalleng, line 2: I dont think a full addition to the README is necessary ultimately maybe just a one line comment here saying 
`#run this file if you are unsure of your installed node-version`

benalleng (approved):
LGMT!

setup.sh
benalleng:
```suggestion

#Run this file if you are unsure of your installed node version
```

benalleng:
Hmm this seems to have caused our prod branches to break again..

PR #644, opened 2023-09-11
Change order of miners for rest of CH3

benalleng:
Closes #623

PR #653, opened 2023-09-15
Fix: Increased Max-width on chapter 3 challenge 4, also removed unused flex , flex-col

### LINKED ISSUE #635: Layout issues in CH 3 Challenge 4
On the left, you can see the text budging directly against the percentage boxes, and text sticking outside of its container. On the right, the first avatar is much smaller than the others causing inconsistencies.

The second issue is probably caused by applying a fixed with to the labels ("Hashrate", "Blocks found"...). Not a good approach, it's much better to relay on [flex layout](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) to space things correctly.

The user boxes have a max-width of 220, which should be 300 per Figma. That will also help fit in all the content and prevent layout issues.

Frame 1

I also see a bunch of unnecessary Tailwind classes again. For example, a text paragraph with `flex` and `flex-col`. Those two properties define how child elements are laid out, but text paragraphs don't have any. Doesn't break anything, just confusing and makes it harder to work with the code.

[Dev site link](https://dev.savingsatoshi.com/en/chapters/chapter-3/split-2?dev=true).

0tuedon:
This PR closes #635,
Fix overflow on chapter 3 challenge 4 and remove unused flex and flex-col on paragraph.

### DIFF
--- a/content/lessons/chapter-3/coop-2.tsx
+++ b/content/lessons/chapter-3/coop-2.tsx
@@ -174,6 +174,7 @@ export default function Coop2({ lang }) {
             username={profile.username}
             avatar={profile.avatar}
             description={profile.description}
+            className={'md:!max-w-[220px]'}
           >
             <Card className="flex">
               <span
--- a/content/lessons/chapter-3/coop-2.tsx
+++ b/content/lessons/chapter-3/coop-2.tsx
@@ -234,20 +234,20 @@ export default function Coop2({ lang }) {
         >
           {step === 0 && (
             <span className="flex flex-col items-start gap-[10px] md:w-[490px] md:min-w-[490px] md:pt-[20px]">
-              <Text className="flex flex-col self-stretch text-center font-nunito text-[24px] font-bold">
+              <Text className="self-stretch text-center font-nunito text-[24px] font-bold">
                 {t('chapter_three.coop_two.step_zero_heading')}
               </Text>
-              <Text className="flex flex-col self-stretch text-center font-nunito text-[18px] font-semibold">
+              <Text className="self-stretch text-center font-nunito text-[18px] font-semibold">
                 {t('chapter_three.coop_two.step_zero_paragraph_one')}
               </Text>
             </span>
           )}
           {step === 2 && (
             <span className="flex flex-col items-center gap-[10px] md:w-[490px] md:min-w-[490px] md:pt-[20px]">
-              <Text className="flex flex-col self-stretch text-center font-nunito text-[24px] font-bold">
+              <Text className=" self-stretch text-center font-nunito text-[24px] font-bold">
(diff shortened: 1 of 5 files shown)


0tuedon:
@afterburn  vercel is having some issue deploying

ecurrencyhodler:
Thanks for submitting this pr @0tuedon! Can you update the title of the pr to describe the action the PR is taking? Also I recommend using the language "closes #635" in the description as that will automatically close the issue once this pr is merged. 

Example title: Fix overflow on chapter 3 challenge 4 and remove unused flex and flex-col on paragraph.

0tuedon:
> Thanks for submitting this pr @0tuedon! Can you update the title of the pr to describe the action the PR is taking? Also I recommend using the language "closes #635" in the description as that will automatically close the issue once this pr is merged.
> 
> Example title: Fix overflow on chapter 3 challenge 4 and remove unused flex and flex-col on paragraph.

ACK, would change that.

ecurrencyhodler:
Hmmm I'm running into the save blocker as before. I can't move past ch.3 intro-1. :( 

@benalleng what worked for you?

benalleng:
We'll need to merge #644 to see those errors disappear

ecurrencyhodler:
@0tuedon can you rebase your pr? I'll try again after.

afterburn (approved):

0tuedon:
> Hey @0tuedon I accidentally started a thread in the issue, but here is feedback from christoph.:
> "Yes, looks better. There is a layout jump that happens when pressing the first "Tell me more" on certain screen sizes. Looks like up to that point, the width setting is different than after. It would be great to fix it so the layout is consistent and this jump does not occur."

Alright would check that out.

0tuedon:
@ecurrencyhodler, i did make an update to fix the skip issue. WDYT? @GBKS

ecurrencyhodler:
Thanks for fixing the skip issue! The values no longer overlap but the space seems really wide inside the box. 

Also tagging @GBKS to review.
wide

0tuedon:
> Thanks for fixing the skip issue! The values no longer overlap but the space seems really wide inside the box. 
> 
> Also tagging @GBKS to review.
> wide
> 

Yh @ecurrencyhodler, i think the previous issue was that it wasn't using the whole box and was just taking the max content,  hence the skip when the content is increased

GBKS:
Thanks for fixing the skip, it's gone now.

I noticed a couple more visual things to tweak. Totally cool to do those in a separate PR.

image

There are still some fixed widths in the code. Those will likely cause layout issues when text size changes with localizations.

afterburn (approved):

afterburn:
Created a new issue for the remaining visual tweaks in #660

PR #654, opened 2023-09-19
fix: Change Challenges copy and chapter 2 design

### LINKED ISSUE #619: Update Ch.3 Challenge 2 with HoloCat Pool. Also change names of Challenges 2 and 3.
**UPDATE**
Here's what we need for this issue:

- [ ] Change name of Challenge 2 to Pool vs. BitRey (including URL, challenge copy, and chapter titles)
- [ ] Change name of Challenge 3 to Coop vs. BitRey (including URL, challenge copy, and chapter titles)
- [ ] We also need to need to add a frame around the cooperators in Challenge 2. But we won't be adding badges yet. Bades will only come into play for Challenge 3.

So this is challenge 2 right now.
Screenshot 2023-09-04 at 8 28 27 AM

Challenge 2 will need the right side updated to the below image:
265672139-f24eed4b-d29d-49fb-aabd-0d622196e1f6

0tuedon:
The PR closes #619 .

### DIFF
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -387,19 +387,19 @@ const translations = {
   chapter_three: {
     title: 'The 51% Attack',
     paragraph_one:
-        'You enter the address of a block explorer blocks are back to ten minutes intervals. Somehow, holocat sleeps through the noise of the machines.',
-      paragraph_two:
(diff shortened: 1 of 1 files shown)


0tuedon:
> Is this ready for review?

No not yet

PR #656, opened 2023-09-19
Adjust overflow and dim contributionBar scrollbar

### LINKED ISSUE #633: Remove horizontal scroll bar in chapter 3 mining challenges
It is not important to show a scroll bar here at all. Horizontal scrolling is a really common interaction people are familiar with. As far as communicating to users that the area is scrollable, the standard is simply that some items are cut off and go beyond the screen edge.

image

Also the content on the right should go all the way to the edge and not be cut off 15px from the browser edge.

If there's insistence that we need a scrollbar, I can come up with a much more subtle one that this big white block.

benalleng:
Closes #633 

[Preview](https://saving-satoshi-git-fork-benalleng-issue633-savingsatoshi.vercel.app/en/chapters/chapter-3/coop-2?dev=true)


ecurrencyhodler:
It's missing the brighter scroll bar when the mouse hovers. But content does get cut off and goes all the way to the end on the right.
Screenshot 2023-09-19 at 4 38 23 PM

ecurrencyhodler (approved):

afterburn (approved):

PR #661, opened 2023-09-20
Chapter 3 scrollbar tweaks

GBKS:
This PR builds on #656 (and in turn #633).

The changes make the scroll bar vertically narrower (5px) to match the design, and adds the hover state.

I also adjusted the spacing of the container of scrollable content slightly for the same reason. 

The diff of the challenge pages look like a ton changed, but all I did was add a wrapper div to better control the layout. GitHub just thinks everything inside that wrapper has also changed. Silly GitHub.

### DIFF
--- a/content/lessons/chapter-3/coop-2.tsx
+++ b/content/lessons/chapter-3/coop-2.tsx
@@ -158,108 +158,110 @@ export default function Coop2({ lang }) {
   }, [])
 
   return (
-    <div className="fade-in flex flex-col items-center  gap-5 py-8 pl-[15px] md:pt-10 2xl:gap-8 2xl:px-[75px]">
-      <HashrateChallenge
-        totalBlocks={TOTAL_BLOCKS}
-        blockRatio={BLOCK_RATIO}
-        step={step}
-        onStepUpdate={handleStepUpdate}
-        onProtagonistUpdate={handleProtagonsitBlockUpdate}
-        onAntagonistUpdate={handleAntagonsitBlockUpdate}
-        protagonists={PROTAGONISTS}
-        antagonists={ANTAGONISTS}
-        profiles={PROFILES.map((profile, i) => (
-          <Profile
-            key={i}
-            username={profile.username}
-            avatar={profile.avatar}
-            description={profile.description}
-            className={'md:!max-w-[220px]'}
-          >
-            <Card className="flex">
-              <span
-                className={clsx('fade-in font-nunito text-[15px] font-bold', {
-                  'text-white text-opacity-25': step === 0,
-                  'fade-in text-[#EDA081]': step !== 0,
-                })}
-              >
-                Blocks found
-              </span>
-              <span
-                className={clsx(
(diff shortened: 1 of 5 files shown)


ui/lesson/HashrateChallenge.tsx
benalleng, line 1: I see the right padding removed for the contributionBar here but I also see the right padding returned for the profiles containers. maybe a rebase is in order?

benalleng: Oh i see nvm

content/lessons/chapter-3/coop-2.tsx
```diff
@@ -158,108 +158,110 @@ export default function Coop2({ lang }) {
...
-        protagonists={PROTAGONISTS}
-        antagonists={ANTAGONISTS}
-        profiles={PROFILES.map((profile, i) => (
-          <Profile
-            key={i}
-            username={profile.username}
-            avatar={profile.avatar}
-            description={profile.description}
-            className={'md:!max-w-[220px]'}
+    <div className="px-[20px] py-[30px] md:py-[75px]">
```
benalleng, line 161: With the addition of the right-padding here the overflow cutoff has returned for the profiles instead of at the edge of the screen they are cutof at the padding edge.

content/lessons/chapter-3/pool-3.tsx
```diff
@@ -179,104 +179,106 @@ export default function Pool3({ lang }) {
...
+                    className="font-space-mono font-normal"
+                    step={step}
+                    hashes={profile.hashes}
+                  />
+                </span>
+              </Card>
+            </Profile>
+          ))}
+        >
+          <div className={`flex flex-col ${showText ? 'fade-in' : 'fade-out'}`}>
```
benalleng, line 252: adding the right padding instead of at the parent container would allow all other elements to have padding, not including the profilesContainer

benalleng (changes requested):
Nice addition! Didn't know the height was adjustable like that. With the way the parent container was adjusted the ProfileContainer overflow cutoff has returned though

content/lessons/chapter-3/coop-2.tsx
GBKS: Yes, I added the back in so the scrollbar doesn't touch the edges and look weird (you can't tell if it's cut off or not). Ideally, we could control the position of the scrollbar itself. But since that's not possible, I think this is a good compromise.

content/lessons/chapter-3/pool-3.tsx
GBKS: Can you please explain this differently, I don't full understand.

benalleng: This is related to the above comment relating to the profileContainer extending all the way to the edge of the screen, since we are rolling back that change this can be disregarded.

For posterity adding padding to this text container would mean all elements on the page EXCEPT the profileContainer would show padding

PR #668, opened 2023-09-28
Elliptic curve visualizer

### LINKED ISSUE #593: Research SVG component for visualizing elliptic curves
Screenshot 2023-08-07 at 19 59 45

Figure out the best way to render something akin to the above, ideally it should support animation.

benalleng:
Closes #593 

[Preview](https://saving-satoshi-git-fork-benalleng-issue593-savingsatoshi.vercel.app/en/chapters/chapter-4/public-key-2?dev=true)


ecurrencyhodler:
Hmmm graph looks a bit strange when I change the width of the screen. This is what it looks like at 1046x740.
Screenshot 2023-09-27 at 7 52 00 PM

GBKS:
Really nice. A few tips for improvement:

- On mobile, the arrow and label are squeezed in between the graph and legend. Could the be positioned in the same place as on desktop?
- The pulsating circle is barely noticeable. I think you can make that a bit strong visually, and also slow down the animation quite a bit (more of a calm pulse)
- You shorten the pauses in between animation steps a bit more

I did not see the issue Andrew mentioned above, so I assume that's fixed.

Public key - Saving Satoshi

Another thing is that I cannot run this locally. npm can't figure out the dependencies when trying to install. Is there a particular node version that I need now, or something else?

benalleng:
Hmm not sure about the installation problem, It does add a new dependencey with react-lottie but I don't think that should be an issue

GBKS (approved):
Looks great. I think the pulse can be a bit smaller towards the end and still more visible, but happy to merge this as is.

PR #669, opened 2023-09-29
fix: issue with crashing when progress is in cha4

0tuedon:
Screenshot 2023-09-29 at 22 00 35

The issue on my local


benalleng (changes requested):
With these progress objects changed, we'll likely need to change them in each lesson content component as well.

0tuedon:
> With these progress objects changed, we'll likely need to change them in each lesson content component as well.



> With these progress objects changed, we'll likely need to change them in each lesson content component as well.

yh just updated this PR.

benalleng (approved):
Hmm I couldn't recreate your error but I think we normally use these 3 letter conventions anyways

PR #670, opened 2023-10-02
Match images to Figma

GBKS:
Various images were outdated, not crisp, or not the necessary minimum size for desktop retina screens. So I reviewed them and re-exported as needed. The [chapters overview page](https://saving-satoshi-git-feature-image-updates-savingsatoshi.vercel.app/en/chapters) should look very different now.

👊[Check the preview](https://saving-satoshi-git-feature-image-updates-savingsatoshi.vercel.app/en/chapters)✊🏼


adamjonas:
There are significant size savings by running these through imageoptim which just removes the metadata. Probably worth doing if we are going to switch these out.

Screenshot 2023-10-02 at 9 53 02 AM

GBKS:
If you are concerned about browser load performance, the framework already optimizes the images (size and compression).

adamjonas:
Yes, that was my main concern so not a necessary change, but generally I prefer to load images that have been run through imageoptim or the like so file sizes are only what we need.

adamjonas:
ACK on overview page and first couple of chapters

GBKS:
The images are now compressed.

benalleng (approved):
LGTM! Ran through the site and all the images look to match their figma counterparts

afterburn (approved):

afterburn:
We're also using [Next Images](https://nextjs.org/docs/pages/building-your-application/optimizing/images) which should optimize automatically for us

Issue #626, opened 2023-09-04
Implement CH4 Challenge 2 - Address generation


afterburn:
Use the `ScriptingChallenge` component to implement chapter 4 challenge 2. For our lesson IDs we can use `CH4AD1` and `CH4AD2`.

[Link to Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=1535-19614&mode=design&t=bUUrHbu3heWV0c7c-0)

Is currently blocked by https://github.com/saving-satoshi/saving-satoshi/issues/624

sabakhilji:
Can I work on this issue?

benalleng:
ATM blocked by #701 due to need for saveData => getData

PR #678, opened 2023-10-10
Implement CH4 challenge 2-Address generation

sabakhilji:
#626

for testing. I made some changes in the address-1 file to implement logic in the Python terminal.

[This PR has not been implemented completely]


content/lessons/chapter-4/address-1.tsx
```diff
@@ -46,30 +46,30 @@ const javascript = {
...
+# Get the sha256 digest of the compressed public key.
+# Then get the ripemd160 digest of that sha256 hash
+# Return 20-byte array
 # Type your code here
 `,
   validate: async (answer) => {
-    if (!answer.startsWith('00000')) {
-      return [false, 'Hash must start with 5 zeroes.']
+   
+    if (len(answer)!==20) {
```
benalleng, line 60: this should be `if (answer.length !== 20) {` remember even though this is checking the result of python code we are still using JS! Also if you are failing to compile remember that you can always run `yarn build` or `yarn lint` to check code quality

benalleng (changes requested):

content/lessons/chapter-4/address-1.tsx
sabakhilji: Thanks @benalleng it worked!!

benalleng (changes requested):
Coming along, a small length issue that might be stopping you

benalleng (approved):
TACK! LGTM

we will need to create a new PR to allow for code from @0tuedon 's challenge but I am happy with this ATM

PR #681, opened 2023-10-10
Fix avatars and typos

ottosch:
I noticed the player avatar in 2 pool screens were the generic `avatars/1.png`. Only in the follow up screens they were updated to the actual avatar.


benalleng (approved):
Good catch! Nice improvement thanks!

PR #689, opened 2023-10-16
StatusBar tweaks

GBKS:
The status bar was cut off on mobile because a fixed height was applied indiscriminately via "h-14". Only the top part of the button was visible. This is only supposed to the fixed on desktop, so I changed it to "md:h-14".

I also put the check mark into a round green circle to match the design. We changed this a while back to make the success state more noticeable.

image

Peace & happiness.


benalleng (approved):
LGTM! nice improvement!

PR #695, opened 2023-10-18
StatusBar div-child-of-p error

benalleng:
Fixes an error where the status bar has a div as a child of a p element, which is not allowed


0tuedon (approved):
LGTM!

PR #697, opened 2023-10-20
Fix editable lines on editor

benalleng:
fixes which lines are editable lines on the editor also moves the code below the comments


GBKS:
Looks good to me.

[Link tested](https://saving-satoshi-git-fork-benalleng-challeng-a640ea-savingsatoshi.vercel.app/en/chapters/chapter-4/public-key-3?dev=true).

Before:
image

After:
image

Probably a separate discussion from this PR, but the vertical line should be all the way, not just the inevitable lines. From Figma:

image

benalleng:
> the vertical line should be all the way, not just the inevitable lines

Could you explain what you mean by this? @GBKS

daviroo (approved):

PR #705, opened 2023-10-26
Terminal success message

### LINKED ISSUE #696: Terminal Success Message is not styled to the figma
Current success result in the terminal
image

And the figma
image

sabakhilji:
closes #696 
I tried to add pseudo-elements for success message styling.
Screenshot (131)
Screenshot (132)


0tuedon:
> closes #696 I tried to add pseudo-elements for success message styling. Screenshot (131)

This might cause more issues with other challenges, is there a more dynamic way to add them iike in the Terminal.tsx something like styles as  props?

benalleng (changes requested):
Not quite there.. just testing it out with scripting-2 [here](https://saving-satoshi-git-fork-sabakhilji-termina-7e317a-savingsatoshi.vercel.app/en/chapters/chapter-2/scripting-2) it seems to add scrolling but doesn't change the background color or make the text size bigger.

Ideally we want a result that has  a slightly larger text size and a background with the correct color
image

If you don't think the line-breaks are possible I wouldn't worry too much about them as every answer will be formatted slightly different

PR #726, opened 2023-10-31
feat: Added privacy statement

daviroo:
## Pull Request checklist

- Clean commit history

### DIFF
--- a/app/[lang]/about/page.tsx
+++ b/app/[lang]/about/page.tsx
@@ -100,7 +100,7 @@ export default function About({ params }) {
           <h3 className="mt-4 text-2xl sm:text-3xl">
             {t('about.privacy.sub_heading_two')}
           </h3>
-          <p className="mt-2 font-nunito text-xl">
+          <div className="mt-2 font-nunito text-xl">
             {t('about.privacy.paragraph_three')}
             <ol>
               <li>
@@ -116,7 +116,7 @@ export default function About({ params }) {
                 {t('about.privacy.list_item_two_text')}
               </li>
             </ol>
-          </p>
+          </div>
           <h3 className="mt-4 text-2xl sm:text-3xl">
             {t('about.privacy.sub_heading_three')}
           </h3>
--- a/app/[lang]/about/page.tsx
+++ b/app/[lang]/about/page.tsx
@@ -87,6 +87,46 @@ export default function About({ params }) {
           </p>
 
           <Divider />
+          <h3 className="text-4xl sm:text-5xl">{t('about.privacy.title')}</h3>
+          <p className="mt-4 font-nunito text-xl">
+            {t('about.privacy.paragraph_one')}
+          </p>
+          <h3 className="mt-4 text-2xl sm:text-3xl">
+            {t('about.privacy.sub_heading_one')}
+          </h3>
+          <p className="mt-2 font-nunito text-xl">
+            {t('about.privacy.paragraph_two')}
+          </p>
+          <h3 className="mt-4 text-2xl sm:text-3xl">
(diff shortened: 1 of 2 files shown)


app/[lang]/about/page.tsx
```diff
@@ -86,6 +86,32 @@ export default function About({ params }) {
...
+          </h3>
+          <p className="mt-2 font-nunito text-xl">
+            {t('about.privacy.paragraph_two')}
+          </p>
+          <h3 className="mt-4 text-2xl sm:text-3xl">
+            {t('about.privacy.sub_heading_two')}
+          </h3>
+          <p className="mt-2 font-nunito text-xl">
+            {t('about.privacy.paragraph_three')}
+          </p>
```
Comment (benalleng) [inline app/[lang]/about/page.tsx:105]: This `<p>` tag cannot be the parent of the `<ol>`

i18n/locales/en.ts
```diff
@@ -65,6 +65,25 @@ const translations = {
...
+      title: 'Our commitment to privacy',
+      paragraph_one:
+        'At Saving Satoshi, we are deeply committed to ensuring the privacy and security of our users. We believe in transparency and want you to understand how and why we track app usage. To achieve this, we use a powerful tool called <Link href="https://umami.is/" className="underline" target="_blank">Umami</Link>.',
+      sub_heading_one: 'What is Umami?',
+      paragraph_two:
+        'Umami is an open-source analytics platform that helps us gather essential insights about how you use our app. It allows us to make data-driven decisions and continuously enhance our product to better meet your needs.',
+      sub_heading_two: 'What Data Do We Collect?',
+      paragraph_three: `Rest assured, we only collect non-personal and anonymized data, such as: 
+      <ol>
+      <li><span className="font-bold">Product Improvement:</span> We use the data to identify areas where our app can be improved. By understanding how you use our app, we can make it more efficient, user-friendly, and secure.</li>
```
benalleng, line 78: upon testing, it seems to be favoring rendering this span element before the list, upon removing the span I was able to get it to work marginally better

i18n/locales/en.ts
```diff
@@ -65,6 +65,25 @@ const translations = {
...
+      <li><span className="font-bold">Compatibility:</span>Understanding the devices and platforms you use allows us to optimize our app for various configurations, ensuring it works seamlessly for you.</li>
+      </ol>`,
```
benalleng, line 80: It also really doesn't seem to play nivcely with having these line breaks, it works better if there are no line-breaks

benalleng (changes requested):
I think the issue seems to be three-fold: the addition of the `<span>` element inside of the `<ol>` tag seems to be causing the most problems but as well the `<ol>` is inside of a `<p>` tag causes errors. Lastly the line-breaks seem to make the regex not able to recognize it

lib/translate.tsx
```diff
@@ -153,6 +156,30 @@ function injectComponent(result, type) {
...
+            ? liMatches.map((liHtml, itemIndex) => {
+                const liLabel = getFirstMatch(liHtml, labelRegex)
+                const liClassName = getFirstMatch(liHtml, classNameRegex)
+                return (
+                  <li key={`item-${itemIndex}`} className={liClassName}>
+                    {liLabel}
+                  </li>
+                )
+              })
+            : []
```
benalleng, line 172: I think the `<span>` element is recognized and rendered before everything inside the `<li>` tag is rendered so thats why removing the `<span>` element allows it to work.

daviroo:
@benalleng I changed my approach entirely so I've marked the other comments as resolved.

daviroo:
@adamjonas do you have any comment on the copy?
image

benalleng (approved):

PR #729, opened 2023-11-01
Release 1.0.6

daviroo:
- Adds Chapter 4 to production
- Adds Tracking with Umami


daviroo:
@benalleng I've raised a new release PR but I've named it with the same version - I think if we merge this without squashing then we will redeploy without losing commit history (if I increment the version in the PR title then we will lose your other 1.0.6 merge commit which will mean we will be skipping that version number in the commit history).

benalleng (approved):

PR #730, opened 2023-11-01
Next-env change

benalleng:
Our next-env file has a new comment change, wanted to add this to make sure not everyone has this as a permanent diff on local


daviroo (approved):
I'm also seeing this diff locally

PR #736, opened 2023-11-03
Derive message input challenge

### LINKED ISSUE #712: Chapter 5 Challenge-1 Part-1 (signature)
This challenge uses a modification of the `<Input />` or `<Hashing />` component so that there is spacing between byte splits every 4 characters

image

Ideally this PR should add both challenge pages as they are very similar in scope

image

[Figma](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=5327%3A8447&mode=dev)

benalleng:
Closes #712 

This does not add the input spacing as shown in the image Figma but I believe that the dev effort required for that change ins't worth it for now. I think we can circle back to this if/when we do an input challenge refactor

[Preview](https://saving-satoshi-git-fork-benalleng-issue716-savingsatoshi.vercel.app/en/chapters/chapter-5/derive-message-3?dev=true)


daviroo (approved):
Looks good just needs a rebase

daviroo (approved):
LGTM

PR #738, opened 2023-11-06
Adjust object position for intro component images

GBKS:
Various images were cut off weirdly at various screen sizes. Reason is that our layouts provide specific aspect ratios. We instruct the browser to fill the available area with the images. The browser needs to decide how to crop and center them in the given area. With this update, we now use the object position CSS property to provide specific points of interest in each image. Basically like a coordinate in the image that should be visible as much as possible. I looked at each image and calculated the coordinates (defined as percentages in Tailwind).

I removed the media query logic because it's not needed. Just by defining the point of interest in the image, the browser can figure out the cropping.

Example of a before and after.

image

📸[Check the preview](https://saving-satoshi-git-feature-adjust-image-centering-savingsatoshi.vercel.app/en/chapters/chapter-5/intro-1?dev=true)🌾
[Same page on dev](https://dev.savingsatoshi.com/en/chapters/chapter-5/intro-1?dev=true)

### DIFF
--- a/content/lessons/chapter-1/intro-2.tsx
+++ b/content/lessons/chapter-1/intro-2.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { useTranslations, useMediaQuery } from 'hooks'
+import { useTranslations } from 'hooks'
 import { Introduction, Text } from 'ui'
 
 export const metadata = {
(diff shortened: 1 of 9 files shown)


benalleng (approved):
Nice! the user of the object percents is smart, defiitely better than the usemediaquery hooks

PR #753, opened 2023-11-15
feat: Chapter 5 challenge 1 part 4

daviroo:
- Changes to runner to support console logs during runtime
- Added verify signature 2
- verify signature 5 now supports console logs during runtime
## Pull Request checklist

- Clean commit history

### DIFF
--- a/content/lessons/chapter-5/verify-signature-2.tsx
+++ b/content/lessons/chapter-5/verify-signature-2.tsx
@@ -44,7 +44,7 @@ console.log("KILL")
   },
   constraints: [
     {
-      range: [4, 1, 6, 1],
+      range: [4, 1, 7, 1],
       allowMultiline: true,
     },
   ],
--- a/content/lessons/chapter-5/verify-signature-2.tsx
+++ b/content/lessons/chapter-5/verify-signature-2.tsx
@@ -1,19 +1,104 @@
 'use client'
 
 import { useMediaQuery, useTranslations } from 'hooks'
+import { getLessonKey } from 'lib/progress'
 import { useEffect, useState } from 'react'
-import { Introduction, Text } from 'ui'
+import { EditorConfig } from 'types'
+import { LessonInfo, ScriptingChallenge, Text } from 'ui'
 
 export const metadata = {
-  title: 'chapter_five.verify_signature_one.title',
-  image: '/assets/images/chapter-4-intro-1.jpg',
+  title: 'chapter_five.verify_signature_two.title',
   key: 'CH5VFS2',
+}
+
+const satoshiTransactionMessage =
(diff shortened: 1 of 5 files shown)


daviroo:
Closes #751

content/lessons/chapter-5/verify-signature-2.tsx
```diff
@@ -1,19 +1,104 @@
...
+  validate: async (answer) => {
+    if (answer !== correctAnswer) {
+      return [false, 'Hash is not valid']
+    }
+
+    return [true, undefined]
+  },
+  constraints: [
+    {
+      range: [4, 1, 6, 1],
```
benalleng, line 44: This range needs to be changed so that the user can edit the code to the end

content/lessons/chapter-5/verify-signature-2.tsx
```diff
@@ -1,19 +1,104 @@
...
+
+    return [true, undefined]
+  },
+  constraints: [
+    {
+      range: [1, 1, 4, 1],
+      allowMultiline: true,
+    },
+    {
+      range: [6, 1, 8, 1],
```
benalleng, line 80: doubled up here

benalleng (changes requested):

benalleng (commented):
The validation process doesn't seem to work as expected, I returned what I think the correct answer is but no success

image

daviroo:
> The validation process doesn't seem to work as expected, I returned what I think the correct answer is but no success
> 
> image

This was happening because it was returning a BigInt not a string which means the number had an "n" appended to it so it failed validation.  Try running the `toString()` function on the BigInt when you return.

content/lessons/chapter-5/verify-signature-2.tsx
daviroo: Not sure what you mean - I am able to edit the code inside the function on my side.  Are you seeing different behaviour?

daviroo:
@benalleng I've raised this PR for the challenge code to reflect what I think should be the correct setup https://github.com/saving-satoshi/challenges/pull/5

content/lessons/chapter-5/verify-signature-2.tsx
benalleng: I just like to be able to edit code to the end, I know if i were to delete the closing bracket it returns an error but I like being able to edit the final 2 lines anyway, just personal preference and consitency

benalleng:
My question is why does the validation not return false with the "n" character added when its still a BigInt

daviroo:
> Why does the validation not return false with the "n" character added when its still a BigInt?

It was actually but I fixed it in this commit -= https://github.com/saving-satoshi/saving-satoshi/pull/753/commits/08f16266856905976b956fb60f87704481ae5af2

benalleng:
so, I mean that the validation only returns true when correct, It doesn't return false when wrong it only seems to return the log but doesn't run any validation
```
if (cleanedAnswer !== correctAnswer) {
      return [false, 'Hash is not valid']
    }
````
This code above doesn't run

File: ui/lesson/ScriptingChallenge/Runner/index.tsx [hunk 3]
```diff
@@ -129,23 +129,15 @@ export default function Runner({
...
-            if (!res) {
-              setHasherState(HasherState.Error)
-              setIsRunning(false)
-              setState(State.Complete)
-              sendTerminal('error', err)
-              ws?.close()
-              return
-            }
-
+            if (!res) break
```
benalleng, line 140: I think it is here where the break occurs instead of returning the error, I guess this is what you had discussed with removing some of the challenge specific errors?

content/lessons/chapter-5/verify-signature-2.tsx
daviroo: That's fair.  I'll change it so that it's editable to the last line.

content/lessons/chapter-5/verify-signature-2.tsx
daviroo: This is ok (I think) - it just means you can edit before the function definition and after.

benalleng (approved):
LGTM!

PR #755, opened 2023-11-21
Remove console.log() and print() requirement for challenges

benalleng:
Allows challenges to be completed without console logging, also improves the scripting-2 validation

(scripting-2 validation will take some more thought)


content/resources/chapter-4/public-key.tsx
```diff
@@ -25,21 +24,22 @@ const encodedPrivateKey = BigInt(0x\${privateKey})
 const generatorPoint = G.mul(encodedPrivateKey)
 // Remember you need to log the answer for it to be validated
 console.log(generatorPoint)`,
```
daviroo, line 26:
```suggestion
return generatorPoint`,
```

content/resources/chapter-4/public-key.tsx
```diff
@@ -60,20 +60,18 @@ encoded_private_key = int(0x\${private_key})
 const generatorPoint = G.mul(encoded_private_key)
 # Remember you need to log the answer for it to be validated
 print(generator_point)`,
```
daviroo, line 62:
```suggestion
return generator_point`,
```

content/resources/chapter-4/public-key.tsx
daviroo: There's also an error in line 22 (not made by you):

```const encodedPrivateKey = BigInt(0x\${privateKey})```

Should be:

```const encodedPrivateKey = BigInt(`0x\${privateKey}`)```

File: content/resources/chapter-4/public-key.tsx (hunk 2 above)
Comment (daviroo) [reply content/resources/chapter-4/public-key.tsx:62]: @benalleng the python solution for this challenge doesn't appear to be valid - again not your doing but we should probably fix it.

File: content/lessons/chapter-4/address-2.tsx [hunk 3]
```diff
@@ -29,7 +29,9 @@ export default function Address2({ lang }) {
   }, [])
 
   const javascript = {
-    program: `console.log("KILL")`,
+    program: `
```
daviroo, line 32: We need to update the solution in `address.tsx` in resources to reflect this change.

content/lessons/chapter-4/address-2.tsx
```diff
@@ -66,14 +71,16 @@ ${prevData?.data && 'compressedPublicKey = ' + "'" + dataObject + "'"}
...
-        range: [9, 1, 9, 1],
+        range: [9, 1, 11, 1],
         allowMultiline: true,
       },
     ],
   }
 
   const python = {
-    program: `print("KILL")`,
+    program: `
```
daviroo, line 81: We need to update the solution in `address.tsx` in resources to reflect this change.

content/lessons/chapter-4/address-3.tsx
```diff
@@ -30,7 +30,9 @@ export default function Address3({ lang }) {
   }, [])
 
   const javascript = {
-    program: `console.log("KILL")`,
+    program: `
```
daviroo, line 33: We need to update the solution in `address.tsx` in resources to reflect this change.

content/lessons/chapter-4/address-3.tsx
```diff
@@ -71,14 +76,16 @@ ${
...
-        range: [166, 1, 166, 1],
+        range: [166, 1, 168, 1],
         allowMultiline: true,
       },
     ],
   }
 
   const python = {
-    program: `print("KILL")`,
+    program: `
```
daviroo, line 86: We need to update the solution in `address.tsx` in resources to reflect this change.

daviroo (changes requested):

content/resources/chapter-4/public-key.tsx
benalleng: had an incorrect case on a param

daviroo (approved):

PR #785, opened 2023-12-04
Fix public key and address 2 validation

### LINKED ISSUE #784: Playing ch. 4, nits and issues
- Even though I am doing each step in python, when i click "next" the next challenge starts off in the javascript REPL

- https://savingsatoshi.com/en/chapters/chapter-4/public-key-4 needs to ask for a hex string, not `"# Return 33-byte array"`

- https://savingsatoshi.com/en/chapters/chapter-4/address-2 not passing for me like expected. Also needs to ask for a hex string but either way, I'm doing something wrong...?

```python 
import hashlib

compressed_public_key = '0339db9056e80c9f5cf13bbf0c6c24bad7a4c16d477f8da829dc50aea0e3f9290c'

# Get the sha256 digest of the compressed public key.
# Then get the ripemd160 digest of that sha256 hash
# Return 20-byte array
def hash_compressed(compressed_public_key):
    return hashlib.new('ripemd160', hashlib.new('sha256', bytes.fromhex(compressed_public_key)).digest()).digest().hex()
```

```
Result
d9b7e9f4c8b78a5db485c62efabca252e384a611
Ensure you are using the correct compressed key and it is being decoded
```

benalleng:
Closes #784 
- Fixes the python and javascript validation for public-key-4 and address 2
- Fixes the private key object  in private-key-3 to be parsed properly 
- Changes the expected answer copy to ask for a hex string instead of bytes

[Preview](https://saving-satoshi-git-fork-benalleng-issue784-savingsatoshi.vercel.app/en/chapters/chapter-4/public-key-3?dev=true)

### DIFF
--- a/content/lessons/chapter-4/address-2.tsx
+++ b/content/lessons/chapter-4/address-2.tsx
@@ -48,7 +48,6 @@ function hashCompressed(compressed) {
 }
 `,
     validate: async (answer) => {
-      console.log(dataObject, answer)
       if (answer.startsWith('<Buffer')) {
         return [false, 'Ensure you are properly decoding your answer']
       }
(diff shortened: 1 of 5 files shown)


0tuedon (commented):
Was able to test it on local and passed the challenges from the  help pages code. LGTM!

PR #794, opened 2023-12-07
Chapter 4/5 update #2

benalleng:
Looks like we diverged from the challenge repo made a change to see if this is right for address 2

also fixes some copy

[Preview](https://saving-satoshi-git-fork-benalleng-challeng-ac8ca6-savingsatoshi.vercel.app/en/chapters/chapter-4/address-2?dev=true)

### DIFF
--- a/content/lessons/chapter-4/address-2.tsx
+++ b/content/lessons/chapter-4/address-2.tsx
@@ -65,16 +65,7 @@ function hashCompressed(compressedPublicKey) {
                 .digest()
             )
             .digest('hex') ||
-        answer ===
-          crypto
-            .createHash('ripemd160')
-            .update(
-              crypto
-                .createHash('sha256')
-                .update(Buffer.from('', 'hex'))
-                .digest()
-            )
-            .digest('hex')
+        answer === 'b472a266d0bd89c13706a4132ccfb16f7c3b9fcb'
       ) {
         return [
           false,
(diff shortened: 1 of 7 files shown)


benalleng:
@pinheadmz when testing, you won't really be able to reach a success page since logging in on this branch is blocked but as long as you are replacing `compressed_public_key` in the function with the actual hex string you can verify the value

pinheadmz (commented):
One suggestion below and a few other things I noticed playing ch4:

## public-key-3:

"Create a script so that it logs out the public key derived from a given private key."

Should be

"Complete the function which accepts a private key as a hex-encoded string and returns the corresponding public key as a GE (Group Element) object."


## public-key-4:

Challenge code change to accommodate strings:
https://gist.github.com/pinheadmz/82e6db34d37d0735ea0d1af40a5004c3


## address-2:

"Complete a function that accepts a 33-byte compressed public key and returns a 20-byte public key hash."

should be

"Complete a function that accepts a 33-byte compressed public key as a hex string and returns a 20-byte public key hash as a hex string."



## address-3

- Replace the word "Signet " with "Testnet"

- Would be nice if the code had a link to the library it refers to: https://github.com/saving-satoshi/challenges/blob/master/chapter4/javascript/lib/bech32.js (and the python version)

- hash-to-address is missing its argument:
https://gist.github.com/pinheadmz/6cf78dc0c9d4f035da601151854b7a2f

benalleng:
I addressed everrything you brought up in your last comment!

pinheadmz (approved):
Great work thanks!

PR #804, opened 2023-12-20
Add temp foss banner

### LINKED ISSUE #802: Add Start Career in FOSS Banner
See here: https://github.com/bitcoin-dev-project/bitcoin-dev-project/issues/27

0tuedon:
This PR closes #802 , Although we may need to have discussion if the banner design is okay

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -12,14 +12,16 @@ export const metadata = {
 
 export default function Home() {
   return (
-    <div className="flex h-screen w-screen flex-col">
-      <Umami />{' '}
-      <div className="bg-gradient-to-b from-[#00000060] via-[#00000020] to-transparent pb-28">
-        <Topbar />
+    <div className="flex flex-col">
+      <div className="relative flex h-screen w-screen flex-col">
+        <FossBanner />
+        <Umami />{' '}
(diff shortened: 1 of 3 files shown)


GBKS:
Here's a mock-up for the design. What do you think?

image

GBKS:
You could also place it at the very top above the navigation. I really would avoid placing it over the image though, because it easily covers up the eyes/face, especially on mobile where the spacing is really tight.

image

0tuedon:
> You could also place it at the very top above the navigation. I really would avoid placing it over the image though, because it easily covers up the eyes/face, especially on mobile where the spacing is really tight.
> 
> image

No need for a close icon then ?

GBKS:
No need to close it if it doesn't cover anything up. Also, this will be only on the home page, right? No other pages.

0tuedon:
> No need to close it if it doesn't cover anything up. Also, this will be only on the home page, right? No other pages.

yh updated with a new commit

adamjonas:
I'd like to put a hold on this. I don't think this is the target audience for this and we've had plenty of applications already. 

Secondarily, the current implementation on mobile becomes the focus rather than an announcement. I'd prefer this effort be put towards other features.

Issue #532, opened 2023-06-26
Update profile descriptions for accessibility


ecurrencyhodler:
Update saving satoshi profiles with the following for accessibility:

Select avatar with...
- White space-suit
- red space-suit
- orange space-suit
- bunny engineer
- badger astronaut

_Originally posted by @benalleng in https://github.com/saving-satoshi/saving-satoshi/issues/529#issuecomment-1601578418_

PR #806, opened 2024-01-03
fix: add descriptive a11y avatar label

umarluqman:
This PR closes https://github.com/saving-satoshi/saving-satoshi/issues/532


benalleng (changes requested):
Ideally I think we should make sure the file names don't contain any spaces for consistency and making sure there are no problems when modifying them in the terminal

umarluqman:
That makes sense, I already pushed the fix.

benalleng (approved):
LGTM! I went ahead and made a new PR #809 to fix the avatars in chapter 3

benalleng (changes requested):
Unfortunately catching this one late before merging but our backend saves the avatars as their old denominations `1-5.png` so with this new change any old account would not be able to load in their chosen avatar. You can reproduce this behavior by logging in with an existing account on this branches preview server and then observe the lackof an avatar in the top right

benalleng:
Instead of changing the names of the files perhaps we could create a simplle array to use as the aria-labels mapped to the index

benalleng (approved):
LGTM

Issue #808, opened 2024-01-04
Adding german as a language option


bTCpy:
**Describe the enhancement**
Hey everybody, I would like to provide a .ts file for the german language. I'm a native speaker and love this project.

**Code Snippet**
const translations = {
  shared: {
    next: 'Fortfahren',
    start: 'Start',
    copy: 'Kopieren',
    copy_acknowledged: 'Kopiert!',
    about: 'Über uns',
    chapter: 'Kapitel',
    chapters: 'Kapitel',
    challenge: 'Herausforderung',
.....

**Additional information**
I'm confident to provide the translated .ts file today or a least until the end of the week.

0tuedon:
> **Describe the enhancement** Hey everybody, I would like to provide a .ts file for the german language. I'm a native speaker and love this project.
> 
> **Code Snippet** const translations = { shared: { next: 'Fortfahren', start: 'Start', copy: 'Kopieren', copy_acknowledged: 'Kopiert!', about: 'Über uns', chapter: 'Kapitel', chapters: 'Kapitel', challenge: 'Herausforderung', .....
> 
> **Additional information** I'm confident to provide the translated .ts file today or a least until the end of the week.

Sounds great @bTCpy

benalleng:
If you want to open a PR with the german file added go ahead! If you want we can add in some additional edits to get your translation working with the codebase

benalleng:
Closed by #810

PR #810, opened 2024-01-08
Adding german as a language option

bTCpy:
See Issue https://github.com/saving-satoshi/saving-satoshi/issues/808#issue-2066234969

Translated up to and including chapter 5.  Happy to keep going as soon as additional chapters are available.

### DIFF
--- a/i18n/config.ts
+++ b/i18n/config.ts
@@ -5,6 +5,7 @@ export const i18n = {
     { locale: 'es', label: 'Español' },
     { locale: 'nl', label: 'Dutch' },
     { locale: 'hi', label: 'Hindi' },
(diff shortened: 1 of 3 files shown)


benalleng (changes requested):
Looks great!

few small things, I think the about translation should try to keep is as on word if possible otherwise we may need to adjust some css to keep it all in one line

image

bTCpy:
> Looks great!
> 
> one small thing, I think the about translation we should try to keep is as on word if possible otherwise we may need to adjust some css to keep it all in one line
> 
> image

I absolutely agree. Changed it to "Info". It's about the same size and means basically the same thing.

benalleng (approved):

PR #814, opened 2024-01-17
Info Challenge Tab should be translation objects

### LINKED ISSUE #811: Info Challenge Tab should be translation objects
On the chapters page the challenge and info tab are not translation objects

image

0tuedon:
This pr close #811


benalleng (approved):
tACK LGTM!

PR #824, opened 2024-01-27
Remove extra slash in README

atsu6:
Adding a / at the beginning incorrectly implies a root directory rather than the repository root, which can lead to unintended paths when using commands like cd or opening with an editor.


daviroo (approved):
Yeah you're right this is clearer - thanks

PR #839, opened 2024-02-01
Chapter 7 Mempool Transaction 1

### LINKED ISSUE #828: Chapter 7 Challenge Page
Implement the challenge page for chapter 7

image

https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?node-id=7100%3A9325&mode=dev

benalleng:
Closes #828 
[Preview](https://saving-satoshi-git-fork-benalleng-issue828-savingsatoshi.vercel.app/en/chapters/chapter-7/mempool-transaction-1?dev=true)


daviroo (changes requested):
I finally got my local docker working 😓 .  I tested it in python only as that's the only solution we have available right now.  The good news is the happy path works - there's just a couple of things I would suggest as changes:

1. A failed run in python displays an error indicating that there's something wrong in the code (there isn't it's just failed validation). e.g.
image

2. Can we change the message displayed to the user while they are waiting on the run to complete?  Right now it says "This should just take a few seconds" but on my testing it took quite a bit longer than that to complete.  Maybe we could default that message to the current value but override it on the level of a lesson so we can change it to "... up to a minute" in this case?

benalleng:
I also removed the raise Exceptions from the test code so It shouldn't return an error on a failed validation unless there actually is an error

daviroo (approved):

PR #842, opened 2024-02-06
feat: Added ability to fold code on load

daviroo:
The intention here is to be able to provide range blocks to fold in the editor.  Unfortunately the ability to pass in ranges to the fold function in the editor is currently broken in the version we are using.  As a workaround I'm using the `foldAllExcept` action instead.  This means that for now it only supports one element in the `rangesToCollapse` array but we should revisit this once the [fix](https://github.com/microsoft/vscode/pull/201315) is available in the [editor](@monaco-editor/react
) we are using.

## Pull Request checklist

- Clean commit history

### DIFF
--- a/content/lessons/chapter-6/put-it-together-5.tsx
+++ b/content/lessons/chapter-6/put-it-together-5.tsx
@@ -156,8 +156,8 @@ console.log("KILL")
   `,
     rangesToCollapse: [
       {
-        start: 115,
-        end: 136,
+        start: countLines(prevData.data) + 1,
+        end: countLines(prevData.data) + 22,
       },
     ],
     defaultFunction: {
@@ -319,8 +319,8 @@ print("KILL")
 `,
     rangesToCollapse: [
       {
-        start: 78,
-        end: 87,
+        start: countLines(prevData.data) + 1,
+        end: countLines(prevData.data) + 10,
       },
     ],
     defaultFunction: {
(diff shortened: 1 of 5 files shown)


content/lessons/chapter-6/put-it-together-5.tsx
```diff
@@ -154,6 +154,12 @@ const public_key_yirutads = ec.keyFromPublic(wit_key_ttefkgdk);
 console.log(public_key_yirutads.verify(hashed_message_bytes_aieoprds, wit_sig_okdakljl) && 'true');
 console.log("KILL")
   `,
+    rangesToCollapse: [
+      {
+        start: 115,
+        end: 136,
+      },
+    ],
```
benalleng, line 162: since the range is able to be anything as this will be collapsing user-generated code we will need to make this a number created from a regex of some kind

benalleng (changes requested):
Works great! looks like we will need to adjust the range to account for user generated code but I was able to collapse it when the line numbers were just right!

content/lessons/chapter-6/put-it-together-5.tsx
daviroo: That one slipped past me :).  Was too focussed on trying to get the collapsing to work in the editor.  Should be fixed now.

benalleng (approved):

Issue #674, opened 2023-10-04
as a learner, when I hit the continue button, the previous screen loads first


adamjonas:
https://github.com/saving-satoshi/saving-satoshi/assets/755825/8fd425c3-9a4c-462a-8e8f-1ddfeb52914d

GBKS:
If you look scrub slowly and look closely, you can see that the lock for challenge 1 very briefly appears. That might be a clue that the functionally that handles checking/updating the users progress might be the problem.

My guess is that's related to `useSaveAndProceed` somehow. I have not verified this, but it might be that it's triggering a load screen while it's storing the user progress both locally and via the API. That takes a split second during which the loader is shown. Then it goes back to the current page and continues with the navigation process, which shows another loader and then finally the next page. If that is the case, it might be best to save progress asynchronously and not bother the UI with it at all.

Apart from this issue, do we even need a loader at all? Can't we expect the page to load so fast that we don't need to blank out the screen? Often, it's good not to show a loader until the load period has taken 300-500ms and the user becomes aware of the lag.

benalleng:
I believe this is correct, looking briefly at this I think it may relate to the fact that checking the progress is queried at the page load and not before that. perhaps adding a check ahead of time would allow this progress load and as a result loading spinner could go away

GBKS:
Since the lock briefly appears, there might also be an unnecessary re-rendering of the page happening with an incorrect state (there's no reason why the user should be not have access to the challenge they are currently on).

0tuedon:
Been having a very tough time solving this need some more eyes here cc: @benalleng  @daviroo

benalleng:
For this issue, Improvements can still be made as the progress context and account context fight against each other causing the re-render. Given the chance I believe we can solve this problem in a more elegant way

PR #846, opened 2024-02-13
Alter progress logic and next routing to prevent page flashes

benalleng:
works towards fixing #674


benalleng:
currently having an issue where the redirect takes me back much further than intended, also the isDevelopment tag doesn't work as expected but i could again try an early return statement for dev=true

0tuedon:
asides always seeing the locks flash, (should be solved in the other PR) there are some times it delays for 1.5-2s before moving )again tried reproducing but couldn't.

daviroo (approved):
Seems to fix the issue for now.

PR #849, opened 2024-02-19
Remove Unused tailwind classes in Hashrate Component

### LINKED ISSUE #584: Resolve Christoph's feedback for HashrateChallenge component
From @GBKS:

I recommend being more deliberate with Tailwind classes. There's just a lot of cruft in there. For example:

An element that has both a fixed with and a min-width (which will have no effect).
image

This element does not need flex, since all children are absolutely positioned.
image

An unnecessary element.
image

A button that only contains text does not need any flex layout for its children (there is only one).
image

The parent has layout rules for its children, which all children overwrite. Text has nunito declared twice as a font.
image

Lots of mixing of flexbox and full width/height. Usually, when you use flexbox, the container sets rules for how the children resize. The children can express a preferred size using flex-basis, and shrink/grow. Using w-full and h-full all over the place is a weird mix of two !
image

0tuedon:
This PR closes #584 a backlog issue from Chapter 3.


benalleng (approved):
Went through chapter 3 and everything looked the same so I'm satisfied all these removed classes do nothing :+1:

PR #856, opened 2024-02-22
Removed run script button from mobile info

### LINKED ISSUE #853: Run program bar is visible on mobile on Info Tab
We don't want to see the run the program button on the info tab in mobile

It should only be visible on the Execute Tab

image

0tuedon:
This PR closes #853


benalleng (approved):
Works great!

PR #863, opened 2024-02-26
chore: Laying some groundwork for save/load code as we go

daviroo:
It's quite a large diff but it only really does 2 things:
- Renames the Providers to Contexts (this is simply a code tidying step to make it easy to find our Contexts)
- Upgrades the monaco editor version to the latest available


benalleng (approved):

benalleng (commented):
Looks like the yarn lock needs a rebase and then we're good to merge

PR #866, opened 2024-02-27
chapter-3/split-1 title format fix

GBKS:
All the page titles seem to usually use the `Title` component in the `ui` folder. This page (chapter-3/split-1) used the `Title` component in the `ui/introduction` folder, which did not apply the correct formatting. So I switched over.

Reason why I am keeping this a draft is that the `ui/introduction/Title` component appears to be no longer used in the site at all. Is this a legacy component that can be removed? What about other components in the `ui/introduction` folder? If so, should I remove it in this PR?

Before and after (on different devices):

image

### DIFF
--- a/content/lessons/chapter-3/split-1.tsx
+++ b/content/lessons/chapter-3/split-1.tsx
@@ -4,7 +4,7 @@ import { useEffect } from 'react'
 import { useTranslations } from 'hooks'
 import { Introduction, Text } from 'ui'
 import { cssVarThemeChange } from 'lib/themeSelector'
-import Title from 'ui/introduction/Title'
+import { Title } from 'ui'
 
 export const metadata = {
   title: 'chapter_three.split_one.title',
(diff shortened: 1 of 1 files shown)


benalleng:
Looking at it I think the reason we haven't used it is because It actually doesn't do anything it looks like this.
```
export default function Title({ children }) {
  return <span>{children}</span>
}
```

we probably should be using it to make sure we don't miss stuff like we did on that page

GBKS:
> we probably should be using it to make sure we don't miss stuff like we did on that page

What do you mean by this?

And yes, that component seems pretty useless.

benalleng:
Just that the Title component indicates it should be bigger than the rest, but the more I think about it i think this isn't important.

You can get rid of it if you'd like and just add the className to the standard `<Text>` component

GBKS:
Taking a closer look, I think more components can be removed from that folder, maybe even all? That might be better suited as a separate PR, so I removed the draft status from this one.

benalleng (approved):

PR #868, opened 2024-02-27
Speed up chapter 2 mining challenge

benalleng:
Now uses a "100x" speed up to dramatically cut the length of the challenge

[Mining Challenge Preview](https://saving-satoshi-git-fork-benalleng-mining-c-72dcee-savingsatoshi.vercel.app/en/chapters/chapter-2/mining-1?dev=true)

This PR does NOT address any narrative inconsistencies in the mining challenge


adamjonas:
Can we speed up the first block bit as well? It's is still taking about 18 seconds. I think under 8-10 is more reasonable.

benalleng:
Should be quite a bit faster now, I shortened the first block to only 5 seconds

0tuedon (approved):
LGTM!

PR #876, opened 2024-02-28
New Account UX

benalleng:
Will add the following

- [x] show's the user's private key with a toggle switch in their account Modal

image

- [x] New "Login to continue" button block on chapters 2+ in the 'chapters` page to prevent the user from being locked at scripting challenges if they haven't logged in

image


GBKS:
For the first part, is it useful to blur the code? It's not something that can be easily read or memorized. How about using the same visual on the left in the screenshots below - the box with the private key and the copy button? Visually also notice the difference in the dashed/dotted outline and the label font weight.

image

We also start by using the terms `code` and `personal code`, and then switch over to `private key`. Should we be consistent with `personal code` here?

I am not sure what the issue is you are looking to address in the second part. Right now, we say `Complete Chapter X to unlock`, which seems pretty clear. Although there is text duplication that we can clean up, as in the screenshot below (for example, just hide the disabled button altogether).

image

Also, we have this state [in the design](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=2019%3A22933&mode=design&t=XdtqyCLop8kVVTJR-1) that confirms the login and lets the user know about their current progress with a simple `Continue` option. Maybe something to consider?

image

benalleng:
> Also, we have this state [in the design](https://www.figma.com/file/LqjK3Tpvd9KJ4buFArCJBQ/Saving-Satoshi?type=design&node-id=2019%3A22933&mode=design&t=XdtqyCLop8kVVTJR-1) that confirms the login and lets the user know about their current progress with a simple `Continue` option. Maybe something to consider?

I ended up using a fragment router to scroll to the user's progress

GBKS:
How about something subtle like this?

image

GBKS:
I think this works. Maybe good if a few other people test this. Feels like we have come back to the login system a few times before, maybe we can reduce iteration by having more testers here?

PR #878, opened 2024-03-03
Holocat Question link to chatbtc

### LINKED ISSUE #698: Instead of tool tips to explain concepts, link to holocat on chatbtc
For example, https://chat.bitcoinsearch.xyz/?author=holocat&question=How%2520is%2520bitcoin%27s%252021%2520million%2520supply%2520cap%2520enforced%253F

0tuedon:
This PR will close #698 
But for now, just created a holocat component to see how the inline works. I  will be taking Ben's Advice to separate the text on the translation files.


benalleng (commented):
Looks good, you need to remember to add the theme arg as it is currently transparent, but it works! Do you think it's possible to move the small arrow to be above the Holocat image like Christoph showed in the designs?

0tuedon:
> Looks good, you need to remember to add the theme arg as it is currently transparent, but it works! Do you think it's possible to move the small arrow to be above the Holocat image like Christoph showed in the designs?

trying to see if I can do this I am testing locally and removed the max-w-width but still lays in the text.

benalleng (commented):
I think tat moving the tooltip to be above the image vs just being above the link itself is not wirth it if we can't figure it out. Lets just go ahead with what we have, so remove any special styling that attempts to move the tooltip from above centered on the link. we can come back to it later if we want, but this looks great as is! lets go ahead and implement this in the places that we know it should be

ui/lesson/HolocatQuestion.tsx
benalleng, line 1: We probably should put this in the `common/` dir

benalleng (approved):
I am tentatively approving this as there is nothing here I consider a blocker anymore but we should continue to add the remining links like we have [here](https://saving-satoshi-git-fork-0tuedon-link-to-holocat-savingsatoshi.vercel.app/en/chapters/chapter-2/scripting-1?dev=true)

0tuedon:
Screenshot 2024-03-17 at 20 37 52
Minor issue here @benalleng  @daviroo

benalleng (approved):
I think that bug is unrelated, it looks like it happens any time the tooltip is squished against the side. I'm happy with this! I think it looks good as is!

GBKS:
One more thing to tweak would be to make sure the link in the tooltip itself opens in a new tab (currently opens in the same tab).

PR #880, opened 2024-03-05
Navigation Granularity

### LINKED ISSUE #872: Navbar tab and Challenge tab navigation should be more granular
We need to give the user more granular movement using the navbar tabs (in challenges) and the challenge tab (in chapters)

currently it only allows them to travel to the 1st instance of a named challenge

benalleng:
Closes #872


GBKS:
I think we need to iterate on this a little bit more. Some things to consider:

- The most obvious part is about coming up with unique titles for each page and section, so the are distinguishable and readers understand what each one is
- The numbering is now inconsistent with the smaller number-based index at the top-right of the screen on desktop. Maybe we reserve numbers for the larger blocks and pages don't have them?
- Text pages and problem pages look the same. Text pages can't really be completed (unless you count reading as completion). Should we visually differentiate text and challenge pages? Maybe text pages don't get a check on completion?

I'd like to take a stab at this from a design perspective, just need to find a little bit of time for it.

benalleng:
@GBKS I have made an update for the chapter 1 keys I added the titles as such

- As a note I made all the challenge pages have an imperative clause as their title 

image

GBKS:
Sorry for taking a million years, but I finally looked at this and gathered my thoughts and put together some design tweaks. Here's a [video overview](https://youtu.be/wHi0t3i1ST8), please let me know what you think.

GBKS:
I like this update, how about you?

Here's a comparison of the current implementation and the design.

image

The font sizes and spacing is a little tighter in the design (and icons also slightly smaller). I intentionally did it that way so we don't run into potential issues where the nav doesn't fit vertically on shorter screens and we end up needing to scroll (which does not play well with the hover activation and our other goal of keeping the UI at 100 window height). The best test is probably to find the chapter with the most pages and see if that still works. But would be nice to tighten things up if you get a chance.

benalleng:
Unfortunately chapter 5 is simply too long, I think I agree that click to open is better so I added that but I am not sure how tall I should make the maximum height tooltip? right now it is `calc(100vh - 70px(navbar) - 56px(successbar))` but this feels too big.

you can look here
https://saving-satoshi-git-fork-benalleng-navigati-184c07-savingsatoshi.vercel.app/en/chapters/chapter-5/derive-message-4?dev=true

GBKS:
I think a max-height of 600px or so is fine, so it doesn't become unwieldy. But, MacOS has very subtle scrollbars that are not very intrusive, so it might look ugly elsewhere. Alternative is of course to keep the individual sections (`1|2|3|4`) at the top and have a separate menu for each one.

Should still be possible to make it close when you click elsewhere.

benalleng:
After some additional feedback I decided to return to the numbered tabs, its a lot easier to find your place and not as busy with all lessons visible. Also I think easier to find a good stopping place

0tuedon:
Ready for review @benalleng ?

benalleng:
@0tuedon Yes! just finished the translations!

GBKS:
I think this works better, we avoid this whole scrolling issue.

The spacing is still a bit off compared to the design, would it be possible to tweak that? Here's a screenshot overlaid on the design (implementation spacing is much tighter, and text is centered instead of left-aligned).

image

0tuedon:
Hey @benalleng, minor addition. Could you add a cursor-disabled to this
Screenshot 2024-04-22 at 17 45 19

content/lessons/chapter-6/put-it-together-3.tsx
```diff
@@ -11,6 +11,7 @@ import { LessonInfo, ScriptingChallenge, Table, Text } from 'ui'
 
 export const metadata = {
   title: 'chapter_six.put_it_together_one.title',
+  navigation_title: 'chapter_six.put_it_together_three.nav_title',
   image: '/assets/images/chapter-4-intro-1.jpg',
```
0tuedon, line 15: <del>image: '/assets/images/chapter-4-intro-1.jpg',</del>

ui/chapter/Chapter.tsx
```diff
@@ -94,13 +94,13 @@ export default function Chapter({ children, metadata, lang }) {
     }
   }, [])
 
-  useEffect(() => {
+  /*useEffect(() => {
     !isAccountLoading &&
       !isLoading &&
       account &&
       progress.at(2) !== '1' &&
```
0tuedon, line 101: We don't need this anymore?

0tuedon (changes requested):
Looks good!, Left some comments

0tuedon (approved):
LGTM!

PR #881, opened 2024-03-06
feat: Now persists language selection

### LINKED ISSUE #857: Resource pages should default to the same language as the code editor the user in on
it would be convienient if the answers were defaulted to the same language the user is using.

This also brings up a higher level question of whether we should add some account level language preference

daviroo:
The change involves a lot of files but it's pretty much the same changes in each one:
- If the language is changed in a challenge then that changes the persisted language
- This language is carried over to future challenges and also to the help pages
- If the language is changed in the help page then it applies only to that code solution the user is looking (my reasoning is that they might look at the other languages solution out of curiosity and we don't want to persist that)

closes #857


ui/lesson/ScriptingChallenge/index.tsx
```diff
@@ -66,32 +72,34 @@ export default function ScriptingChallenge({
...
+  const [language, setLanguage] = useState(getLanguageString(currentLanguage))
+  console.log(language)
   const [challengeSuccess, setChallengeSuccess] = useState(false)
   const [hydrated, setHydrated] = useState(false)
   const [errors, setErrors] = useState<string[]>([])
   const debouncedCode = useDebounce(code, 500)
 
   useEffect(() => {
-    const savedCode = localStorage.getItem(lessonKey)
+    const savedCode = localStorage.getItem(`${lessonKey}-${language}`)
```
daviroo, line 94: @benalleng this is the fix for the saved code issue I mentioned.  Adding the language to the local storage key makes it unique for each challenges language selection.

benalleng (approved):
Looks good! In a future iteration do you think it would be possible to add a scrubber button in the terminal to clear out any unwanted saved code?

daviroo:
> Looks good! In a future iteration do you think it would be possible to add a scrubber button in the terminal to clear out any unwanted saved code?

Do you mean like a revert to what was the original code before the user began changing it?  That should be possible yeah since we have that available in the component.

PR #884, opened 2024-03-11
feat: Added help suggestion component

### LINKED ISSUE #855: Add some pointer to show that the user should check out the help page
Currently we have nothing that tells the user the help page exists, I think we could add a tooltip or pointer early on to show that the help page exists!

daviroo:
Closes #855 

image

@GBKS I think I've followed your design but would be good to get your eyes on this.   You can see an example of the component on [this preview link](https://saving-satoshi-git-fork-daviroo-highlight-5eeb17-savingsatoshi.vercel.app/en/chapters/chapter-4/address-2?dev=true).

### DIFF
--- a/components/HelpSuggestion.tsx
+++ b/components/HelpSuggestion.tsx
@@ -1,0 +1,18 @@
+import React from 'react'
+import { Text } from 'ui'
+
+const HelpSuggestion = () => {
+  return (
+    <div className="mt-4 flex items-center rounded-xl bg-black bg-opacity-20 p-4 bg-blend-darken">
+      <div className="mr-4 flex h-12 w-12 flex-shrink-0 items-center justify-center border border-white">
+        <Text className="text-lg md:text-xl">?</Text>
+      </div>
+      <Text className="md:text-md flex-grow text-sm">
+        If you&apos;re ever stuck, press the &ldquo;?&ldquo; button in the
+        top-right corner. It provides helpful resources and tips.
+      </Text>
+    </div>
+  )
+}
+
+export default HelpSuggestion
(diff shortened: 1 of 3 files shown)


GBKS:
Looks great. The only thing to weak would be to change the white border to 35% opacity.

Is this the location we want this tip to be in? Chapter 4, challenge 2? My thinking was that we put it in the first script/terminal challenge, on a screen that has otherwise little text (to increase the chance that people actually look it).

components/HelpSuggestion.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const HelpSuggestion = () => {
+  return (
+    <div className="mt-4 flex items-center rounded-xl bg-black bg-opacity-20 p-4 bg-blend-darken">
+      <div className="mr-4 flex h-12 w-12 flex-shrink-0 items-center justify-center border border-white">
+        <Text className="text-lg md:text-xl">?</Text>
+      </div>
+      <Text className="md:text-md flex-grow text-sm">
+        If you&apos;re ever stuck, press the &ldquo;?&ldquo; button in the
+        top-right corner. It provides helpful resources and tips.
```
benalleng, line 12: We should make this a trasnlation object

daviroo:
> Looks great. The only thing to weak would be to change the white border to 35% opacity.
> 
> Is this the location we want this tip to be in? Chapter 4, challenge 2? My thinking was that we put it in the first script/terminal challenge, on a screen that has otherwise little text (to increase the chance that people actually look it).

I put it in the generic `Info.tsx` component which we use in every coding challenge - I could change it to just be in one if you think that's better?

daviroo:
@GBKS @benalleng this slipped off my radar for a bit - ok to merge as it is?

benalleng:
Yeah I think it's good! If later we want to reduce how many places this is rendered we can but for now I think it's good!

benalleng (approved):

PR #886, opened 2024-03-12
fix python code comment regex

benalleng:
Fixes an error Jonas found whereby the code comment removal regex had some errors

Also fixes an error whereby the wrong default language was selected when loading in from an existing account without localStorage

### DIFF
--- a/ui/lesson/ScriptingChallenge/index.tsx
+++ b/ui/lesson/ScriptingChallenge/index.tsx
@@ -84,7 +84,6 @@ export default function ScriptingChallenge({
     config.languages[getLanguageString(currentLanguage)].hiddenRange
   )
   const [language, setLanguage] = useState(getLanguageString(currentLanguage))
-  console.log(language)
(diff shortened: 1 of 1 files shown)


content/lessons/chapter-2/scripting-2.tsx
```diff
@@ -125,7 +125,11 @@ const config: EditorConfig = {
 export default function Scripting2({ lang }) {
   const t = useTranslations(lang)
   const { currentLanguage } = useDataContext()
-  const [language, setLanguage] = useState(getLanguageString(currentLanguage))
+  const [language, setLanguage] = useState(
+    currentLanguage
+      ? getLanguageString(currentLanguage)
+      : config.defaultLanguage
+  )
```
benalleng, line 132: I ended up going the route of not immediately setting a default `currentLanguage` @daviroo do you think there is a more elegant way of solving this default to Javasscript issue?

benalleng:
@daviroo rebased

daviroo (approved):

PR #894, opened 2024-03-28
Scafold for chapter 8

0tuedon:


lib/progress.ts
```diff
@@ -88,6 +88,19 @@ export const keys = [
...
+  'CH8INT2',
+  'CH8INT3',
+  'CH8BBM1',
+  'CH8BBM2',
+  'CH8BBM3',
+  'CH8BBM4',
+  'CH8BBM5',
+  'CH8BBM6',
+  'CH8BBM7',
+  'CH8BBM8',
```
benalleng, line 102: should be "BBK"

content/chapters/chapter-8/index.tsx
```diff
@@ -22,8 +16,18 @@ export const metadata = {
...
+    'building-blocks-2',
+    'building-blocks-3',
+    'building-blocks-4',
+    'building-blocks-5',
+    'building-blocks-6',
+    'building-blocks-7',
+    'building-blocks-8',
+  ],
+  outros: ['outro-1'],
+  challenges: ['building-blocks-3'],
```
benalleng, line 32: the challenge here would still start at 'building-blocks-1' even though they are just text pages

content/lessons/chapter-8/building-blocks-2.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { getData } from 'api/data'
+import { Data } from 'types'
+import { chapters } from 'content/chapters'
+
+export const metadata = {
+  title: 'chapter_four.address_one.title',
+  key: 'CH8BBK2',
+}
+
+export default function BuildingBlocks1({ lang }) {
```
benalleng, line 17: lets correct the function names in 2-8 to increment up correctly

benalleng (changes requested):
3 small changes and then we should be good to merge!

benalleng (approved):
LGTM!

PR #903, opened 2024-04-04
Refactor chapter 6

benalleng:
Changes the chapter 6 challenges to be a little more concise

[Preview](https://saving-satoshi-git-fork-benalleng-chapter-377280-savingsatoshi.vercel.app/en/chapters/chapter-6/intro-1?dev=true)


benalleng:
Marking as ready for review even with the copy being wrong

benalleng:
the code collapsing on put-it-together-2 is the only thing that I feel doesn't work quite right

daviroo (changes requested):
From what we discussed in discord the only outstanding issues (not including copy) are to do with code folding in `put-it-together-2` for both javascript and python.

benalleng:
I added a comment that is on the same tab-level as the transaction class, therby fooling the Monaco editor into thinking that the comment is the parent. folding the transaction class that is actually the parent seperately

daviroo:
> I added a comment that is on the same tab-level as the transaction class, therby fooling the Monaco editor into thinking that the comment is the parent. folding the transaction class that is actually the parent seperately. This does not affect how the code is run

I'm not sure how that works but it does indeed work 😅 .  I think the PR is good to go once you have done whatever copy changes you had in mind.

daviroo (approved):

PR #905, opened 2024-04-10
chore: Added jotai state management for auth

daviroo:
I'm pulling out the changes I'm making to the state management to different branches/PRs.  It will make it a lot easier to review/test.  This one is focused on the authentication state/functions so best to test any challenges which use the account & login/logout functionality.


benalleng (approved):
tACK

- logged in and logged out (got some flashing when opening the logout modal)

- tested progress on chapter 1 and it works as expected.

- tested chapter 3 with the account image objects (used existing account storage)

- correctly hid chapter 8 behind feature flag

PR #907, opened 2024-04-11
chore: Now using jotai for modal state

daviroo:
Same as #905 but for modal state


benalleng (approved):
Nice catch on fixing the flashing!

PR #908, opened 2024-04-15
Add jotai state management feature context

daviroo:
Similar to #907 except for feature state.


benalleng:
Unfortunately I don't think this one works at the moment, since its using the prod backend we should be able to see chapters 1 + 2-7 and 8 should be hidden but 2-8 are all hidden right now

benalleng (approved):

PR #914, opened 2024-04-23
handle old keys

benalleng:
This is needed to prevent errors if the user is on `put-it-together-4/5/6`

These keys will not be hit by any user but likely that in Jonas' testing he may have progress stuck here


benalleng:
Closed, can't redirect properly

PR #917, opened 2024-04-24
Dynamic viewport height test

GBKS:
This is a test PR for switching thee CSS use of `100vh` to `100dvh`. The `dvh` is a [dynamic value](https://developer.mozilla.org/en-US/docs/Web/CSS/length#dynamic) that adjusts to the size of the browser chrome on mobile as you scroll, the keyboard expands/collapses/etc. This is used to calculate the screen height in certain places, like the code editor and mobile nav. With just `vh`, it can happen that elements go beyond the visible screen window and are blocked by browser chrome.

Reason for testing is that I found scrolling in the mobile nav a bit funky, some content at the bottom was not reachable. The nav shares a class to restrict this with the code editor, which I updated here. But, this may cause other problems, so the final solution might be for the mobile nav to have its own CSS instead of using the shared one. It's just a line or two anyways.

But let's test and see.

💼[Check the preview](https://saving-satoshi-git-feature-dynamic-viewpor-fae955-savingsatoshi.vercel.app/en/chapters/chapter-2/scripting-2?dev=true)🧳

### DIFF
--- a/app/[lang]/chapters/[slug]/[lesson]/page.tsx
+++ b/app/[lang]/chapters/[slug]/[lesson]/page.tsx
@@ -94,6 +94,10 @@ export default function Page({ params }) {
     return (
       <Portal id="head">
         <title>{`${title} - Saving Satoshi`}</title>
+        <meta
+          name="viewport"
+          content="width=device-width, initial-scale=1 maximum-scale=1"
+        />
       </Portal>
     )
   }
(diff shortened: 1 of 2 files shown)


GBKS:
Does anyone know what these pixel values stand for? I think `70px` is the top nav bar, but not sure about the others.

image

benalleng:
A lot of those values have to do with the monaco editor and its inability to easily be resized, the `40px` is the language selection and the `56px` is the bottom success bar

benalleng:
Ahh I see, i definitely accidentally included that class, it shouldn't be here

PR #922, opened 2024-04-25
Set maximum viewport scale

### LINKED ISSUE #918: Disable viewport zoom
When focusing the code editor on mobile, the screen currently zooms into that field (at least on mobile). This is an automatic browser behavior meant to help you focus on the edit task at hand. But in this case, it's just weird. Some of the UI is now cut off and it takes some messing around to get back to the full 100% view.

Solution would be to prevent this automatic zoom by updating the `viewport` tag to the following.

`<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">`

It currently is:

`<meta name="viewport" content="width=device-width, initial-scale=1">`

I could not figure out how to do this in Next, since it's not as easy as just putting that tag in the header (you end up with duplicate meta tags).

This would be a good one to get a test PR up and running to see if it fixes the behavior.

benalleng:
Closes #918 

the duplicate meta tag is there as we can't set metadata on `use-client` components so I added it in with the `<Portal/>` component

### DIFF
--- a/app/[lang]/chapters/[slug]/[lesson]/page.tsx
+++ b/app/[lang]/chapters/[slug]/[lesson]/page.tsx
@@ -94,6 +94,10 @@ export default function Page({ params }) {
     return (
       <Portal id="head">
         <title>{`${title} - Saving Satoshi`}</title>
+        <meta
+          name="viewport"
+          content="width=device-width, initial-scale=1 maximum-scale=1"
+        />
       </Portal>
     )
   }
--- a/config/metadata/index.ts
+++ b/config/metadata/index.ts
@@ -2,6 +2,11 @@ import openGraph from './opengraph'
 import twitter from './twitter'
 
 const metadata = {
+  viewport: {
+    width: 'device-width',
+    initialScale: 1,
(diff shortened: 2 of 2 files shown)


app/[lang]/chapters/[slug]/[lesson]/page.tsx
```diff
@@ -108,6 +108,10 @@ export default function Page({ params }) {
     return (
       <Portal id="head">
         <title>{`${title} - Saving Satoshi`}</title>
+        <meta
```
Comment (daviroo) [inline app/[lang]/chapters/[slug]/[lesson]/page.tsx:111]: There's a `Head.tsx` in the app which already has a viewport defined although I couldn't find it being imported/used in any application code - is this not being used and should it be perhaps?

benalleng:
Wanted to to a follow-up on these @daviroo where did you see that `Head.tsx` file? I can't seem to find it on the History and @GBKS does this PR fix your problem? I can't seem to reproduce on prod, probably my browser doesn't have that viewport zoom functionality

GBKS:
ACK. Just tested it on iPhone and this fixes the problem.

PR #923, opened 2024-04-30
chore: Moved progress state to jotai

daviroo:
Marked as draft as I want to do a bit more testing.  I started this one a couple of weeks ago so some of the context on what I did is a little fuzzy.  @benalleng would appreciate if you could give it a test as well.

### DIFF
--- a/api/local/getProgress.ts
+++ b/api/local/getProgress.ts
@@ -1,4 +1,4 @@
-import { defaultProgressContext } from 'contexts/ProgressContext'
+import { defaultProgressContext } from 'state/state'
 
 export default async function getProgressLocal(): Promise<string> {
   try {
--- a/api/progress/getProgress.ts
+++ b/api/progress/getProgress.ts
@@ -1,4 +1,4 @@
-import { defaultProgressContext } from 'contexts/ProgressContext'
+import { defaultProgressContext } from 'state/state'
 import { get } from 'utils'
 
 export default async function getProgress(): Promise<string> {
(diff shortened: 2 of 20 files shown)


daviroo:
@benalleng I've fixed the login issue - I've also tested the bug mentioned in https://github.com/saving-satoshi/saving-satoshi/issues/925 and it seems to be fixed in this branch.

benalleng:
while #925 is fixed I had the reverse problem in this branch where i completed chapter 1, then went through the outro page signup flow and after the saveAndReturn() flow it did not save my progress to the next chapter.

Additionally on a page refresh we seem to be encountering the issue where the default progress of `keys[0]` is being flashed in the state so I get booted back to intro-1 if I refresh from any page :cry:

benalleng:
I know this is just an update to the progress context so perhaps we can address these problems in a follow up PR that focuses sepecifically on these issues

daviroo:
@benalleng this could do with another test when you get the time.  I think I've fixed the highlighted issues.  I decided to not remove the local storage saving but instead opted to change the `saveProgress` method so that it selects the furthest progress key out of [key_from_state, key_from_local_storage, key_from_server].  This seems to have fixed the behaviour you saw.

benalleng (approved):
Works great! tested on a fresh browser going from `CH1INT1` to `Ch3INT1` and progress was smooth, even on page refresh. Also tested with an existing account on an incognito browser to check backend progress

PR #928, opened 2024-05-07
Navbar links use fully defined routes

### LINKED ISSUE #927: Resolve issue with navigation after entering into help page
**Describe the bug**
When you click the "help" section of any page/ challenge, using the navigation again doesn't work. 

**To Reproduce**
1. Go to any page/ challenge with a "help" section
2. Click on Help
3. Try to use the navigation again
4. ???
5. Borked

**Expected behavior**
Navigation should take you to where you click when you want to go to a place that is in the navigation button yes. 

**Screenshots and/or logs**
Vid: https://drive.google.com/file/d/1iNnjcCu_UF_2PZWXV76pSLbdg9hHQ2Go/view?usp=sharing

Desktop
- Device: Windows
- OS: 11
- Browser: Brave

benalleng:
Closes #927

### DIFF
--- a/components/Navbar/NavbarDesktop/TabDesktop.tsx
+++ b/components/Navbar/NavbarDesktop/TabDesktop.tsx
@@ -112,6 +112,7 @@ export default function Tab({
               const navTitle =
                 lessons[slug][navLessonId].metadata.navigation_title
               const ComponentType = isLessonUnlock ? Link : 'div'
+              console.log(challenge.lessonId)
               return (
                 <div
                   key={index}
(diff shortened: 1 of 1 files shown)


satsie (approved):
(Again, given my limited knowledge of React) code looks good to me. Tested the preview and I am now able to navigate from the help page.

I don't know what the right answer is, but since the help page launches in a new tab, fixing the navigation makes it tempting for the user to return to the challenge from the help tab. If you're a user that uses the help pages a lot, you could end up with a bunch of tabs open for the game, potentially creating issues with multiple user sessions at the same time. One fix could be to remove the nav bar altogether from the help pages, forcing the user to return to the original tab that launched the help page. Not sure how much effort is involved with that, and probably want a UX expert to weigh in here.

benalleng:
Yeah, we are in the beginning stages of doing a redesign of the help pages, I think ultimately we might get rid of them and opt for a modal on the page that holds some limited information about that challenge. 
For now though let's just make sure we don't have any 404's!

PR #929, opened 2024-05-07
tooltip no longer nested link

### LINKED ISSUE #920: Link within a Link in HolocatQuestion
We have a situation hwerby there is an `<a>` tag nested within another `<a>` tag in this component, but that is a no-no in html

benalleng:
Closes #920

### DIFF
--- a/ui/common/HolocatQuestion.tsx
+++ b/ui/common/HolocatQuestion.tsx
@@ -25,9 +25,8 @@ const HolocatQuestion = ({
       <Tooltip
         id={id}
         theme={theme || ''}
-        href={href}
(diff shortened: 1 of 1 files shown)


satsie (approved):
Can't say I know anything about React, but I did poke around the code when I saw this bug filed. The code changes look good to me and I tested them using the Vercel link. As expected, nothing has changed about the Holocat tool tip, specifically with the chapter 2, "The Warehouse" page that I was looking at: https://saving-satoshi-git-fork-benalleng-issue920-savingsatoshi.vercel.app/en/chapters/chapter-2/intro-1

PR #931, opened 2024-05-07
WIP More feedback!

benalleng:
Closes #933 

Addresses the following from #930

- [x] add `⌘+return` or `alt+return` to run scripting terminals
- [x] login flow
- [x] Holocat chatbtc question formatting
- [x] copy fixes
- [x] computing... copy while the runner has no result
- [x] outro page topbar opacity
- [x] replace generator point help link with chatbtc link
- [x] 404 page links fixes (happened in #928)
- [x] extra block explorer hint
- [x] fix some chatbtc link errors
- [x] help page scrolling
- [x] address bar copy changed from challenge number to challenge name
- [x] more hints for paying the change

doesn't fix
- responsive pages sizing
- user vs teammates avatar styling
-  holocat icons from chapter 3 are too similar to chatbtc questions
- chapters page image sizing with hidden navigation being really long
-  verify-signature-5 redesign still not there
- need chatbtc links on cerify-signature-3
- chapter 6 in-out-1 we're sending too much bitcoin for a flight
- chapter 6 in-out-4 too much copy for an easy-mode transition
- code highlights when referencing code in copy
- image quality (blurry or wonky)

Issue #170, opened 2023-02-21
Namespace data we store in Local Storage


afterburn:
**Describe the enhancement**
When storing data in Local Storage we should namespace our keys (see #161 for context). Right now we use the words `user` and `loggedIn` in. These have a very high probability of conflicting with the data from other projects that contributors are running on localhost:3000.

I suggest we prefix with `saving-satoshi-` so for example `saving-satoshi-user` and `saving-satoshi-loggedIn`, or something similar.

vinayaksh42:
I would like to work on this issue.

afterburn:
@vinayaksh42 all yours!

PR #936, opened 2024-05-09
Copy updates: chapters 1 - 5

satsie:
I have some copy updates for chapters 1-5. I've been sitting on them, but after the feedback that came in earlier this week I figure it's good to get these merged  so nobody is doing duplicate work. Most of these are straightforward but I have comments for  changes that have additional context.

It's possible I might have to rebase after https://github.com/saving-satoshi/saving-satoshi/pull/931 is merged.

### DIFF
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -963,7 +963,7 @@ const translations = {
       success_message_line_three: 'number of inputs:',
       success_message_line_four: 'hash of the tx that input #0 came from:',
       success_message_line_five:
-        'index of input #0 in the tx that created it:',
+        'index of input #0 in the funding transaction:',
       success_message_line_six:
         'scriptSig to authorize spending input #0:',
       success_message_line_seven: 'input #0 sequence:',
@@ -1365,12 +1365,12 @@ const translations = {
       list_three:
         'scriptcode is the raw Bitcoin script being evaluated. We also added this to our Input class back in step 2.',
       paragraph_three:
-        "We'll dive in to this more in the next section, but to spend from your pay-to-witness-public-key-hash address, your scriptcode would be:",
+        "We'll dive in to this more in the next section, but to spend from your pay-to-witness-public-key-hash (P2WPKH) address, your scriptcode would be:",
       paragraph_four: '...which decodes to the following Bitcoin script.',
       paragraph_five:
         'For more information about scriptcode see <Link href="https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki" target="_blank" className="underline">BIP 143</Link>.',
       paragraph_six:
-        'The raw transaction has preimage is the serialization of:',
+        'The raw transaction has a preimage that is the serialization of:',
       headings: {
         item_one: 'Description',
         item_two: 'Name',
@@ -1435,7 +1435,7 @@ const translations = {
           item_four: '4',
         },
         row_ten: {
-          item_one: 'Signature hassh type, we will use 1 to indicate "ALL',
+          item_one: 'Signature hash type, we will use 1 to indicate "ALL"',
           item_two: 'sighash',
           item_three: 'int',
           item_four: '4',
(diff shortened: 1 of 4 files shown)


i18n/locales/en.ts
```diff
@@ -217,7 +217,7 @@ const translations = {
       title: 'Outro',
       nav_title: 'Off to the address',
       paragraph_one:
-        'You run down to your garage, step into your dad’s old solocopter, punch in the coordinates, and make your way to the address.',
+        'You run down to your garage, step into your dad’s old Budgetcopter, punch in the coordinates, and make your way to the address.',
```
satsie, line 220: At some point between chapters 1 and 2 solocopter was changed to Budgetcopter . That might be intentional (maybe a Budgetcopter is a type of solocopter?) but I updated the copy and replaced "solocopter" with "Budgetcopter" just to keep things simple.

i18n/locales/en.ts
```diff
@@ -961,11 +961,11 @@ const translations = {
         'This is the raw transaction with each component labeled:',
       success_message_line_two: 'version:',
       success_message_line_three: 'number of inputs:',
-      success_message_line_four: 'hash of tx being spent by input #0:',
+      success_message_line_four: 'hash of the tx that input #0 came from:',
       success_message_line_five:
-        'index of output of tx being spent by input #0:',
+        'index of input #0 in the tx that created it:',
```
satsie, line 966: Still don't love this wording, but I think it's an improvement. Open to suggestions here.

i18n/locales/en.ts
```diff
@@ -1027,7 +1027,7 @@ const translations = {
       paragraph_two:
         'There are two 32-byte numbers we need to extract. They are referred to as R and S respectively and are each prefixed by the bytes 0220 in the DER sequence. Instead of fully decoding the DER blob, just look for the prefixes and paste the R and S values.',
       paragraph_three:
-        "# Satoshi's signature, from the input scriptSig of the tx to Hal Finney",
+        "# Satoshi's signature, from the input scriptSig of the tx to Hal Finney (block 170)",
```
satsie, line 1030: In general I think it's really helpful to refer to block 9 and block 170 whenever we can in this chapter because we're working with two transactions (the one from block 9 where Satoshi got the coins that were spent in block #170), and the output of the first transaction is the input of the second, meaning a specific UTXO can be considered both an output and an input and that's just all around confusing, even to the seasoned developer.

I plan to do another pass to identify more places where we can add clarification on if the text is referring to the transaction in block 9 or block 170

i18n/locales/en.ts
```diff
@@ -1182,13 +1182,11 @@ const translations = {
     intro_two: {
       nav_title: 'Paying Mike Ramen',
       paragraph_one:
-        '—HOLOCAT: “Hardly the only thing that’s made you sound stupid until now.”',
```
satsie, line 1185: This line is a duplicate. It appears on the previous screen.

i18n/locales/es.ts
```diff
@@ -260,7 +260,7 @@ const translations = {
       paragraph_one:
         'Para derivar una clave pública a partir de una clave privada, realizamos una operación de curva elíptica repetidamente con el punto generador (descubre por qué). El punto generador es un punto específico en la curva. Su valor es parte del estándar secp256k1 y siempre es el mismo:',
       paragraph_two:
-        'La operación de curva elíptica es similar a la adición y, por lo tanto, su repetición es similar a la multiplicación. Utilizamos el símbolo * para describir el algoritmo (aprende más), donde k es la clave privada y P es la clave pública correspondiente:The elliptic curve operation is similar to addition and therefore its repetition is similar to multiplication. We use the * symbol to describe the algorithm (learn more), where `k` is the private key and `P` is the corresponding public key:',
```
satsie, line 263: lucky for me it looks like this sentence was never translated anyway

i18n/locales/en.ts
benalleng: `index of input #0 from the funding tx` ?

benalleng (approved):

PR #939, opened 2024-05-09
Terminal Reset

### LINKED ISSUE #932: Create Terminal reset button
The coding terminal saves progress on a half-second debouce but there is no way to reset if a user makes a mistake soft locking themself at a lesson, we need a reset button adjacent to the language tabs to reset a coding challenge terminal back to its inital onMount state

benalleng:
Closes #932 

Adds a terminal reset that takes the state back to before the user interacted with it

image

One small issue that I think the effort to fix it isn't worth it is that code that is initially collapsed isn't collapsed on this refresh

[Preview](https://saving-satoshi-git-fork-benalleng-issue932-savingsatoshi.vercel.app/en/chapters/chapter-2/scripting-2?dev=true)

or to see that it also includes loaded code go to the below link and get to `put-it-together-2`

[Preview for chapter 6](https://saving-satoshi-git-fork-benalleng-issue932-savingsatoshi.vercel.app/en/chapters/chapter-6/in-out-4?dev=true)


daviroo:
This looks good and it does work but I'm wondering if we should add the text "Reset" or something so that they know this will clear their code.

benalleng:
went with a tooltip on hover

satsie (approved):
I think this looks good to me. I tested the reset button on the preview deployment and it worked as expected. The tooltip you added is present.

I actually haven't made it to the Chapter 6, Putting it together 2 challenge (that's actually the one I'm on, hah! It's hard :laughing:), so I couldn't check what you were talking about with regard to loaded code, but I tested the reset button on a few chapter 5 challenges which I believe also use loaded code and things look fine.

What happens if the user presses reset on a challenge they have already completed? Will they lose their solution? I hope not because I have found it very useful to go back to old challenges and reference code that worked. I know the solutions are available in the help tab, but it's ~~always~~ usually easier to read code you wrote than someone else's. I tried to look into this on the preview deployment but I'm not familiar with how code solutions are persisted, nor do I know what database/environment Vercel is pointing to.

benalleng:
Resetting the terminal does overwrite old data.

satsie:
Oh that is neat! I didn't realize we even had that feature where your solution can become pre-loaded code in a subsequent lesson. I agree with the way it's implemented in this PR: doing a reset on a previously completed lesson shouldn't be expected to update the subsequent lessons (if applicable)

PR #944, opened 2024-05-15
Chapter 6: Clarify "n" value in solution

satsie:
**Chapter 6, Putting it all together, Build the witness class**

One part of this exercise is implementing the following (last step for signing an input with your private key):

```
    // Extra Bitcoin rule from BIP 146
    // https://github.com/bitcoin/bips/blob/master/bip-0146.mediawiki#user-content-LOW_S
    //   s = -s mod n, if s > n / 2
```

In the solutions page, it is implemented as:

Screenshot from 2024-05-14 23-27-35

~~However line 45 should be if (s > (secp256k1.ORDER / 2)) (as opposed to 2n). n is not defined in the code, and if you look through the rest of the solution, n actually represents the order of the curve, secp256k1.ORDER.~~

This is confusing because n represents the order of the curve, but it's also being used here to coerce a number into a BigInt.

This PR replaces `2n` with `BigInt(2)` so the "n" character isn't used to represent two different things.

### DIFF
--- a/content/resources/chapter-6/put-it-together.tsx
+++ b/content/resources/chapter-6/put-it-together.tsx
@@ -168,7 +168,7 @@ const javascriptChallengeTwo = {
     const r = R.x.val % secp256k1.ORDER;
     let s = ((r * key) + BigInt(\`0x\${msg.toString('hex')}\`)) * k_inverted % secp256k1.ORDER;
 
-    if (s > (secp256k1.ORDER / 2n))
+    if (s > (secp256k1.ORDER / BigInt(2)))
       s = secp256k1.ORDER - s;
     return [r, s];
   }
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -1454,7 +1454,7 @@ const translations = {
       title: 'Putting it all together',
       nav_title: 'Build the witness class',
       heading: 'Sign and Populate the Witness!',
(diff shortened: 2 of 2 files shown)


daviroo:
> However line 45 should be `if (s > (secp256k1.ORDER / 2))` (as opposed to `2n`). `n` is not defined in the code, and if you look through the rest of the solution, `n` actually represents the order of the curve, `secp256k1.ORDER`.

The n is used to [coerce the number to a BigInt](https://stackoverflow.com/questions/59977095/what-does-character-n-after-numeric-literal-mean-in-javascript#:~:text=A%20BigInt%20is%20created%20by,of%20the%20least%20significant%20digits.).  Maybe it's not necessary in this case though?

satsie:
oh my @daviroo! you are absolutely right! I tried to run the code without it and it failed. 

I briefly reviewed the cryptography material that is linked in this challenge and I don't think renaming the "n" value in the equations that are in the code comments is a good idea. I updated this PR to add a comment in the solution code that clarifies this "n" is different from the one in the equation.

benalleng:
We could alternatively just use `BigInt(2)` as bigint math is relatively uncommon in js and I think a lot of people may not recognize the `2n` without that comment

satsie:
> We could alternatively just use BigInt(2) as bigint math is relatively uncommon in js and I think a lot of people may not recognize the 2n without that comment

I like this! I will make that change

benalleng (approved):

PR #948, opened 2024-05-16
Chapter 6: update solution for last challenge

satsie:
For the last challenge in chapter 6, "Sign the transaction", the player needs to update the change amount so miner fees are accounted for. I know we've [already added a comment](https://github.com/saving-satoshi/saving-satoshi/pull/931/files#diff-ea94042999c5e65bd33509c7502b7ab2a3ed6f8e8c24f93d2db08b56886e59c4R195) in the pre-loaded code that reminds the player to do this, but I think it also makes sense to call this out in the solution page as well. 

Does the JS text editor/code display frame dislike comments in solutions?
Screenshot from 2024-05-15 23-51-35

Python looks good
Screenshot from 2024-05-15 23-58-08

### DIFF
--- a/content/resources/chapter-6/put-it-together.tsx
+++ b/content/resources/chapter-6/put-it-together.tsx
@@ -291,7 +291,10 @@ const javascriptChallengeThree = {
     locktime.writeUInt32LE(this.locktime);
 
     return Buffer.concat([buf, locktime]);
-  }`,
+  }
+  
+  // Update the change amount to account for miner fees
+  const out1 = Output.from_options(addr, 60999000);`,
   validate: async (answer) => {
     return [true, undefined]
   },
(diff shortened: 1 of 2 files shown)


daviroo (approved):

daviroo:
> Does the JS text editor/code display frame dislike comments in solutions?
I think it's fine since we have [code](https://github.com/saving-satoshi/saving-satoshi/blob/93016dfb92c8b4a5541aa23d41c733d7497518d6/ui/lesson/ScriptingChallenge/index.tsx#L139) which trims out all the comments in the submitted answers.

benalleng:
One thing to make it readable is the Editor length on line 520, we need to increase it slightly so we don't need to scroll, the math is generally # of lines `(n x 19) + 10`

benalleng:
One thing to make it readable is the Editor length on line 520, we need to increase it slightly so we don't need to scroll, the math is generally # of lines `(n x 19) + 10`

PR #950, opened 2024-05-17
chore: Add jotai state data context

daviroo:


benalleng (approved):

benalleng:
Tested through chapter-6 works with both incoming data and code. Checked localstorage to confirm the code was being saved correctly

PR #955, opened 2024-05-29
Addressing Satsies feedback: ch. 4-6

pinheadmz:
https://docs.google.com/document/d/1ksMyK-PtV7x6xyLyECfcm3RuPP7NcrWMbIeweYmN7sw

### DIFF
--- a/content/lessons/chapter-6/in-out-4.tsx
+++ b/content/lessons/chapter-6/in-out-4.tsx
@@ -235,7 +235,7 @@ class Input:
           <Title>{t('chapter_six.in_out_four.heading_three')}</Title>
         </div>
         <Table headings={tableHeading} rows={inputRows} />
-        <Text className="mt-4 text-lg md:text-3xl">
+        <Text className="font-nunito text-2xl font-bold text-white">
           {t('chapter_six.in_out_four.paragraph_six')}
         </Text>
       </LessonInfo>
--- a/content/lessons/chapter-6/in-out-4.tsx
+++ b/content/lessons/chapter-6/in-out-4.tsx
@@ -1,9 +1,17 @@
 'use client'
 
+import { chapters } from 'content/chapters'
 import { useTranslations } from 'hooks'
 import { useState } from 'react'
 import { PlainEditorConfig } from 'types'
-import { LessonInfo, CodeExample, Title, Table, Text } from 'ui'
+import {
+  LessonInfo,
+  CodeExample,
+  Title,
+  Table,
+  Text,
+  HolocatQuestion,
+} from 'ui'
 import PlainEditorWrapper from 'ui/lesson/ScriptingChallenge/PlainEditorWrapper'
 
 export const metadata = {
@@ -158,6 +166,16 @@ class Input:
     setLanguage(language)
   }
 
+  const [tooltipVisible, setTooltipVisible] = useState(false)
+
+  const handleMouseEnter = () => {
+    setTooltipVisible(true)
+  }
+
+  const handleMouseLeave = () => {
+    setTooltipVisible(false)
+  }
+
   return (
     <PlainEditorWrapper
       fixedCode={''}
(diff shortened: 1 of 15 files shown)


i18n/locales/en.ts
```diff
@@ -1233,21 +1241,20 @@ const translations = {
...
-      paragraph_one: 'It should have the following method:',
+      heading: 'Looking at the Input class implementation',
+      paragraph_one:
+        'Bitcoin transaction inputs always point to existing, unspent transaction outputs. Therefore, our Input class has a method <span className="text-green"> from_output() </span> which is used to construct an Input by passing in the output description:',
       paragraph_two:
-        'The First two arguments are the transaction ID and the index of the output of that transaction you want to spend from.',
+        'The first two arguments are the transaction ID and the index of the output of that transaction you want to spend from. Eventually we will pass in the txid and vout values you got above from listunspent. ',
       paragraph_three:
-        'Eventually we will pass in the txid and vout values you got above from listunspent. Note that hashes in Bitcoin are little-endian, which means that you will need to reverse the byte order of the txid string!',
+        'Note that raw-byte hashes in Bitcoin are reversed, which is why we need to reverse the byte order of the hex-string txid argument that is passed in.',
```
satsie, line 1254:
```suggestion
        'Note that raw-byte hashes in Bitcoin are reversed (aka little-endian), which is why we need to reverse the byte order of the hex-string txid argument that is passed in.',
```

I'm so sorry, I missed that you had little-endian mentioned here. I think I got so deep into the code portion of the challenge that I forgot to revisit the challenge copy before filing some of my feedback. I really like the improvements you made here but I also agree with the original text that mentioned the term little-endian.

content/lessons/chapter-6/in-out-4.tsx
```diff
@@ -200,6 +200,9 @@ class Input:
           <Title>{t('chapter_six.in_out_four.heading_three')}</Title>
         </div>
         <Table headings={tableHeading} rows={inputRows} />
+        <Text className="mt-4 text-lg md:text-3xl">
```
satsie, line 203:
```suggestion
        <Text className="mt-4 font-nunito text-xl text-white">
```

Did you mean for this bit ("And remember...") to be big? It looks a little out of place next to the rest of the challenge copy:

Screenshot from 2024-05-29 23-13-21

satsie (approved):
MZ!!! You are a machine!!! Holy moly thanks for addressing so many things so quickly and efficiently. 

I left two suggestions, neither of which are showstoppers. Ping me if a re-approval is needed. ACK 434a30f1d2c9a628feff966bd3fbf8995c966578

i18n/locales/en.ts
pinheadmz: SOooooooo... technically "little endian" only accurately describes integers. I asked Pieter what the best term is to describe what we do to hashes and he shrugged out "byte swapped?".

One level deeper, goes like this: if you were to look at the actual output of a sha256 digest for a bitcoin block, you would actually see all the 0's on the *right side*. Satoshi decided to interpret that as a little-endian integer and that is why the int256 values of block hashes gets lower as difficulty increases: more zeros on the right side means a smaller little-endian integer.

Whenever the RPC was added to bitcoin, and for the first time block hashes were printed as strings for humans, someone decided to reverse the bytes of the hash so it looked like a human-readable integer, meaning more zeroes on the *left side* are smaller numbers.

Okay so that's the story as I understand it... how do we get that into the game?

i18n/locales/en.ts
```diff
@@ -813,6 +813,10 @@ const translations = {
         'The first step is to perform a SHA-256 hash on your compressed public key. Then perform a RIPEMD-160 hash on that SHA-256 output digest. The final result will be 20 bytes encoded as a hex string.',
       paragraph_three:
         'Complete a function that accepts a 33-byte compressed public key as a hex string and returns a 20-byte public key hash as a hex string.',
+      paragraph_four:
+        'Here is the documentation for the hashing libraries we imported for you:\n' +
+        '<Link href="https://nodejs.org/api/crypto.html#class-hash" target="_blank" className="underline">JavaScript: crypto</Link>\n' +
```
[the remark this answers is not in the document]
pinheadmz: Hm that's weird. Here is what I get, still not right on target but at least has an example of the `digest()` function

Screenshot 2024-05-30 at 9 52 30 AM

content/lessons/chapter-6/in-out-4.tsx
pinheadmz: Haha yes I intentionally blew that up to make it "loud" -- is that annoying? I can put it back to the default font size

i18n/locales/en.ts
satsie: Whoa!!!!! That is really cool! That's a fascinating bit of development history that we absolutely have to include in the game! This is going to require some design review but I'm thinking it could be some variation of the following ideas:

- A modal similar to the ChatBTC Holocat icon
- A piece of paper or dropped item that you discover in the game. The player will feel like they found something secret or hidden, but we'll intentionally put it somewhere easy to find :)
- A "History of Bitcoin"/Lore section in the help page

I would love for us to get this into the game sooner than later so that it's at least somewhere, then we can iterate on the best way to display it. I think we should add it to the help page for now. I am going to make note of this on my feedback doc.

As for the code suggestion that started this thread, you were right to remove "little-endian" :D Feel free to resolve this conversation.

content/lessons/chapter-6/in-out-4.tsx
satsie: Got it :D What about making it the same size/weight as the "Outpoint" and "Input", so `font-nunito text-2xl font-bold text-white`?

i18n/locales/en.ts
pinheadmz: Ok for now, I expanded on the reverse hash line in the text and added a Holocat tooltip to a longer explanation

content/lessons/chapter-6/in-out-4.tsx
pinheadmz: done, here and in the next page as well

satsie:
Updates look great! Merging

Issue #951, opened 2024-05-21
Chapter 7, "Assemble a Block" - Issues with block weight instructions


satsie:
Chapter 7 tasks the player with successfully filling a block with transactions. They face several challenges in doing this: 
- they can't go over the max block weight
- they need to take into account that some transactions have parents that also need to be mined, and 
- ideally, the solution optimizes for fees

The first thing that needs to be done is to add a check for the block weight because the code that is initially presented to the player has a loop that just puts every transaction from the mempool into the block. In the mempool [JSON file](https://raw.githubusercontent.com/saving-satoshi/challenges/master/chapter7/javascript/mempool.json) provided, there is a `weight` property on each transaction. The only place that units are mentioned is the table, which says the transaction weights are in virtual bytes (vBytes). If you match the values up, this indeed appears to be true:

![Screenshot from 2024-05-21 14-07-53](https://github.com/saving-satoshi/saving-satoshi/assets/1823216/8893c6b1-e44c-41ee-b499-1df35d2e95bd)

![Screenshot from 2024-05-21 14-11-01](https://github.com/saving-satoshi/saving-satoshi/assets/1823216/cc597fc9-d3da-4436-8a0f-391b38bc293f)

It follows that checking for max block weight should then be done terms of vBytes. The max block weight is:

- 4 million weight units (WU), or
- 1 million vBytes

Source: https://bitcoinvisuals.com/chain-block-weight and Holocat from ChatBTC ;)

![Screenshot from 2024-05-17 23-48-51](https://github.com/saving-satoshi/saving-satoshi/assets/1823216/21690a45-ae47-401c-9406-9d9967a68c28)

However, if you proceed to do this and set the max block size to 1 million in your code, your solution will never meet the threshold for even an "okay" answer because the solution checker is expecting a max block weight of [4 million](https://github.com/saving-satoshi/saving-satoshi/blob/3722b0fe21cca6ba29fc8e7dc90e873857909091/content/lessons/chapter-7/mempool-transaction-1.tsx#L99). Your answer ends up being off by about (probably a little less in practice) a factor of 4!

4 million WU makes more sense instead of 1 million vBytes because [WU is used for](https://learnmeabitcoin.com/technical/transaction/size/) adding transactions to a block. Many people are probably familiar with vBytes though because that's what is used for calculating fees. And honestly, I bet a lot of people don't even realize there are two ways to measure the weight of a transaction.

This leads me to several questions:

[...]

PR #956, opened 2024-05-30
Chapter 7: clarify consensus rules with holocat links

pinheadmz:
closes https://github.com/saving-satoshi/saving-satoshi/issues/951

One issue which I'll need @benalleng to help. This page has two tooltips, and they both share "visible" state:

Screenshot 2024-05-30 at 10 49 41 AM

### DIFF
--- a/content/lessons/chapter-7/mempool-transaction-1.tsx
+++ b/content/lessons/chapter-7/mempool-transaction-1.tsx
@@ -1,6 +1,13 @@
 'use client'
 
-import { ScriptingChallenge, LessonInfo, Title, Table } from 'ui'
+import { chapters } from 'content/chapters'
+import {
+  ScriptingChallenge,
+  LessonInfo,
+  Title,
+  Table,
+  HolocatQuestion,
+} from 'ui'
 import { EditorConfig } from 'types'
 import { useTranslations } from 'hooks'
 import { Text } from 'ui'
@@ -37,11 +44,15 @@ export default function MempoolTransaction1({ lang }) {
     </>,
     t('chapter_seven.mempool_transaction_one.headings.item_four'),
   ]
+  // txid
+  // fee
+  // weight
+  // [ancestors]
   const inputRows = [
     [
       'b27f86d3',
       '43430',
-      '220',
+      '2020',
       <span className="inline-block w-[72px]">&nbsp;</span>,
     ],
     [
@@ -84,12 +95,7 @@ export default function MempoolTransaction1({ lang }) {
       <span className="inline-block w-[72px]">&nbsp;</span>,
     ],
     ['88016f17', '15200', '1600', '5e518bbe'],
-    [
-      '8fa820d5',
-      '20221',
-      '1108',
-      <span className="inline-block w-[72px]">&nbsp;</span>,
-    ],
+    ['8fa820d5', '20221', '1108', <>38a62dcc</>],
   ]
 
   const [account] = useAtom(accountAtom)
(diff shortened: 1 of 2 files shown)


content/lessons/chapter-7/mempool-transaction-1.tsx
```diff
@@ -339,6 +345,15 @@ def run():
   const handleSelectLanguage = (language: string) => {
     setLanguage(language)
   }
+  const [tooltipVisible, setTooltipVisible] = useState(false)
```
benalleng, line 348: lets make them separate useStates

```suggestion
const [tooltipVisibleOne, setTooltipVisibleOne] = useState(false)
const [tooltipVisibleTwo, setTooltipVisibleTwo] = useState(false)
```

content/lessons/chapter-7/mempool-transaction-1.tsx
```diff
@@ -339,6 +345,15 @@ def run():
...
+
+  const handleMouseEnter = () => {
+    setTooltipVisible(true)
+  }
+
+  const handleMouseLeave = () => {
+    setTooltipVisible(false)
+  }
```
benalleng, line 356: and then these functions will also need to be indexed

benalleng:
```
  const handleMouseEnterOne = () => {
    setTooltipVisibleOne(true)
  }

  const handleMouseLeaveOne = () => {
    setTooltipVisibleOne(false)
  }
  
  const handleMouseEnterTwo = () => {
    setTooltipVisibleTwo(true)
  }

  const handleMouseLeaveTwo = () => {
    setTooltipVisibleTwo(false)
  }
  ```

**Comment by pinheadmz** [comment]
Thanks @benalleng updated

File: i18n/locales/en.ts [hunk 3]
```diff
@@ -1621,8 +1621,21 @@ const translations = {
...
+          href: 'https://chat.bitcoinsearch.xyz/?author=holocat&question=what%2520are%2520weight%2520units',
+          question: 'What are weight units?',
+        },
+        c: ' and the ',
+        d: {
+          text: 'order of transactions ',
+          href: 'https://chat.bitcoinsearch.xyz/?author=holocat&question=what%2520order%2520do%2520transactions%2520need%2520to%2520be%2520in%2520a%2520block',
+          question: 'What order do transactions need to be in a block?',
+        },
+        e: 'in the block.',
```
satsie, line 1637:
```suggestion
        e: ' in the block.',
```

nit

satsie (commented):
This is perfect! Before closing https://github.com/saving-satoshi/saving-satoshi/issues/951, where did we land with providing `const MAX_BLOCK_WEIGHT = 4000000;` for the user in the pre-loaded code (for context, see [Ben's comment](https://github.com/saving-satoshi/saving-satoshi/issues/951#issuecomment-2124991846))? We don't necessarily need to discuss this as part of this PR. I think it's less important now that we've made it clear the challenge uses weight units.

i18n/locales/en.ts
pinheadmz: fixed, thanks

pinheadmz:
> where did we land with providing `const MAX_BLOCK_WEIGHT = 4000000`

I think the holocat answer is sufficient. There's hints in the text about size and the first error you get is about size. It's chapter 7 i *think* we can expect players to figure this one out.

satsie (approved):

PR #959, opened 2024-06-01
Design: Darken the CodeExample element

satsie:
In chapter 6 we have some [really nice tables](https://savingsatoshi.com/en/chapters/chapter-6/in-out-5) where the background color is an opaque black. I thought it might be nice to update code examples to match this. What do we think? Are there any other areas of the project (other than challenges in other chapters) that use this CodeExample element that I should check on?

Screenshot from production:
Screenshot from 2024-05-31 21-51-38

After this PR:
Screenshot from 2024-05-31 21-51-30

### DIFF
--- a/ui/common/CodeExample.tsx
+++ b/ui/common/CodeExample.tsx
@@ -13,7 +13,7 @@ export default function CodeExample({
   className?: string
 }) {
   return (
-    <pre className={clsx('border-2 border-dashed border-white p-2', className)}>
+    <pre className={clsx('bg-[#00000033] p-2', className)}>
       <span
         className={`language-${language} flex items-center justify-between break-all pl-2 pr-0`}
       >


benalleng (approved):
I like it! It feels much more akin to the `Code Example` darkening as it is elsewhere on the internet

benalleng:
looks like you already caught the only other place I can think that the Code Example style is used in the help pages

benalleng:
what do you think about this page http://localhost:3000/en/chapters/chapter-5/derive-message-1 not a code example ui element but with the same styling points. do you think this should be distinct or should we move it over to being the same styling?

satsie:
> what do you think about this page http://localhost:3000/en/chapters/chapter-5/derive-message-1 not a code example ui element but with the same styling points. do you think this should be distinct or should we move it over to being the same styling?

That is a great question! I think my answer is yes... to both options :laughing: For Saving Satoshi in general, I prefer a darkened background over a dashed border. I think it's less noisy and goes better with the rest of the game (I like dashed borders too, but for other kinds of styles/themes).

For this particular screen (Vanderpoole's signed message http://localhost:3000/en/chapters/chapter-5/derive-message-1), how cool would it be to show it in your (fake) computer terminal, maybe with some color and ascii art? We'd want to start with a mock up before implementing, but even a simple green/black terminal (with vi commands at the bottom?) would be pretty sweet IMO. If this seems like a worthwhile idea I'm happy to make a draft that we can get Christoph's opinion on.

Screenshot from 2024-06-04 21-25-46

PR #960, opened 2024-06-01
Design: Resize HolocatQuestion avatar

satsie:
Don't hate me for this one but the little Holocat avatar that links to ChatBTC is slightly too large for the surrounding text. It's ever so slightly tall enough that it changes the leading for the line that it's on, creating a bit more space between only the lines above and below it.

In this PR I've made the avatar smaller, added a small amount of left and right margin, and set the `vertical-align` CSS property to to middle instead of bottom. There is still something that seems off to me, but I think this gets things in the right direction. I wish I could bump the avatar up just a smidge however I tried all the other `vertical-align` values and this one seems the most appropriate.

Before:

Screenshot from 2024-05-31 22-04-55

After:

Screenshot from 2024-05-31 22-56-33

### DIFF
--- a/ui/common/HolocatQuestion.tsx
+++ b/ui/common/HolocatQuestion.tsx
@@ -26,11 +26,11 @@ const HolocatQuestion = ({
         id={id}
         theme={theme || ''}
         content={question}
-        parentClassName="inline-flex align-bottom min-h-fit min-w-fit"
+        parentClassName="inline-flex align-middle min-h-fit min-w-fit mx-1"
         arrowPosition={arrowPosition}
         visibleOverride={visible}
       >
-        <Avatar avatar="/assets/avatars/holocat.jpg" size={32} />
+        <Avatar avatar="/assets/avatars/holocat.jpg" size={24} />
       </Tooltip>
     )) || (
       <a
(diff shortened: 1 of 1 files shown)


benalleng (approved):
Looks good, as far as alignment I wonder if items-center would do more what you are hoping for, since it has `inline-flex` class already

ui/common/HolocatQuestion.tsx
```diff
@@ -26,11 +26,11 @@ const HolocatQuestion = ({
         id={id}
         theme={theme || ''}
         content={question}
-        parentClassName="inline-flex align-bottom min-h-fit min-w-fit"
+        parentClassName="inline-flex align-middle min-h-fit min-w-fit mx-1"
```
benalleng, line 29: I think actually `align-sub` is the class we might want to use. It seems more inline than `align-middle`

benalleng:
```suggestion
        parentClassName="inline-flex align-sub min-h-fit min-w-fit mx-1"
```

satsie:
@benalleng that looks perfect!! I never would have guessed to try `align-sub` :raised_hands:

PR #962, opened 2024-06-03
README: add linting notes

satsie:
Add notes to the README about linting and the Husky hook. Also update `public/manifest.json` with the changes made after running `yarn prettier`. The Husky hook is not configured to look at `.json` files so this never had `pretttier` run against it.  It changed the direction of the slash marks from back slashes to forward slashes. When I ran the changes locally and briefly clicked around, everything seemed okay.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,17 @@ To run this project locally:
 6. Run `yarn start` or `yarn dev` to spin up the development environment
 7. Open `http://localhost:3000` in your browser to access the site
 
+### Linting
+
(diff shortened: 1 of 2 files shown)


benalleng (approved):
Nice! these json files were modified from an old contributor's vscode config, nice to see we've got the in-repo prettier config back in charge

satsie:
@benalleng just to be clear, I only documented what was already there :) (Husky and prettier) I did none of the heavy lifting, hah!

PR #952, opened 2024-05-23
Help page modal and resources update

benalleng:
This changes the resources page into a modal instead of a separate page.

Also makes the help page individual for each individual challenge you are on

### TODO
- [ ] modify the spoilers to be more like pseudo-code answers
- [x] narrow the focus of the copy for the help pages to be related to the page you are on only
- [ ] expand these changes through the rest of the game

PR #968, opened 2024-06-06
feat: Adapted state management to handle chapter difficulty

daviroo:
The progress state has been modified to use a json object which can now store a history of the users progress through the course.  This PR also consolidates a lot of the logic for handling course progress into a single file `progressState.tsx`.


benalleng:
One question after merging #952 i guess we should split easy mode hard mode into separate tips if necessary? Shouldn't be too hard, likely just an export name change to include the difficulty

benalleng:
additionally the `?dev=true` url param seems to be broken with the new progress map, how can I access chapter 6 using `?dev=true`

components/Navbar/NavbarDesktop/TabDesktop.tsx
```diff
@@ -87,9 +114,9 @@ export default function Tab({
       offset={0}
       theme={theme}
       className={clsx('cursor-default no-underline', {
-        'cursor-not-allowed': !isUnlocked,
+        'cursor-not-allowed': !isTabUnlocked,
       })}
-      disabled={!isUnlocked}
+      disabled={!isTabUnlocked}
```
benalleng, line 119: It seems like the "disabled" behavior for the top page selection tabs is inconsistent. I found it to not work correctly on the first few pages, CH1INT1-CH1GEN1

benalleng (commented):
Made a few blunders at the start without properly setting up my backend but its all looking promising!

benalleng (commented):
image

the individual lesson page isn't tracked as unlcoked at the very first page, my current progress should allow CH2INT1 to be unlocked

benalleng (commented):
After completing chapter 2 and pressing 'continue' on the chapter end page my focus was still on chapter 2 in the chapter page and not chapter 3

components/Navbar/NavbarDesktop/TabDesktop.tsx
benalleng: this seems consistent in all chapters for the first 3 pages in each chapter

benalleng (commented):
Get a hiccup on Chapter 6 put-it-together-1 once I complete `CH6PUT1-normal` it takes me straight to `CH6PUT3-normal`

benalleng (commented):
Logging in with an existing account to a clean browser doesn't seem to work properly, i takes me straight to CH1INT1

benalleng (changes requested):
We need to change this code to account for the new difficulty selection
```
// /normal/put-it-together-2.tsx
const getPrevLessonData = async () => {
    const data = await getData('CH6PUT1')
    if (data) {
      setPrevData({
        lesson_id: 'CH6PUT1',
        data: data?.code?.getDecoded(),
      })
    }
}
```

benalleng (changes requested):
getting a loading error on put-it-together-3.tsx with this code on line 55
`(detectLanguage(prevData['CH6PUT2']) === Language.JavaScript`
needs to be 
`(detectLanguage(prevData['CH6PUT2_NORMAL']) === Language.JavaScript`
also line 166
```
 prevData['CH6PUT2'] +
'\n'
```
to
```
 prevData['CH6PUT2_NORMAL'] +
'\n'
```

benalleng (commented):
Also when going to an already unlocked an completed lesson pressing continue does not progress to the next page, it goes all the way to the next tip of progress

hooks/useProceed.ts
```diff
@@ -1,27 +1,45 @@
...
+  const Proceed = () => {
+    let route
+    if (isDevelopment && nextLessonUsingCurrentRoute) {
+      route =
+        routes.chaptersUrl + nextLessonUsingCurrentRoute?.path + queryParams
+      markLessonAsComplete(currentLessonId)
+    } else {
+      route = routes.chaptersUrl + nextLessonPath + queryParams
+      progressToNextLesson()
+    }
```
benalleng, line 42: Haven't fully parsed the logic yet but I think we need to look at `nextLessonUsingCurrentRoute?.path` on both the if and else side of this conditional

hooks/useProceed.ts
```diff
@@ -1,27 +1,45 @@
...
-    const route = routes.chaptersUrl + nextLessonPath + queryParams
-
+  const Proceed = () => {
+    let route
+    if (isDevelopment && nextLessonUsingCurrentRoute) {
+      route =
+        routes.chaptersUrl + nextLessonUsingCurrentRoute?.path + queryParams
+      markLessonAsComplete(currentLessonId)
+    } else {
+      route = routes.chaptersUrl + nextLessonPath + queryParams
```
benalleng, line 40: Yeah, it seems like changing this to 
`route = routes.chaptersUrl + nextLessonUsingCurrentRoute?.path + queryParams`
fixes it but then the conditional doesn't seem to have a reason to exist

benalleng (approved):
tACK! great change! the upcoming PR will add the ui change to use this refactor

PR #981, opened 2024-06-11
Chapter 8 Updates

satsie:
I just finished playing chapter 8 and this PR contains a bunch of updates I have for it. It's best to review this PR by looking at each commit in sequential order.

**Commit 1** - copy changes and a few style updates similar to previous PRs I've made:

Screenshot from 2024-06-10 21-22-38

**Commit 2** - [building-blocks-5] Fix hashes in the challenge instructions. Prior to this change, the instructions did not match what is in
the pre-loaded code, and what the validation code is looking for.

**Commit 3** - [building-blocks-5] Add a default return value to the preloaded code. The user eventually has to change this, but it at least fixes an exception that prevented the user from debugging with `console.log()` statements.

**Commit 4** - [building-blocks-6] Add another custom (JavaScript) error message, this time for an incorrect solution that allows the subsidy to be a fraction of a satoshi. Can't remember how I ended up in this scenario but I think whatever I did, I missed the custom error message that the answer must be of type number. Also a few minor copy updates to this file.

**Commit 5** - [building-blocks-7] This challenge is a little buggy. It asks you to make a `validateBlock(block)` function that, when used on a a list of blocks, prints the hash of the one valid block. It's unusual that this method, not whatever calls it, is responsible for printing the hash of a valid block. However, to change that would require a larger rework. For now, I want to point out that our validation code checks for a specific hash (`88fd124d747cde1d8494d589ec6b82ce11356dd869823dfec8e84b111a72bc87`) that has been hard coded into `building-blocks-7.tsx` (`validate` method on the `javascript` and `python` objects).

We tell the user that the validation code will loop through a list of blocks and execute the following if statement on each block. If the block is valid, the hash will be returned:
```
  if(validateBlock(block)){
    return bhash;
  }
```
That sample code we provide assumes `validateBlock()` returns a boolean, but the only requirement given to the user is that `validateBlock()` prints the hash of the valid block. I propose we update this check to explicitly look for a `true` boolean, because if `validateBlock()` returns a string like "peanuts", it would still be truth-y and pass the check. 

Screenshot from 2024-06-10 22-55-05

That addresses the sample validation code provided in the instructions :) The actual validation code is slightly different:

```
  if(validateBlock(block)){
    console.log(bhash)
  }
```

In this, all `validateBlock()` has to do is return something truth-y and the block hash is printed. That "peanuts" example I gave before will pass the `validateBlock()` check and print the block hash. The validation code will loop over every block, all  the `validateBlock()` calls will evaluate to true, it will print them all, and the part that compares what is printed  with the hash of the valid block (`88fd124d747cde1d8494d589ec6b82ce11356dd869823dfec8e84b111a72bc87`) will find what it's looking for and the user will pass the challenge.

In the instructions we tell the user to print the hash of the valid block, so before leaving the `validateBlock()` method, if the block is valid, the user should print the hash. That means our actual validation code doesn't need to do any printing, it just needs to execute `validateBlock(block)`.

In a follow up PR, I think it would be worthwhile to update this challenge so `validateBlock()` just needs to return a boolean, and the code that checks the answer can make sure that `validateBlock(88fd124d747cde1d8494d589ec6b82ce11356dd869823dfec8e84b111a72bc87)` returns true, while false for the other block candidate hashes.

### DIFF
--- a/content/lessons/chapter-8/building-blocks-7.tsx
+++ b/content/lessons/chapter-8/building-blocks-7.tsx
@@ -59,18 +59,24 @@ export default function BuildingBlocks7({ lang }) {
 
   const javascript = {
     program: `//BEGIN VALIDATION BLOCK
-let HEIGHT = 6929851
-let candidates = Bitcoin.rpc("getblocksbyheight", HEIGHT)
-for (const bhash of candidates){
+    const fs = require('fs');
+    const state = JSON.parse(fs.readFileSync('chainstate.json', 'utf8'));
 
- let block = Bitcoin.rpc("getblock", bhash)
+    const HEIGHT = 6929851;
+    let candidates = Bitcoin.rpc("getblocksbyheight", HEIGHT);
 
-  if(validateBlock(block)){
-    console.log(bhash)
-  }
+    for (const bhash of candidates){
+      let block = Bitcoin.rpc("getblock", bhash);
 
-}
-console.log("KILL")`,
+      if (state.blocks_by_hash[bhash].valid === false && validateBlock(block) == true) {
+        console.log('Looks like some invalid blocks are getting through. Try again!');
+        console.log('KILL');
+      }
+      if (state.blocks_by_hash[bhash].valid === true && validateBlock(block) === true) {
+        console.log(bhash);
+      }
+    }
+    console.log("KILL");`,
     rangeToNotCollapse: [
       {
         start: countLines(cleanedCombinedCode) + 5,
@@ -108,12 +114,19 @@ function validateBlock(block) {
 
   const python = {
     program: `# BEGIN VALIDATION BLOCK
+from json import load
+with open(f"{'chainstate.json'}", "r") as file:
+  state = load(file)
+
 HEIGHT = 6929851
 candidates = Bitcoin.rpc("getblocksbyheight", HEIGHT)
 
 for bhash in candidates:
   block = Bitcoin.rpc("getblock", bhash)
-  if validate_block(block):
+  if no state["blocks_by_hash"][bhash]["valid"] and validate_block(block):
+    print("Looks like some invalid blocks are getting through. Try again!")
+    print("KILL")
+  if state["blocks_by_hash"][bhash]["valid"] and validate_block(block):
       print(bhash)
 print("KILL")`,
     defaultFunction: {
@@ -173,14 +186,14 @@ for bhash in candidates:
     if validate_block(block):
         print(bhash)
         break`
-  const jsValidation = `const HEIGHT = 6929851
-const candidates = Bitcoin.rpc("getblocksbyheight", HEIGHT)
+  const jsValidation = `const HEIGHT = 6929851;
+const candidates = Bitcoin.rpc("getblocksbyheight", HEIGHT);
 for (const bhash of candidates) {
-let block = Bitcoin.rpc("getblock", bhash)
-if(validateBlock(block)){
-  return bhash;
-    }
-  }`
+  let block = Bitcoin.rpc("getblock", bhash);
+  if (validateBlock(block) == true) {
+    return bhash;
+  }
+}`
   const codeValidation =
     detectLanguage(combinedCode) === Language.JavaScript
       ? jsValidation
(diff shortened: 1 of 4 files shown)


i18n/locales/en.ts
```diff
@@ -1809,9 +1807,9 @@ const translations = {
...
+      heading_one: 'SHOWTIME!!',
+      paragraph_one: `The cameras are rolling, there's two billion humans world-wide tuned in to the live stream, and there's only a few minutes left before the next commercial break. Deborah Chunk is sweating, Holocat is sweating. Somewhere on the other end of the call-in phone line, Vanderpoole must be sweating. This is it!`,
+      paragraph_two: `Starting with the valid block just before the one you found at height 6929851, find the longest chain of valid blocks you can. Store the chain as an array of block hashes. While you're at it, maintain an array of every invalid block you find as well, just to show the world how hard Vanderpoole tried to break Bitcoin. It doesn't matter what order these invalid block hashes are in, but your valid chain MUST start with the hash of block 6929850 followed by one block hash at each height all the way up to the chain tip.`,
```
satsie, line 1812: Updated the copy here because this challenge expects the answer to include the hash of block 6929850

content/lessons/chapter-8/building-blocks-7.tsx
```diff
@@ -176,11 +174,11 @@ for bhash in candidates:
   const jsValidation = `const HEIGHT = 6929851
 const candidates = Bitcoin.rpc("getblocksbyheight", HEIGHT)
 for (const bhash of candidates) {
-let block = Bitcoin.rpc("getblock", bhash)
-if(validateBlock(block)){
-  return bhash;
-    }
-  }`
+  let block = Bitcoin.rpc("getblock", bhash)
+  if(validateBlock(block) == true){
```
benalleng, line 178: I think we can improve the validation even further here. We have access to `chainstate.json` in the backend which has the validity of each block in it already

benalleng: I haven't tested it yet but I think we will want something akin to this.
```
    program: `//BEGIN VALIDATION BLOCK
const fs = require('fs');
let HEIGHT = 6929851
let candidates = Bitcoin.rpc("getblocksbyheight", HEIGHT)

const state = JSON.parse(fs.readFileSync('chainstate.json', 'utf8'));

const isValid = JSON.stringify(state.valid) === JSON.stringify(answer.valid)
for (const bhash of candidates){
  let block = Bitcoin.rpc("getblock", bhash)

  if(validateBlock(block) == true && state.block.valid){
    console.log(bhash)
  }
}
console.log("KILL")`,
```

satsie: I just tested this and if I set my solution for `validateBlock()` to just be `return true;`, I pass the challenge. The valid block check is good, but it's not checking the invalid blocks.

I'm going to change the challenge so the pre-loaded code includes the loop that tries all the block candidates and prints the hash of the valid one. Then we can just check that the user printed the correct block hash.

satsie: On second thought, I think I should do that in a follow up PR. This challenge is currently a little buggy on prod and I'd prefer to push out a fix as soon as possible. What are your thoughts?

satsie: I just rebased that last commit, please see eed9fa67fcb428843b544fc4b76e89912d6169bb for an update

In that commit I updated the validation code to make sure `validateBlock()` returns false for invalid blocks, and added a custom error message for when invalid blocks are getting through. Also updated the challenge instructions to tell the user to return a boolean. Now that the validation code is updated, the player no longer needs to print the hash of the valid block and the spoilers page does not need any updates.

benalleng (approved):
Nice work on the updated validation!

benalleng (commented):
For consistency I think we should probably add a `return 0` in the python code here https://github.com/saving-satoshi/saving-satoshi/pull/981/files#diff-fb3713567b73d7165b89ffc49e454db0c632c37a107618eebfa9643575b8581eL82

satsie:
Thanks for taking a look at that! It was one of those things that every time I thought I had the right solution, I noticed something that could have been done better :sob: 

I just rebased to fix some conflicts and added that default `return 0` statement to the matching Python challenge. I'm not sure if Python coders would run into the same problem I did in JS (I could not log anything until I made the method return something), but I like that the preloaded code for both languages is now consistent.

Issue #926, opened 2024-05-07
Experiment with "list of resources" page


aassoiants:
Some people said they'd like a list of additional readings & resources covered in each chapter. 

This issue is a nudge to prototype what one of these can look like.

PR #983, opened 2024-06-11
chapter 4 help pages update

### LINKED ISSUE #971: Chapter 4 help pages
This chapter is where we should definitely remove all of the actual answers and transition completely over to psuedo-code solutions forcing the user to get there themselves

benalleng:
Closes #971 

[Preview](https://saving-satoshi-git-fork-benalleng-chapter-4-help-savingsatoshi.vercel.app/en/chapters/chapter-4/intro-1?dev=true)


i18n/locales/en.ts
```diff
@@ -855,22 +857,7 @@ const translations = {
         'You withdraw the bitcoin from the mining pool into the wallet you just created. You are now fully funded and ready for whatever Vanderpoole and BitRey may throw at you next.',
     },
     resources: {
-      public_key: {
-        elliptic_curve_reason_heading:
-          'The reason for elliptic curve operations',
```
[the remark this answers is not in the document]
benalleng: I'm considering moving more generalized resources like this to the end of the chapter with #926.  There we can add the "why" and also we have considered that as a good place to link directly to the secp256k1 library the core uses to show users how it is used in production

content/lessons/chapter-4/public-key-4.tsx
```diff
@@ -57,13 +57,18 @@ export default function PublicKey4({ lang }) {
...
+console.log(compressPublicKey(uncompressedKey))
 console.log("KILL")`,
     defaultFunction: {
       name: 'compressPublicKey',
       args: ['publicKey'],
     },
-    defaultCode: `${prevData?.data && `const uncompressedKey = "${dataObject}"`}
+    defaultCode: `${
+      prevData?.data &&
+      `const uncompressedKey = ["${dataObject
```
daviroo, line 68: Should this be an array?   In the previous code this followed a similar pattern to the python `uncompressed_key` object in line 110.

daviroo (changes requested):

content/lessons/chapter-4/public-key-4.tsx
satsie: I think the change to array is intentional. Here's some context from the ticket: https://github.com/saving-satoshi/saving-satoshi/issues/971#issuecomment-2161230315

i18n/locales/en.ts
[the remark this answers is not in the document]
satsie: I love that idea!

content/resources/chapter-4/address-2.tsx
```diff
@@ -22,11 +22,11 @@ const javascript = {
   },
   defaultCode: `function hashCompressed(compressedPublicKey) {
   // First we need to hash the compressedPublicKey with SHA256
-  const sha256Hash = crypto.createHash('sha256').update(Buffer.from(compressedPublicKey, 'hex')).digest()
```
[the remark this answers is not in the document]
satsie: https://github.com/saving-satoshi/challenges/pull/11

content/lessons/chapter-4/public-key-4.tsx
benalleng: I changed it to use an array instead of a string with a comma because it required the user to do some rather unusual transformation I found not relevant to this lesson

benalleng: see this code
```
  const regex = /([a-f0-9]+),([a-f0-9]+)/;
  const cleanedPublicKey = publicKey.replace(/\\s/g, '');
  const match = cleanedPublicKey.match(regex);
  ```

satsie (approved):

PR #986, opened 2024-06-12
Fix scripting challenge terminal height

### LINKED ISSUE #985: Terminal height shifted
The scripting challenge terminal height seems to have been shifted slightly by something. The monaco editor is rather finicky when it comes to height so its possible something very minor caused this

image

benalleng:
Closes #985

### DIFF
--- a/styles/globals.css
+++ b/styles/globals.css
@@ -210,10 +210,10 @@ div.react-terminal-active-input {
 }
 
 .code-editor {
-  height: calc(100vh - 70px - 48px);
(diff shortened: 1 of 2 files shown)


0tuedon:
is it possible to put a before and after can't really tell the difference

benalleng:
@0tuedon 
Original

image


This PR

image

note the overlap in the original

0tuedon (approved):
Great Work!, looks good.

PR #987, opened 2024-06-12
Terminal tooltip now not clickable

benalleng:
This modifies the reset terminal tooltip to not be clickable anymore, I found that it caused some weird behavior before that may cause a user to accidentally click the reset button

Only the icon is clickable now


0tuedon (approved):
LGTM!

PR #988, opened 2024-06-13
Chapter name and more missing assets

benalleng:
fixes some minor bugs


i18n/locales/en.ts
```diff
@@ -1210,7 +1210,7 @@ const translations = {
   },
 
   chapter_six: {
-    title: `Offense is the best defense`,
+    title: `The Keyholder`,
```
satsie, line 1213: Just an FYI that on prod, chapter 5 is also called "The Keyholder"

benalleng: doh!

PR #989, opened 2024-06-13
1.8.2

benalleng:
Adds a few more hotfixes and notably some changes to clean up the challenges in chapter 8

- terminal reset button now only clickable on the icon
- swap overflow-scroll for overflow-auto
- fix terminal height after some regression
- chapter 8 fixes
- chapter 2 help pages update
- small copy and asset fix


0tuedon (approved):
Tested, works as expected

PR #990, opened 2024-06-15
chore: Fix image caching config on vercel

daviroo:
I found a git issue where people had a similar problem to us and this was the suggested solution.

https://github.com/vercel/next.js/discussions/19146

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -1,0 +1,14 @@
+{
+    "headers": [
+      {
+        "source": "/_next/image(.*)",
+        "headers": [
+          {
+            "key": "Cache-Control",
+            "value": "public, max-age=31536000, immutable"
+          }
+        ]
+      }
+    ]
+  }
+  


daviroo:
> **The latest updates on your projects**. Learn more about [Vercel for Git ↗︎](https://vercel.link/github-learn-more)
> 
> Name	Status	Preview	Comments	Updated (UTC)
> **saving-satoshi**	✅ Ready ([Inspect](https://vercel.com/savingsatoshi/saving-satoshi/FVR2S7jUZn8ehahgVbWkoo4i4Nzg))	[Visit Preview](https://vercel.live/open-feedback/saving-satoshi-git-fork-daviroo-fix-image-caching-savingsatoshi.vercel.app?via=pr-comment-visit-preview-link&passThrough=1)	💬 [**Add feedback**](https://vercel.live/open-feedback/saving-satoshi-git-fork-daviroo-fix-image-caching-savingsatoshi.vercel.app?via=pr-comment-feedback-link)	Jun 15, 2024 3:29pm

@benalleng I wonder if you loaded a page with one of those images in the preview then looked at that page in vercel we looked at on Friday if it would now show the _next images?

benalleng:
Still the same things being cached, no new images it seems 
image

benalleng:
I actually think that we were looking at the wrong thing now. I think we were looking at networking, not data cache

benalleng:
I found this site, https://photoshot.app/prompts

They use the tag `unoptimized` in their `<Image>` components here https://github.com/baptadn/photoshot/blob/main/src/components/home/TiltImage.tsx in addition to the direct path to their images which seems to allow them to bypass `_next/image`

daviroo:
@benalleng do we still want/need this?

benalleng:
I don't think so, I found the behavior to be sporadic with my new computer. I think we should just leave an issue up that it may be a problem on really slow Internet/ computers

PR #1005, opened 2024-06-23
Chapter 4, address-3: add prefix table, provide more instructions to user

satsie:
Chapter 4, address-3: add human readable prefixes to a table, provide more detailed instructions to the user

Before:

Screenshot from 2024-06-23 01-05-36

After:

Screenshot from 2024-06-23 01-06-08

### DIFF
--- a/ui/common/Table.tsx
+++ b/ui/common/Table.tsx
@@ -18,8 +18,12 @@ export default function Table({
               key={index}
               className={clsx({
                 'w-full': index === 0,
-                'w-1/3': index > 0 && index < headings.length - 1,
-                'w-1/6': index === headings.length - 1,
+                'w-1/2': headings.length === 2, // only two columns in the table
+                'w-1/3':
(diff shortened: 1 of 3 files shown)


ui/common/Table.tsx
```diff
@@ -18,8 +18,12 @@ export default function Table({
               key={index}
               className={clsx({
                 'w-full': index === 0,
-                'w-1/3': index > 0 && index < headings.length - 1,
-                'w-1/6': index === headings.length - 1,
+                'w-1/2': headings.length === 2, // only two columns in the table
```
satsie, line 21: I added a check for the number of columns so that when there are only 2 columns, they each take up 50% of the table width. Not doing this would force this table to adopt the column widths of the table in the chapter 7 mempool-transaction-1 exercise, which has 4 columns and those settings look weird on this new chapter 4 table.

benalleng (approved):

PR #1008, opened 2024-06-25
Chapter 7 resources update

### LINKED ISSUE #974: Chapter 7 help pages
Since this will only add one page to edit I think we should have some more reading resources than we do in 4/5/6. Also again replace the answer with a psuedo-code solution

benalleng:
Closes #974

[Preview](https://saving-satoshi-git-fork-benalleng-issue974-savingsatoshi.vercel.app/en/chapters/chapter-7/mempool-transaction-1?dev=true)

### DIFF
--- a/content/resources/chapter-7/mempool-transaction-1.tsx
+++ b/content/resources/chapter-7/mempool-transaction-1.tsx
@@ -24,33 +24,41 @@ const javascript = {
     `function assembleBlock(mempool) {
   // This block constructor opportunistically includes transactions with unconfirmed parents.
   // If a transaction has unconfirmed parents, but those parents are all already included in the block, then it is valid for inclusion.
-  const block = [];
-  let block_weight = 0;
-  for (const tx of mempool) {
-      tx.feerate = parseFloat(tx.fee) / parseFloat(tx.weight);
-  }
-  // Construct dictionary for fast lookup
-  const txs = new Map([...mempool.map(tx => [tx.txid, tx])].sort((a, b) => b[1].feerate - a[1].feerate));
+
+  initialize empty list block
+  initialize block_weight to 0
+
+  // Calculate the fee rate for each transaction
+  for each transaction in mempool:
+    transaction.feerate = transaction.fee / transaction.weight
+
+  // Create a dictionary for fast lookup and sort transactions by fee rate in descending order
+  initialize dictionary txs with transactions from mempool sorted by feerate in descending order
+
   while (true) {
-    let added = false;
-    for (const tx of txs.values()) {
-      // Opportunistically include txs with unconfirmed parents if their parents
-      if (tx.parents.some(parent_tx => txs.has(parent_tx))) {
+    initialize added to false
+
+    for (each transaction in txs) {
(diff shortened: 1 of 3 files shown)


i18n/locales/en.ts
```diff
@@ -1711,7 +1717,7 @@ const translations = {
         item_two: 'Fee',
         item_two_b: 'Satoshis',
         item_three: 'Weight',
-        item_three_b: 'Weight units (WU)',
+        item_three_b: 'Weighted units (WU)',
```
pinheadmz, line 1720: it is actually "weight units" ... did you find a source that calls them "weighted" ?

content/resources/chapter-7/mempool-transaction-1.tsx
```diff
@@ -24,33 +24,41 @@ const javascript = {
...
   // This block constructor opportunistically includes transactions with unconfirmed parents.
   // If a transaction has unconfirmed parents, but those parents are all already included in the block, then it is valid for inclusion.
-  const block = [];
-  let block_weight = 0;
-  for (const tx of mempool) {
-      tx.feerate = parseFloat(tx.fee) / parseFloat(tx.weight);
-  }
-  // Construct dictionary for fast lookup
-  const txs = new Map([...mempool.map(tx => [tx.txid, tx])].sort((a, b) => b[1].feerate - a[1].feerate));
+
```
pinheadmz, line 27: Seems overkill to have both JS and Python versions of pseudocode, which is really a third language. The editor colors/highlighting makes it look a bit weird as well. Could it just be a monospace text field?

i18n/locales/en.ts
```diff
@@ -1729,7 +1735,21 @@ const translations = {
...
+      mempool_transaction_one: {
+        bytes_v_weight_heading: 'Bytes vs Weighted Units',
+        bytes_v_weight_paragraph_one:
+          'In Bitcoin block construction, "weighted units" (WU) and "bytes" are two different metrics used to measure the size of transactions and blocks.',
+        bytes_v_weight_subheading_bytes: 'Bytes',
+        bytes_v_weight_paragraph_two:
+          'Bytes refer to the raw size of a transaction or block in terms of actual data. It is the literal size of the transaction data when it is serialized (converted into a format that can be stored or transmitted). Before the implementation of Segregated Witness (SegWit), the block size was limited to 1 megabyte (1 MB), measured in bytes.',
+        bytes_v_weight_subheading_wu: 'Weighted Units',
+        bytes_v_weight_paragraph_three:
+          'Weighted units are a more complex metric introduced with SegWit in Bitcoin Improvement Proposal 141 (BIP 141). This system aims to more fairly allocate block space by considering the impact of witness data separately.',
```
pinheadmz, line 1748: recommend link to https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#user-content-Other_consensus_critical_limits

pinheadmz: Might also mention the new segwit limit is 4000000 weight units since you mentioned the legacy 1 MB limit in the last paragraph.

pinheadmz (approved):
ACK, Update to "weight" units was only real blocker for me.

PR #1009, opened 2024-06-25
WIP: OPCODE RUNNER

0tuedon:
## Pull Request checklist

- Clean commit history


ui/lesson/OpCodeParser/index.ts
```diff
@@ -0,0 +1,440 @@
...
+  if (stack[0] === false || parseInt(stack[0]) === 0) {
+    error('Script evaluates false: top stack item is false')
+    return false
+  }
+
+  return true
+}
+
+// Opcode Functions
+const opFunctions: { [key: string]: Function } = {
```
benalleng, line 111: One thing we might want to consider is to separate these into operations and objects so that we can more easily make sense of these when trying to create this styling element for the stack 

image

alternatively we might want to create them as objects to allow us to add some key like `isOperation`

benalleng: OP_1 and OP_2 simply add to the stack while OP_ADD modifies it

0tuedon: Oh yh, and operation clears out on the next stack

ui/lesson/OpCodeParser/index.ts
```diff
@@ -0,0 +1,440 @@
...
+  scriptFieldRef: InterpreterRef,
+  initialStackRef: InterpreterRef,
+  executionStackRef: InterpreterAreaRef,
+  currentOpCodeRef: InterpreterDivRef,
+  heightRef: InterpreterRef
+): InterpreterState => {
+  const state: InterpreterState = {
+    ...initialState,
+    script: scriptFieldRef.current?.value.split(' '),
+    stack: [],
```
benalleng, line 284: we will want this to be initialized by the initial stack input
``` suggestion
    stack: initialStackRef.current?.value.split(' '),
```

ui/lesson/OpCodeParser/OpClass.tsx
```diff
@@ -0,0 +1,83 @@
...
+    )
+    runInterpreter.onReset()
+  }
+  return (
+    <div className="flex flex-col px-8">
+      <div className="mt-4 flex flex-col">
+        <input ref={scriptField} type="text" onChange={scriptOnchange} />
+      </div>
+      <div className="mt-4 flex flex-col">
+        <input ref={initialStack} type="text" onChange={scriptOnchange} />
```
benalleng, line 60: then I think we will want changing the inital stack to reset everything
``` suggestion
        <input ref={initialStack} type="text" onChange={reest} />
```

ui/lesson/OpCodeParser/index.ts
benalleng: we could also `split(', ')`

ui/lesson/OpCodeParser/index.ts
benalleng: I think we could use the headings from this page https://en.bitcoin.it/wiki/Script to organize them, it would help with thinking of how to style / color them

benalleng: there are
- constants
- flow control
- stack
- splice
- bitwise logic
- arithmetic
- crypto
- locktime
- reserved words

benalleng: ultimately I think this logic should be taken out and put separately and then imported here

ui/lesson/OpCodeParser/index.ts
```diff
@@ -0,0 +1,440 @@
...
+interface Refs {
+  scriptFieldRef: InterpreterRef
+  initialStackRef: InterpreterRef
+  executionStackRef: InterpreterAreaRef
+  currentOpCodeRef: InterpreterDivRef
+  heightRef: InterpreterRef
+}
+
+const initialState: InterpreterState = {
+  exec_ptr: -1,
```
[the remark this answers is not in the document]
benalleng: Not critical but something we might want to look into

benalleng (commented):
i've also been having trouble getting if statements to work correctly

ui/lesson/OpCodeParser/index.ts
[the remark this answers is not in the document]
0tuedon: okay, taking a look

0tuedon:
> i've also been having trouble getting if statements to work correctly

Okay

ui/lesson/OpCodeParser/index.ts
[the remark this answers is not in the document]
benalleng: nope nevermind, I see what I did wrong

ui/lesson/OpCodeParser/OpClass.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const reset = () => {
+    runInterpreter = initializeInterpreter(
+      scriptField as InterpreterRef,
+      initialStack as InterpreterRef,
+      executionStack as InterpreterAreaRef,
+      currentOpcode as InterpreterDivRef,
+      height as InterpreterRef
+    )
+    runInterpreter.onReset()
+  }
```
benalleng, line 53:
``` suggestion
  }

  const step = () => {
    runInterpreter.onStepClick()
  }

```

ui/lesson/OpCodeParser/OpClass.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <div className="mt-4 flex flex-col">
+        <textarea ref={executionStack} />
+      </div>
+
+      <div className="flex gap-3">
+        <button type="button" className="cursor-pointer" onClick={runOnClick}>
+          Run
+        </button>
+        <button onClick={reset}>Reset</button>
+        <button>Step</button>
```
benalleng, line 77:
``` suggestion
        <button onClick={step}>Step</button>
```

benalleng: we should be able to allow step changes like this

ui/lesson/OpCodeParser/index.ts
0tuedon: Should we do this after the stack has been updated
for example

`export const opCodeTypes = (opcode: string) => {
  if (opcode.match(/OP_[0-9]*/)) {
    return { opcode, operation: 'constants' }
  }
  switch (opcode) {
    case 'OP_DUP':
      return { opcode, operation: 'stack' }
    case 'OP_ADD':
      return { opcode, operation: 'arithmetic' }
    case 'OP_EQUAL':
      return { opcode, operation: 'bitwise' }
      }
      `
      
      I have something like this, isn't it best to check on the stack and not during the operation

ui/lesson/OpCodeParser/OPFunctions.ts
```diff
@@ -0,0 +1,186 @@
...
+    case 'OP_DUP':
+      return { opcode, operation: 'stack' }
+    case 'OP_DUP':
+      return { opcode, operation: 'stack' }
+    case 'OP_DUP':
+      return { opcode, operation: 'stack' }
+    default:
+      break
+  }
+}
```
benalleng, line 25: what does this do?

0tuedon: So it should be able to get what type of operation each opcode code does

kouloumos: why the repetition of `OP_DUP`?

ui/lesson/OpCodeParser/OPFunctions.ts
kouloumos, line 1: the correct is "Opcode" instead of "OpCode"

ui/lesson/OpCodeParser/OPFunctions.ts
0tuedon: It's updated now, it was a test of how it would look for other Opcodes.

0tuedon:
Did a rewrite of how the OpRunner works In https://github.com/saving-satoshi/saving-satoshi/pull/1011

PR #1011, opened 2024-07-10
WIP: Challenge Component Rework

0tuedon:


benalleng (approved):
Looks good! we can continue to modify this as we add in the copy half of the lesson

PR #1013, opened 2024-07-17
Update Chapter 5 help pages

### LINKED ISSUE #972: Chapter 5 help pages
Similarly to chapter 4 these pages need some psuedo code solutions instead of actual answers.

Additionally this chapter is where we really dropped off adding any additional reading resources so we should at least try and add some

benalleng:
Closes #972

[Preview](https://saving-satoshi-git-fork-benalleng-chapter-5-help-savingsatoshi.vercel.app/en/chapters/chapter-5/intro-1?dev=true)


content/resources/chapter-5/verify-signature-5.tsx
```diff
@@ -68,15 +57,16 @@ function verify(sig_r, sig_s, pubkey_x, pubkey_y, msg) {
...
     return x1;
   }
-  // Implement ECDSA!
-  //   u1 = m / s mod n
-  //   u2 = r / s mod n
-  //   R = G * u1 + A * u2
-  //   r == x(R) mod n
   const sig_s_inverted = invert(sig_s);
+  // Implement ECDSA!
+  //   u1 = msg * s mod n
```
satsie, line 62:
```suggestion
  //   u1 = msg * s^(-1) mod n
```

This should be msg * the **inverse** of s (recall that 1/n is the same as 1 * n^-1, I haven't thought of this in at least 10 years and was caught off guard when I originally did this chapter :laughing:). 

I personally would restore the original `// u1 = m / s mod n` code comment because I think that's how it appears in the cryptography documentation. I see that you kept it this way for the python solution. Not sure if @pinheadmz wants to give a second opinion.

content/resources/chapter-5/verify-signature-5.tsx
```diff
@@ -68,15 +57,16 @@ function verify(sig_r, sig_s, pubkey_x, pubkey_y, msg) {
...
   const u1 = (msg * sig_s_inverted) % ORDER;
+  //   u2 = r * s mod n
```
satsie, line 64:
```suggestion
  //   u2 = r * s^(-1) mod n
```

Same thing as previous comment above

i18n/locales/en.ts
```diff
@@ -1192,19 +1192,59 @@ const translations = {
...
+          'OP_PUSHDATA in Bitcoin script plays a crucial role in facilitating the insertion of arbitrary pieces of data into the blockchain. It is actually a whole category of opcodes that allow for the inclusion of data elements of varying sizes, making Bitcoin’s scripting language more versatile. This opcode is particularly significant in enabling the implementation of various smart contract functionalities and custom transaction types. Given that we know this is supposed to push 71 bytes of data onto the stack can you work out how this opcode might be represented in hex? You can read more about some of these OP_CODES and more <Link href="https://en.bitcoin.it/wiki/Script#Constants" target="_blank" className="underline">here</Link>.',
```
satsie, line 1198:
```suggestion
          'OP_PUSHDATA in Bitcoin script plays a crucial role in facilitating the insertion of arbitrary pieces of data into the blockchain. It is actually a whole category of opcodes that allow for the inclusion of data elements of varying sizes, making Bitcoin’s scripting language more versatile. This opcode is particularly significant in enabling the implementation of various smart contract functionalities and custom transaction types. Given that we know this is supposed to push 71 bytes of data onto the stack, can you work out how this opcode might be represented in hex? You can read more about some of these OP_CODES and more <Link href="https://en.bitcoin.it/wiki/Script#Constants" target="_blank" className="underline">here</Link>.',
```

satsie: I didn't know it is a category of opcodes. that is cool!

i18n/locales/en.ts
```diff
@@ -1192,19 +1192,59 @@ const translations = {
...
       },
-      verify_signature: {
-        signature_verification_heading: 'Signature Verification',
-        signature_verification_paragraph_one:
-          'Signature verification is crucial in Bitcoin to confirm that a transaction is authorized by the holder of the private key. In the context of ECDSA, it involves checking that a signature (comprising two numbers, r and s) is valid for a given public key and message. This verification ensures the integrity and authenticity of a transaction.',
+      derive_message_four: {
+        op_checksig_heading: 'OP_CHECKSIG',
+        op_checksig_paragraph_one:
+          'OP_CHECKSIG in Bitcoin script is critically important to ensuring that the correct private key is able to spend a given transaction. In almost every Bitcoin script there will be an OP_CHECKSIG to ensure that the person attempting to spend the bitcoin is able to do so with the given key. You can read more about some of these OP_CODES and more <Link href="https://en.bitcoin.it/wiki/Script#Constants" target="_blank" className="underline">here</Link>.',
```
satsie, line 1203:
```suggestion
          'OP_CHECKSIG in Bitcoin script is critically important to ensuring that the correct private key is able to spend a given transaction. In almost every Bitcoin script there will be an OP_CHECKSIG to ensure that the person attempting to spend the bitcoin is able to do so with the given key. You can read about some of these OP_CODES and more <Link href="https://en.bitcoin.it/wiki/Script#Constants" target="_blank" className="underline">here</Link>.',
```

i18n/locales/en.ts
```diff
@@ -1192,19 +1192,59 @@ const translations = {
...
+      },
+      derive_message_six: {
+        transaction_parts_heading: 'Transaction Parts',
+        transaction_parts_one:
+          'Version: This version indicates how the the transaction will be organzied.',
```
satsie, line 1208:
```suggestion
          'Version: This version indicates how the the transaction will be organized.',
```

i18n/locales/en.ts
```diff
@@ -1192,19 +1192,59 @@ const translations = {
...
+        transaction_parts_two:
+          'Number of inputs: The number of inputs in theis transaction.',
```
satsie, line 1210:
```suggestion
          'Number of inputs: The number of inputs in this transaction.',
```

i18n/locales/en.ts
```diff
@@ -1192,19 +1192,59 @@ const translations = {
...
+        transaction_parts_three:
+          'Hash on input#0: This is the hash of the tx corresposing to input#0 subsequent inputs will go after part six.',
```
satsie, line 1212:
```suggestion
          'Hash on input#0: This is the hash of the tx corresponding to input#0. Subsequent inputs will go after part six.',
```

i18n/locales/en.ts
```diff
@@ -1192,19 +1192,59 @@ const translations = {
...
+        transaction_parts_four:
+          'Index of input#0: This is the index of the input#0 from the funding transaction as it could have many outputs.',
```
satsie, line 1214:
```suggestion
          'Index of input#0: This is the index of the input#0 from the transaction that created it. This is necessary since the transaction could have many outputs.',
```

In chapter 10 we get into Lightning and the term "funding transaction" is used. While I like using "funding transaction" in this context, I don't want to confuse readers.

i18n/locales/en.ts
```diff
@@ -1192,19 +1192,59 @@ const translations = {
...
+        transaction_parts_five:
+          'Scriptsig: This is the scriptsig that signs the input for this tx.',
+        transaction_parts_six:
+          'Input#0 sequence: This is the sequence for the spending input.',
```
satsie, line 1218:
```suggestion
          'Input#0 sequence: This is the sequence number for the spending input.',
```

By itself, "sequence" implies order, but the sequence number field doesn't really correspond to that.

i18n/locales/en.ts
```diff
@@ -1192,19 +1192,59 @@ const translations = {
...
+        transaction_parts_seven:
+          'Number of outputs: This indicates how many outputs the tx should have.',
+        transaction_parts_eight:
+          'Output#0 value: This is the amount that should be spent to the first output in little-endian bytes.',
+        transaction_parts_nine:
+          'Output#0 scriptPubKey: This is the public key plus OP_CHECKSIG to ensure that the transaction is being spent correctly.',
```
satsie, line 1224:
```suggestion
          'Output#0 scriptPubKey: This is the public key plus OP_CHECKSIG to ensure that the output will be spent correctly.',
```

i18n/locales/en.ts
```diff
@@ -1192,19 +1192,59 @@ const translations = {
...
+          'Output#1 value: This is the amount that should be spent to the second output in little-endian bytes.',
+        transaction_parts_eleven:
+          'Output#1 scriptPubKey: This is the public key of the sender so that the change output will go back into their possesion.',
+        transaction_parts_twelve:
+          'Locktime: Amount of time that this tx should be locked before the coins can be spent.',
+      },
+      derive_message_seven: {
+        sighash_type_flag_heading: 'SigHash Type Flag',
+        sighash_type_flag_paragraph_one:
+          'SIGHASH flags are a mechanism in Bitcoin that define which parts of a transaction are included in the hash that is signed by a private key. Essentially, they determine the scope of commitment by the signer to specific parts of the transaction data. The SIGHASH flag is a single byte appended to each signature and can vary between inputs within the same transaction. There are several types of SigHash flags you can learn about <Link href="https://river.com/learn/terms/s/sighash-flag" target="_blank" className="underline">here</Link>.',
```
satsie, line 1235:
```suggestion
          'SigHash flags are a mechanism in Bitcoin that define which parts of a transaction are included in the hash that is signed by a private key. Essentially, they determine the scope of commitment by the signer to specific parts of the transaction data. The SigHash flag is a single byte appended to each signature and can vary between inputs within the same transaction. There are several types of SigHash flags you can learn about <Link href="https://river.com/learn/terms/s/sighash-flag" target="_blank" className="underline">here</Link>.',
```

Opting for "SigHash" over "SIGHASH" because that's what chat.bitcoinsearch.xyz used in its answer to me :laughing: but I don't really care as long as it's consistent in this paragraph.

i18n/locales/en.ts
```diff
@@ -1192,19 +1192,59 @@ const translations = {
...
+        signature_verification_heading: 'Signature Verification',
+        signature_verification_paragraph_one:
+          'Signature verification is crucial in Bitcoin to confirm that a transaction is authorized by the holder of the private key. In the context of ECDSA, it involves checking that a signature (comprising two numbers, r and s) is valid for a given public key and message. This verification ensures the integrity and authenticity of a transaction.',
+      },
+      verify_signature_three: {
+        signature_encoding_heading: 'Signature Encoding',
+        signature_encoding_paragraph_one:
+          'A (DER) signature or Distinguished Encoding Rules is simply a formate used to encode an ECDSA signature in Bitcoin. An ECDSA signature is generated using a private key and a hash of the signed message. It consists of two 32-byte numbers (r,s). It has multiple components you can learn more about <Link href="https://bitcoin.stackexchange.com/questions/12554/why-the-signature-is-always-65-13232-bytes-long/12556#12556" target="_blank" className="underline">here</Link>.',
```
satsie, line 1247:
```suggestion
          'A (DER) signature or Distinguished Encoding Rules is simply a format used to encode an ECDSA signature in Bitcoin. An ECDSA signature is generated using a private key and a hash of the signed message. It consists of two 32-byte numbers (r,s). It has multiple components you can learn more about <Link href="https://bitcoin.stackexchange.com/questions/12554/why-the-signature-is-always-65-13232-bytes-long/12556#12556" target="_blank" className="underline">here</Link>.',
```

satsie: This is a pretty good not bad resource on DER signatures in Bitcoin :wink: https://technicaldifficulties.io/2020/07/22/bip-66-unpacking-der-signatures/

Matthew even gets a shoutout in it for catching an error in a previous revision :smile:

i18n/locales/en.ts
```diff
@@ -1226,11 +1266,26 @@ const translations = {
...
       },
+      validate_signature_two: {
+        base64_encoding_heading: 'Base64 Encoding',
+        base64_encoding_paragraph_one:
+          'Base64 is a basic byte to text encoding scheme that just allows for the data to be convereted to bytes which can then be used in buffers or from bytes to text in such a way that it avoids any problems with url paths and query params. This differs from the baese58 encoding scheme in that it still includes the similar characters that may otherwise be confusing to a user when trying to copy or dicate the encoded message.',
```
satsie, line 1277:
```suggestion
          'Base64 is a basic byte to text encoding scheme that just allows for the data to be convereted to bytes which can then be used in buffers or from bytes to text in such a way that it avoids any problems with url paths and query params. This differs from the base58 encoding scheme in that it still includes the similar characters (zero, uppercase 'O', uppercase 'I', and lowercase 'l') that may otherwise be confusing to a user when trying to copy or dictate the encoded message.',
```

i18n/locales/en.ts
```diff
@@ -1226,11 +1266,26 @@ const translations = {
...
+      },
+      validate_signature_three: {
+        signing_and_ownership_heading: 'Message Signing and Ownership',
+        signing_and_ownership_paragraph_one:
+          "Based on how Bitcoin's psuedo-anonymity we are limited in what we are able to prove because someone can refuse to sign a message even if they do infact have onwership of the correct key to sign a message. However, given this particular key we are able to prove that this key and the owner of said key is not the same person as Satoshi in this case because of the fact that this key does not properly sign the message.",
```
satsie, line 1282:
```suggestion
          "Based on Bitcoin's pseudo-anonymity we are limited in what we are able to prove because someone can refuse to sign a message even if they do in fact have ownership of the correct key to sign a message. However, given this particular key we are able to prove that this key and the owner of said key is not the same person as Satoshi  because this key does not properly sign the message.",
```

i18n/locales/en.ts
```diff
@@ -1226,11 +1266,26 @@ const translations = {
...
+      },
+      validate_signature_four: {
+        one_for_one_heading: 'One for One',
+        one_for_one_paragraph_one:
+          'To ensure the reliablility and security of the ECDSA algortithm A signature created by a private key can only be verified by the corresponding public key. If different private keys could produce the same public key or signature, it would undermine the security and reliability of ECDSA. Through this we can know that the public key that was able to sign this message has ownership of this bitcoin.',
```
satsie, line 1287:
```suggestion
          'To ensure the reliability and security of ECDSA, a signature created by a private key can only be verified by the corresponding public key. If different private keys could produce the same public key or signature, it would undermine the security and reliability of ECDSA. Through this we can know that the public key that was able to sign this message has ownership of this bitcoin.',
```

satsie (approved):
Great job with this! I enjoyed reading the new resources you added. Found some typos and minor things but otherwise this looks good to go :raised_hands:

i18n/locales/en.ts
pinheadmz: Hm yeah the language here is a little funny. Satoshi could have created the invalid signature but it wouldn't prove anything. That's all we can know for sure: this invalid signature is NOT PROOF that whoever created it is Satoshi

pinheadmz (commented):
Thanks great work on this Ben and great review from Stasie. I just left a few annoying nits

PR #1014, opened 2024-07-19
WIP: Lesson Component for OPRunner

0tuedon:
## Pull Request checklist

- Clean commit history
[Preview](https://saving-satoshi-git-fork-0tuedon-chap9-less-ba33fc-savingsatoshi.vercel.app/en/chapters/chapter-9/splits-1?dev=true)


pinheadmz:
So cool Tuedon! I know this is WIP but here's some initial feedback:

- I might not have been specific in the help text but it should be clear that "initial stack" is stuff separated by commas:

Screenshot 2024-07-23 at 2 27 22 PM

- After separating on commas, be sure to trim whitespace around each opcode
- `0 + 0 = 0`, I think the `null` here is a JS oddness
- Got "requires 2 items" even though 2 items were provided
- The sum of the two operands is not being pushed back to the stack by OP_ADD

Screenshot 2024-07-23 at 2 28 27 PM

pinheadmz:
So cool Tuedon! I know this is WIP but here's some initial feedback:

- I might not have been specific in the help text but it should be clear that "initial stack" is stuff separated by commas (not spaces):

Screenshot 2024-07-23 at 2 41 56 PM

- `0 + 0 = 0`, I think the `null` here is a JS oddness
- Got "requires 2 items" even though 2 items were provided

Screenshot 2024-07-23 at 2 44 06 PM


- Unrecognized data should probably throw an error. In real Bitcoin its all just bytes anyway but in our game we only allow:
    - Integers
    - hex strings prefixed with "0x" so we can do the pre-image / hash thing
    - PUBKEY(...) SIG(...) HASH256(...)


Screenshot 2024-07-23 at 2 45 14 PM

benalleng:
I modified the placeholder and the logic to use spaces as well as adding a title to the input so that users may see rules for the input if they stay focused on it long enough

image

As well I added the relevant copy for this challenge

benalleng:
I also changed the stack bucket placeholder copy to be hopefully clearer of its purpose

image

benalleng:
> It looks like we are not going with the automatically updating visualization. In that case, the run/reset/step buttons need to be treated very differently and be more prominent. They might also be better positioned in the visualizer block itself, since that is the area they are affecting. We also need better indicators for when you have made changes that are not visualized yet.
> 
> Making users manually update the visualization feels a bit backwards, but if you think it's the better way to go, then I'll be quiet. I'm not the target audience.

This had completely left my mind, but I do think we can do this. My only concern is about errors and rapid testing, Do you think the whole result should be cleared the moment they start typing or wait until they finish typing to reset?

benalleng:
> Are you planning to add those details to the visualizer that indicate which blocks were added in each step (blue), a successful or erroneous last step (red), and the arrows and outlines that visually describe what happened?

We are but I do have some questions for @pinheadmz about the possibilities for the arrows, I might be overthinking how we render these but other than things like OP_ADD going from 2 => 1 OP_DUP going from 1 => 2 and OP_HASH going from 1 => 1 are there any other op_codes you can think of that we might need to handle with different arrow svgs

pinheadmz:
> are there any other op_codes you can think of that we might need to handle with different arrow svgs

Yes for example CHECKSIG pops two items and pushes one (pubkey, sig -> boolean) and CHECKMULTISIG pops even more depending on arguments themselves! When you get to the full script tool there'll be a ton more like ROT pops three and then pushes three, etc. I can make a chart for you or we could look an existing script interpreter from bitcoin core or bcoin to figure out the best way

GBKS:
> In other words do we think their old stack should stay until they finish typing or disappear immediately

I would keep it around until the next re-render.

GBKS:
Here's a quick mock-up with some visual tweaks for the OP code explainers on the left, validation error message, stack tweaks, and I also moved the run/reset/step buttons into the stack visualizer to make them more contextual. Hope that's helpful.

saving-satoshi-script-editor-240726

benalleng (approved):
I think this is a great place to move on to the rest of the challenges in chapter 9.  I know there are a few additional design tweaks that can be made but this is a great mvp for the component!

PR #1015, opened 2024-07-22
feat: Added chapter 6 hard mode

daviroo:
I haven't been able to test this yet as I ran into that docker bug locally again so going to have to clear out my local docker again before progressing.  If someone else could give it a test it would be great 🙏 .

### DIFF
--- a/state/progressState.ts
+++ b/state/progressState.ts
@@ -1025,3 +1025,23 @@ export const markLessonAsCompleteAtom = atom(
     })
   }
 )
+
+export const setChapterDifficultyAtom = atom(
+  null,
+  (get, set, { chapterId, difficultyLevel }) => {
+    const courseProgress = get(syncedCourseProgressAtom)
+    const updatedChapters = courseProgress.chapters.map((chapter) => {
+      if (chapter.id === chapterId) {
+        return {
+          ...chapter,
+          selectedDifficulty: difficultyLevel,
+        }
+      }
+      return chapter
+    })
+    set(syncedCourseProgressAtom, {
+      ...courseProgress,
+      chapters: updatedChapters,
+    })
+  }
+)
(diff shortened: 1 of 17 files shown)


0tuedon:
Will check it out

benalleng:
Looks great! tested on dev but I was able to access hard mode and normal mode respectively when toggling.

ui/chapter/Chapter.tsx
```diff
@@ -140,6 +142,10 @@ export default function Chapter({ children, metadata, lang }) {
             {t(chapter.metadata.title)}
           </h3>
 
+          {chapterWeAreRendering?.hasDifficulty && (
```
benalleng, line 145: We should probably also make sure this doesn't render until that chapter has been unlocked

benalleng (approved):
I wanted to get to chapter 6 naturally and everything is in order except it looks like the first page of each chapter (excluding chapter 1) is locked even though I can start the chapter via the start button.

image

image

state/progressState.ts
benalleng, line 1: One thing I think we missed is a `markChapterAsComplete` function for the end of a chapter. We include and read a `completed` key on each chapter but leave no way to modify that key AFAIK

benalleng: nvm I found where it is but I think the logic might be broken, looking into it

state/progressState.ts
```diff
@@ -984,8 +984,8 @@ export const markLessonAsCompleteAtom = atom(
       }
     }
 
-    // If no next lesson is found in the current chapter, look for the next chapter
-    if (!nextLesson) {
+    // If the next lesson is not in the same chapter as the current chapter, look for the next chapter
+    if (currentChapter?.id && currentChapter.id !== nextChapter) {
```
benalleng, line 988: I need to take a look at this logic again

PR #1016, opened 2024-07-23
chore: Removed unused python import

daviroo:
in chapter-5/verify-signature-2 we don't need to import base64 in python since this challenge is just hashing not a string decoding.


pinheadmz (approved):
ACK

benalleng (approved):

PR #1017, opened 2024-07-23
chore: Merge default state with saved progress state

daviroo:
Merging the state from the server over the default state means we can release new chapters by adding to the default state.

https://github.com/TehShrike/deepmerge


benalleng (approved):

daviroo:
Now using custom merge function because the library merge functions weren't working.

PR #1019, opened 2024-07-28
fix: Fixed race condition for local storage write on signup

daviroo:
There was a race condition when writing the session token on signup so occasionally the signup modal would fail to close.

### DIFF
--- a/state/progressState.ts
+++ b/state/progressState.ts
@@ -794,7 +794,7 @@ export const loadProgressAtom = atom(null, async (get, set) => {
 
   set(isLoadingProgressAtom, true)
 
-  let progress = null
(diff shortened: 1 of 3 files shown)


state/progressState.ts
benalleng, line 1: I changed the loadProgressAtom function as it would soft lock a user at the enc of chapter one in this commit https://github.com/saving-satoshi/saving-satoshi/pull/1019/commits/32d8b238b6dffd2ea0a0734ce096abae9822734a 

[Preview](https://saving-satoshi-2mazp4eoe-savingsatoshi.vercel.app/)

benalleng (approved):
Ok! one step closer!

PR #1021, opened 2024-07-28
Test: Logging out no longer immediately clears user progress

benalleng:
This is a test to try and probe where there is a failure in fetching user progress from the backend.


benalleng:
Closed, our backend wasn't deployed properly

PR #1022, opened 2024-07-28
Text color change reverted from statusbar back to white

benalleng:
reverts a change I think I let slip through that changed the text color of the status bar


0tuedon:
One more change, I think we need to remove 

` setScript(event.target.value.toUpperCase())` to ` setScript(event.target.value)`, it affects the second challenge or better still make initial script uppercase too?

0tuedon (approved):
LGTM!

PR #1023, opened 2024-07-28
Script case

benalleng:
This makes all text inputs in the OPRunner upper case and fixes an error in `OP_EQUALVERIFY` not returning anything


0tuedon (approved):
Looks Good, I will have to fix some minor UI issues with the stack overflowing to the status bar.

PR #1030, opened 2024-07-30
Chapter 9 OpCode Challenges

benalleng:
Adds all but the copy intro for the opcode challenges in chapter 9.

[Preview](https://saving-satoshi-git-fork-benalleng-chapter-4c77c3-savingsatoshi.vercel.app/en/chapters/chapter-9/opcodes-2?dev=true)


pinheadmz:
Screenshot 2024-07-31 at 10 52 04 AM
a bit funny overflow behavior here

pinheadmz:
Screenshot 2024-07-31 at 10 53 12 AM

Do we need to force all caps in this field? I was thinking `SIG(holocat)` made it a bit more clear that there was a function and an argument

pinheadmz:
Screenshot 2024-07-31 at 10 55 51 AM

OP_CHECKMULTISIG should push `false` to the stack if inputs don't verify.

pinheadmz:
Screenshot 2024-07-31 at 10 58 17 AM

another funny overflow here

pinheadmz:
Screenshot 2024-07-31 at 11 04 32 AM

I find this mix of fonts a little distracting

pinheadmz:
Screenshot 2024-07-31 at 11 07 54 AM

shouldn't allow unrecognized data types in the stack

pinheadmz:
Screenshot 2024-07-31 at 11 09 35 AM

I think we should explain the script failure when we can. In this example there is more than one item left on the stack, and the top item is false. (either or both)

pinheadmz:
Screenshot 2024-07-31 at 11 17 42 AM

The default height is a x10 short I think it should be 6,300,001

pinheadmz (commented):
This is looking really dope great work!

0tuedon:
Do we need @pinheadmz  other changes first or we can merge? @benalleng

benalleng:
I'm going to try and fix the check multisig error and multiple stack items error first

0tuedon:
> I'm going to try and fix the check multisig error and multiple stack items error first

Okay will rebase on your pr for my issue. also adding the design fix overflow to it

GBKS:
Aloha. I was just poking around a little and noticed a few things. Biggest one is probably how long the stack text can get. What do you think of having them stretch horizontally if the content is wider than 160px? And formatting function names in there with a different color?

image

Also, the stack text in the preview goes to 3 lines because there is no text size specified. It should be 13px.

This visual is getting quite complex, I hope we can keep it simple enough for users to easily make sense, while still visualizing all the logic.

benalleng:
> I think we should explain the script failure when we can. In this example there is more than one item left on the stack, and the top item is false. (either or both)

I made a new check for the multiple items on the stack but we can certainly add more as we find different failure states

pinheadmz:
Screenshot 2024-08-05 at 3 09 58 PM

Found another way to break it :-) floats on the stack!

benalleng:
> Found another way to break it :-) floats on the stack!

fixed? lol js, never change 
image

0tuedon (approved):
Looks way better, we can have other minor changes in a new PR as discussed. 🥳🥳🥳

PR #1034, opened 2024-08-02
fix: correct elliptic curve equation s/x^2/x^3/ in chapter 4

theStack:
Adapt the curve equation to match secp256k1: $y^2 = x^3 + 7$
Note that the curve $y^2 = x^2 + 7$ would look very different, see e.g. https://www.desmos.com/calculator/ for trying out.


benalleng (approved):
Good catch! that's an important typo, here's the [preview](https://saving-satoshi-git-fork-thestack-fixcurveequation-savingsatoshi.vercel.app/en/chapters/chapter-4/public-key-2?dev=true)

PR #1035, opened 2024-08-07
Add step function

benalleng:
My attempt at rebasing the arrow and step functionality into our current dev HEAD

[Preview](https://saving-satoshi-git-fork-benalleng-add-step-3490f8-savingsatoshi.vercel.app/en/chapters/chapter-9/opcodes-5?dev=true)


0tuedon:
I noticed we can't step or Run when we get our answer and we get a Success message when I just keep changing the stack

daviroo:
> I noticed we can't step or Run when we get our answer and we get a Success message when I just keep changing the stack

Let's go over it today on the call.  I think we need to align on what the best behaviour is.

0tuedon:
> > I noticed we can't step or Run when we get our answer and we get a Success message when I just keep changing the stack
> 
> Let's go over it today on the call. I think we need to align on what the best behaviour is.

Yh Looking forward to that

content/lessons/chapter-9/opcodes-5.tsx
```diff
@@ -24,7 +24,7 @@ export default function OpCodes5({ lang }) {
         answerScript={['OP_ADD', 'OP_3', 'OP_EQUAL']}
         prePopulate
         showRunButtons
-        readOnly
+        //readOnly
```
0tuedon, line 27: remove comment

0tuedon (approved):
LGTM!

Issue #1038, opened 2024-08-13
OpCode Challenge Design cleanup


benalleng:
We've made great progress on our opcode challenge but I think we should maybe make a small list about things we really would like as nice-to-haves while they are fresh.

They can be anything, design, ux/ui, speed, difficulty, etc.

daviroo:
@GBKS made a comment in our last call that it would be good to have some subtle animations in the script runner.  I'm going to take a look into our options around that and follow up here.

benalleng:
I am going to create a small render function to create colored borders on opcodes of type constant and push-data like so to make it clear what is being added

![image](https://github.com/user-attachments/assets/386bedc3-09ba-45d2-8c5b-6bf87670478e)

PR #1040, opened 2024-08-14
feat: Added animation to script executor

daviroo:
Work as part of #1038


daviroo:
https://github.com/user-attachments/assets/6d6b30e4-6301-4609-870d-c51d9b55f73e

benalleng:
Looking great!

benalleng (approved):
Code looks good, we like the arrows rendering before the next stack history, but @GBKS if you don't like it we can change it to render after both are visible

PR #1047, opened 2024-08-16
chore: Protect dev mode in production

daviroo:
@benalleng Is this what you had in mind?  This would mean we would never be able to use dev mode in the production build.

### DIFF
--- a/hooks/useEnvironment.ts
+++ b/hooks/useEnvironment.ts
@@ -7,6 +7,6 @@ export default function useEnvironment() {
   const devParam = searchParams?.get('dev') || ''
 
   return {
-    isDevelopment: devParam === 'true',
+    isDevelopment: process.env.NODE_ENV !== 'production' && devParam === 'true',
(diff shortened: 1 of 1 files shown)


0tuedon:
Not really, from the call jonas said to add a .env param that we can change on production.

- For example if during TabConf we want to go through the chapters quickly or so.

daviroo:
@0tuedon so it would be something like `dev=0b46708e-ae39-47d8-9c96-0f322eefcef7` which would match the value provided in the env?

0tuedon:
> @0tuedon so it would be something like `dev=0b46708e-ae39-47d8-9c96-0f322eefcef7` which would match the value provided in the env?

Yh it can be something like that and if it get's link it can be changed

daviroo:
@benalleng @0tuedon this is ready now I believe.

benalleng (approved):
LGTM!

PR #1052, opened 2024-08-16
Chapter 4: fix secp256k1 lib 404

satsie:
Currently this link points to a private repo and users get a 404.

When do we anticipate the next release? If it's not for a while, should I make a PR to hot fix this to master?


benalleng:
I'm hoping we can make our next push next week, though I think we should consider starting to push quick hotfixes to master directly as we get close to having an official release

benalleng (approved):

satsie:
Thanks for merging this @benalleng! I agree about using the hotfix workflow as we get closer to the official release. It would also be good just to try it out and get used to using it.

Issue #1053, opened 2024-08-19
Explore draggable scratch component for OP_Codes


benalleng:
Perhaps we could use drag and drop functionality for our op_codes so that users dont actually need to type in every opcode
something like scratchpad https://scratch.mit.edu/projects/editor/?tutorial=getStarted

we can use react dnd https://github.com/atlassian/react-beautiful-dnd

PR #1055, opened 2024-08-20
Add drag and drop functionality to OpRunner

### LINKED ISSUE #1038: OpCode Challenge Design cleanup

benalleng:
Closes #1038, #1053

I have created a test drag and drop for the OpRunner to remove the text area input we have for the OpRunner now

built off of this codesandbox https://codesandbox.io/p/sandbox/react-beautiful-dnd-copy-and-drag-5trm0?file=%2Findex.js

[Pre-filled Preview](https://saving-satoshi-git-fork-benalleng-drag-and-drop-savingsatoshi.vercel.app/en/chapters/chapter-9/opcodes-5?dev=true)

[Advanced Challenge Preview](https://saving-satoshi-git-fork-benalleng-drag-and-drop-savingsatoshi.vercel.app/en/chapters/chapter-9/proposal-2?dev=true)

### TODO
- ~~state refinement to get the cards to not jump when grabbing them~~
  - It was a versioning issue with react-beautiful-dnd, I reverted to 10.0.4


benalleng:
Having an issue with the state seemingly being updated only after I let go on the drop, you see it with the whole table moving behind the one grabbed

GBKS:
Interesting idea. The preview does not let me interact, but the concept is very clear.

1. For devs, does that make it too easy?
2. It's nice that all options are laid out, but not nice how they are in a small scrollable area. I wonder if some dropdown approach would work better?

I'm happy to mock up some variations of this, just let me know.

benalleng:
I added the blank slate advanced challenge you can take a look at and play around with

> For devs, does that make it too easy?

We have the earlier challenges pre-filled in both the textarea and drag-and-drop version, but I am thinking about transitioning some of the challenges over to the advanced version to push the user a little harder

> It's nice that all options are laid out, but not nice how they are in a small scrollable area. I wonder if some dropdown approach would work better?

Space is the main challenge I am having here. The x-y-scrolling for each type of opcode was the most compact, readable, and easily accessible way to mock it up at the time but I welcome some design revisions

benalleng:
Here's a quick recording of me going through the challenge

https://github.com/user-attachments/assets/f71a70c8-ab15-474a-8bf6-bd3b5bd695fe

GBKS:
Really good work with this.

There are some visual details that I would tighten up, as outlined in the image below. Mostly smaller fonts and slightly tighter spacing on the various containers, lightening the divider colors, etc. 

Some extra thoughts:

1. For the OP_CODE, I don't like the puzzle piece look very much. I'd either use the arrows or just got with a plain box, which I think works just fine.
2. For interactions, how about letting you just click/tap an OP_CODE to add it to the script?
3. If we're going with our good old play button from the previous chapter at the bottom, we should also use the "Run the script" label instead of "Complete the challenge above to continue". It's a bit more clear.
4. Another idea is to put the play button next to "Your script". That connects it directly with what is being executed.
5. Vertical scrolling is not quite right yet. The execution stack goes underneath the bottom bar and I can't reach it on a shorter screen.

Frame 457

On mobile, I wonder if we can use the available space a little better (like getting rid of blank space by expanding the list of OP_CODEs). I see we're still using the "vh" unit instead of "dvh", which considers the browser chrome. Maybe that could help with the placement of the bottom bar directly at the bottom of the visible screen height. Would also be good if the bottom bar could adjust in height vertical if the text breaks to two lines.

image

That's it for now. Hope it's helpful. And thanks for pushing in the drag-and-drop direction, I think it works really well.

0tuedon (approved):
LGTM!, waiting for @GBKS  review

GBKS:
Nice progress. There are still some spacing and font size differences. Is that something you could tackle?

Regarding the boxy OP_CODEs, I think it's more consistent this way. We also show OP_CODE boxes in the copy on the left, and those did not have any arrows or puzzle shapes. That being said, what do you think about making them even more consistent? As of right now, the OP_CODEs on the left have blue text, and the ones on the right have white text. Should those be the same? Is it legible enough? Should the font sizes and padding also match?

image

GBKS:
Thanks for tweaking the sizing. I still see various differences:

- Headers are too big ("Your script", "Initial stack"...)
- OP_CODE: font is too big, side and top/bottom spacing does not match the design.
- OP_CODE group header is too big ("constant", "arithmetic"...)
- There is some blank space between the divider line and the darker background at the top of the OP_CODE library
- Stack and block height placeholder text should be white with 50% transparency
- Stack and block height input text sizes are too big
- Too much space between the "|>" button and "Run the script"
- Font size of the "OP_CODE" label in the empty stack placeholder is too big

Goal of this very precise and tight spacing is to get as much of editor on the screen as possible, even on less-tall screens.

benalleng:
Ok! I went through and made sure everything was pixel perfect!

PR #1056, opened 2024-08-21
Run Section Oprunner

### LINKED ISSUE #1048: Use Run button for OpCode Challenge
Right now we are using a typing timeout to run our opcode script, but ultimately I think we should transition to UX where thee user has control to run the script as the timeout makes the user feel like they are racing against the clock.

something like the run button below

image

0tuedon:
This PR closes #1048 

[preview](https://saving-satoshi-git-fork-0tuedon-run-button-opcode-savingsatoshi.vercel.app/en/chapters/chapter-9/proposal-2?dev=true)

### DIFF
--- a/components/SignIn.tsx
+++ b/components/SignIn.tsx
@@ -40,7 +40,6 @@ export default function SignIn({
       const loginSuccess = await attemptLogin(privateKey.toLowerCase())
       if (loginSuccess) {
         setPrivateKey('')
-        console.log('loading progress')
         await loadProgress()
         setHasAccount(true)
       } else {
--- a/content/lessons/chapter-3/coop-1.tsx
+++ b/content/lessons/chapter-3/coop-1.tsx
@@ -62,7 +62,7 @@ export default function Coop1({ lang }) {
             width={35}
             height={35}
           />
-          <Title>HoloCat Pool</Title>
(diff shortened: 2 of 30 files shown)


benalleng (changes requested):
It looks like the success status is not carrying over the "looking good so far..." state

benalleng:
As well the background needs to carry over for success states

benalleng:
looks like the top-border is doubled up, note the left side is thinnner than the right, also the background is still not green on success

image

benalleng:
Think we should disable the run while it is in the process of running?

0tuedon:
> Think we should disable the run while it is in the process of running?

I think we have a running state on other chapters

benalleng (changes requested):
We have some final sizing issue between a normal status Bar and the Run Button

in this PR
image

normal
image

and I think we should have the run button disappear on success status and share the background of the status bar otherwise

0tuedon:
I do usually have issues building when lottie isn't on dynamic,
https://github.com/saving-satoshi/saving-satoshi/pull/1056/commits/5aa964511be4942d8c1853fed0c94ddf79e27cbb
could it be the typescript version?

benalleng (approved):
Looks great!

PR #1062, opened 2024-08-26
Chapter 6, put-it-together-2-normal: fix bug with curly brace

satsie:
In the Chapter 6 put-it-together-2-normal exercise, the default code that is loaded has an extra curly brace, causing a syntax error in the text editor. It puts the `compute_input_signature` method outside of the `Transaction` class, which I don't think is correct.

I believe the extra curly brace is coming over from the `combinedCode` variable, so I adjusted the slice parameter to not bring it in.

Before: offending curly brace on line 85
Screenshot from 2024-08-26 15-05-46

After: (the `constructor()` and `digest()` methods were collapsed in the Before screenshot)
Screenshot from 2024-08-26 15-16-01

### DIFF
--- a/content/lessons/chapter-6/normal/put-it-together-2.tsx
+++ b/content/lessons/chapter-6/normal/put-it-together-2.tsx
@@ -183,8 +183,8 @@ console.log("KILL")`,
     },
     defaultCode: `const secp256k1 = require('@savingsatoshi/secp256k1js');
 const {randomBytes} = require('crypto');
-${combinedCode.slice(0, -2)}
-// YOUR CODE HERE
+${combinedCode.slice(0, -3)}
+  // YOUR CODE HERE
   compute_input_signature(index, key) {
     assert(typeof key === 'bigint');
     assert(Number.isInteger(index));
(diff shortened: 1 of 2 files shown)


satsie:
Can someone double check this for me? If you try to reset the terminal for this exercise, do you have the extra curly brace? I feel like there's a slight chance whatever this bug is, is specific to my account. I'm no longer able to reproduce it locally.

benalleng:
I am going to run through the chapter for this now, I am hesitant to play with the string slicing as It can often bite you if a user adds an extra line or doesn't close their function on a new line but I will make sure this extra curly bracket is not the default.

benalleng:
I wasn't able to reproduce this one, it's possible this was another unfortunate bug from when we pushed some edits to chapter 6 with the normal-hard mode transition..?

Was this one after another? like you had just finished the digest challenge and then proceeded to the next page or did you see this after loading in some progress?

satsie:
@benalleng thanks for testing! I can't recall how I got to it. I've been toggling between normal/hard a lot to write down the topics that each exercise covers. I'm going to close this PR out for now. If I can reliably reproduce it then I'll open it again with some more detail on how it happens.

satsie:
Good news! I was not able to reproduce the bug in incognito mode with `dev=true`. It's probably just related to my account, haha.

PR #1063, opened 2024-08-28
fix: Fixed logout progress loss bug

daviroo:
Fixed a bug noticed by @satsie where progress would be reverted to default when switching between 2 different accounts.

### DIFF
--- a/state/AuthFunctions.ts
+++ b/state/AuthFunctions.ts
@@ -3,6 +3,8 @@ import { accountAtom, isAuthLoadingAtom } from './state'
 import { login, getSession, logout } from 'api/auth'
 import { SAVING_SATOSHI_TOKEN } from 'config/keys'
 import { defaultProgressState, syncedCourseProgressAtom } from './progressState'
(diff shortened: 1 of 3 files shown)


state/AuthFunctions.ts
```diff
@@ -32,8 +38,6 @@ export const useAuthFunctions = () => {
   const attemptLogout = async () => {
     await logout()
     localStorage.clear()
-    localStorage.removeItem('SavingSatoshiProgress')
-    setCourseProgress(defaultProgressState)
```
satsie, line 36: Are we sure we should remove these lines? When I log out, a bunch of challenges remain unlocked.

benalleng: I think we can keep `setCourseProgress(defaultProgessState)` if we also redirect the user to the `chaptersUrl` with no need to save the local Progress if they have an account.
But localStorage.clear() "should" do what we need already

benalleng (approved):

Issue #930, opened 2024-05-07
Feedback up to Chap 6, challenge 3


adamjonas:
Can we run the submit code with a keyboard shortcut instead of making people use their mouse so much? ⌘+return perhaps? 

After I input my secret key, I am redirected to the homepage and the avatar changes. I have no other indication that I put my key in properly. I would recommend an update to the modal that gives the option to "go where I left off" or at least redirect to the chapters page.

the message and advancement are still below the fold:
<img width="1364" alt="Screenshot 2024-05-06 at 10 45 11 AM" src="https://github.com/saving-satoshi/saving-satoshi/assets/755825/93c180b6-9553-497e-aa6b-963557059780">

<img width="833" alt="Screenshot 2024-05-06 at 10 48 54 AM" src="https://github.com/saving-satoshi/saving-satoshi/assets/755825/f78202c5-dd14-459a-87bf-8dbe5a46ef9d">

https://savingsatoshi.com/en/chapters/chapter-2/hashing-2 on the other hand has great vertical spacing.

Not clear to me what us going on here:
<img width="254" alt="Screenshot 2024-05-06 at 12 09 59 PM" src="https://github.com/saving-satoshi/saving-satoshi/assets/755825/2938d964-a453-4d8f-80aa-7ce0ef74a0e7">
It doesn't really look like a link and the spacing is crammed together
<img width="218" alt="Screenshot 2024-05-06 at 12 14 08 PM" src="https://github.com/saving-satoshi/saving-satoshi/assets/755825/44335bc3-bd18-47dc-aac0-f0e31ae2413d">
This one has better spacing but now there is a comma and I still don't immediately recognize this as a link
<img width="250" alt="Screenshot 2024-05-06 at 12 15 15 PM" src="https://github.com/saving-satoshi/saving-satoshi/assets/755825/39fee666-e9c6-413f-94ec-9a1f51e80d6c">
This one is yet another format

We lost the context as to why this is funny. Wow a mechanical keyboard, I've heard about these legendary devices or I've heard they were banned because they were so loud everyone lost their hearing. Need a copy update here.
<img width="698" alt="Screenshot 2024-05-06 at 12 11 21 PM" src="https://github.com/saving-satoshi/saving-satoshi/assets/755825/bb539aaa-2140-44bc-923f-2049dfa2511c">

Good addition here:
[...]

PR #1068, opened 2024-08-29
ch6 in-out-four-normal: add more detail to instructions

satsie:
In this PR I've fleshed out some of the instructions for chapter 6's in-out-four-normal challenge. I'd like to copy some of these updates over to the hard version of the challenge, but will do so as a separate PR.

Before:

Screenshot from 2024-08-29 14-54-03

After:

Screenshot from 2024-11-06 21-52-19

Screenshot from 2024-11-06 21-56-41

[Preview](https://saving-satoshi-git-fork-satsie-ch6-challen-611092-savingsatoshi.vercel.app/en/chapters/chapter-6/in-out-4-normal?dev=true)

### DIFF
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -1358,22 +1358,30 @@ const translations = {
         nav_title: 'The input class',
         heading: 'Looking at the Input class implementation',
         paragraph_one:
-          'Bitcoin transaction inputs always point to existing, unspent transaction outputs. Therefore, our Input class has a method <span className="text-green"> from_output() </span> which is used to construct an Input by passing in the output description:',
+          'Here we have code for two classes: an <span className="font-bold">Input</span> class and an <span className="font-bold">Outpoint</span> (not "output"!) class.',
         paragraph_two:
-          'The first two arguments are the transaction ID and the index of the output of that transaction you want to spend from. Eventually we will pass in the txid and vout values you got above from listunspent. ',
-        paragraph_three: {
-          a: 'Hashes in bitcoin are ',
+          'Inputs come from unspent transaction outputs. If you provide the description of an output to the <span className="text-green p-1 font-mono bg-[#00000033] m-1 text-base">from_output()</span> method, it will create an instance of the Input class:',
+        paragraph_three: 'The first two arguments are:',
+        paragraph_four:
+          '1. <span className="font-bold">txid: </span>the ID of the transaction that created the output, and',
+        paragraph_five:
+          '2. <span className="font-bold">vout: </span>the index of the output in the transaction\'s entire list of outputs',
+        paragraph_six:
+          'Together, these two pieces of information make up an <span className="font-bold">Outpoint</span>. Eventually we will pass in the txid and vout values that came from executing the <span className="text-green p-1 font-mono bg-[#00000033] m-1 text-base">listunspent</span> command in the previous exercise. ',
+        paragraph_seven:
+          "The second two arguments are the value of the output we want to spend (in satoshis) and something called a scriptcode. That data is not needed until later so let's temporarily use an empty byte array.",
+        paragraph_eight: {
+          a: 'Hashes in bitcoin are',
           b: {
-            text: ' reversed ',
+            text: ' reversed',
             question: 'Why do we reverse hashes in bitcoin?',
             href: 'https://chat.bitcoinsearch.xyz/?author=holocat&question=why%2520are%2520hashes%2520reversed%2520in%2520bitcoin',
           },
-          c: ' when we see them as hexadecimal strings. When we accept hashes as strings from a user we must reverse the byte order before storing or transmitting them as raw bytes. This is why we reverse the byte order of the txid argument that is passed in here.',
+          c: `, but only when presented to or entered by a user. When a hash is provided in hexadecimal format, the byte order must be reversed before storing or transmitting the data as raw bytes.`,
         },
-        paragraph_four:
-          "The second two arguments are the value of the output we want to spend (in satoshis) and something called a scriptcode. For now, we will just store that data as an empty byte array, we won't need it until later.",
-        paragraph_five: `We also need a <span className="text-green"> serialize() </span> method that returns a byte array according to the specification. This is how the transaction is actually sent between nodes on the network, and how it is expressed in a block:`,
-        heading_two: 'Outpoint',
+        paragraph_nine: `You can see an example of this in the <span className="text-green p-1 font-mono bg-[#00000033] m-1 text-base">from_output()</span> method where it handles the txid argument.`,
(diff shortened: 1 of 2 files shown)


benalleng:
Making a note that this relates to one of our checkboxes!
https://github.com/saving-satoshi/saving-satoshi/issues/930#issuecomment-2101034899

benalleng:
Do you think we maybe should move some of this to a hint modal? its a lot of text to have on a read-only page

satsie:
@benalleng That's fair. I'm happy to move over anything you or anyone else specifically wants over in the hints section. 

Though I feel that precisely because this is a read only exercise, it's not that big of an ask to have the user read up on inputs, outpoints, endianness, and serialization. Nothing new was really added, I just made an effort to clarify what was already there. For example, I think it would be a loss if we didn't at least identify the term endianness since it's something not all developers outside of the Bitcoin industry have to work with. You could go your whole career as a Java dev and never learn what it means.

Is there a worry that if there is too much text the player won't read anything at all? Some additional context is I am writing a detailed curriculum for Art that lists what each chapter covers. I think it would be beneficial for some of these things to be front and center. It's tough to balance education with fun though, and I think any updates I propose to the hard mode of this challenge will not be this detailed (or should they be? should hard mode have more information than easy?). I guess worst case scenario is the player just doesn't read it all, or we can add additional, non coding screens, though that seems like a disproportionately large amount of work for what is gained.

i18n/locales/en.ts
```diff
@@ -1343,22 +1343,36 @@ const translations = {
...
+          a: 'When we see a Bitcoin hash as a hexadecimal string, the data is actually',
           b: {
             text: ' reversed ',
             question: 'Why do we reverse hashes in bitcoin?',
             href: 'https://chat.bitcoinsearch.xyz/?author=holocat&question=why%2520are%2520hashes%2520reversed%2520in%2520bitcoin',
           },
-          c: ' when we see them as hexadecimal strings. When we accept hashes as strings from a user we must reverse the byte order before storing or transmitting them as raw bytes. This is why we reverse the byte order of the txid argument that is passed in here.',
+          c: '. This ordering of bytes is called ',
+          d: {
+            text: 'endianness',
```
pinheadmz, line 1368: Only numbers have endianness. It is not correct to say that a hash is big or little endian. Bitcoin block hashes actually come out of the hash function with the 0's on the *right side*. Satoshi decided to interpret that as a little-endian integer. Sometime later when the JSON-RPC was eventually added, the decision was made to display those block hashes backwards so that they looked more like conventional big endian integers.

So what we do to hashes (sipa would say) is "byte swap" or just "reverse" them.

pinheadmz (commented):
I'm not sure about the endianness changes (see comment) but I do like a deeper explanation of outpoint. Maybe the endianness / reverse-hash section can be reworked as a pop up hint or on the hints page

i18n/locales/en.ts
satsie: Endianness only applies to numbers. Endianness only applies to numbers. Endianness only applies to numbers. 

I need to get a tattoo of this!!!!! Ty for the feedback! To refresh my memory, I also dug up some old comments you had  made that basically reiterate what you said here. 

I'll update this PR to use "byte swap" or "reverse"

satsie:
Updated this PR to remove the incorrect application of "endianness" to hashes. Also added a ChatBTC tool tip for endianness at the end where we remind the user that integers in bitcoin are little-endian.

There is no existing help page for this screen because it's not an interactive exercise. If we decide to move anything out to a help page, can that be done in a separate PR?

i18n/locales/en.ts
```diff
@@ -1358,22 +1358,30 @@ const translations = {
         nav_title: 'The input class',
         heading: 'Looking at the Input class implementation',
         paragraph_one:
-          'Bitcoin transaction inputs always point to existing, unspent transaction outputs. Therefore, our Input class has a method <span className="text-green"> from_output() </span> which is used to construct an Input by passing in the output description:',
+          'Here we have code for two classes: an <span className="font-bold">Input</span> class and an <span className="font-bold">Outpoint</span> class.',
```
pinheadmz, line 1361: Maybe something like this so right at the top, the distinction is clear. Otherwise its so easy to mis-read 

```suggestion
          'Here we have code for two classes: an <span className="font-bold">Input</span> class and an <span className="font-bold">Outpoint</span> (not "output"!) class.',
```

i18n/locales/en.ts
```diff
@@ -1358,22 +1358,30 @@ const translations = {
...
+        paragraph_eight: {
+          a: 'Hashes in bitcoin are',
           b: {
-            text: ' reversed ',
+            text: ' reversed',
             question: 'Why do we reverse hashes in bitcoin?',
             href: 'https://chat.bitcoinsearch.xyz/?author=holocat&question=why%2520are%2520hashes%2520reversed%2520in%2520bitcoin',
           },
-          c: ' when we see them as hexadecimal strings. When we accept hashes as strings from a user we must reverse the byte order before storing or transmitting them as raw bytes. This is why we reverse the byte order of the txid argument that is passed in here.',
+          c: `. When a hash is provided in hexadecimal format, the byte order must be reversed before storing or transmitting the data as raw bytes.`,
```
pinheadmz, line 1380:
```suggestion
          c: `, but only when presented to or entered by a user. When a hash is provided in hexadecimal format, the byte order must be reversed before storing or transmitting the data as raw bytes.`,
```

pinheadmz (approved):
Great! thanks.

i18n/locales/en.ts
satsie: Ah good point!!! It's funny how normal these small distinctions become to us after a while :)

i18n/locales/en.ts
```diff
@@ -1358,22 +1358,30 @@ const translations = {
...
+          'Inputs come from unspent transaction outputs. If you provide the description of an output to the <span className="text-green p-1 font-mono bg-[#00000033] m-1 text-base">from_output()</span> method, it will create an instance of the Input class:',
+        paragraph_three: 'The first two arguments are:',
+        paragraph_four:
+          '1. <span className="font-bold">txid: </span>the ID of the transaction that created the output, and',
+        paragraph_five:
+          '2. <span className="font-bold">vout: </span>the index of the output in the transaction\'s entire list of outputs',
+        paragraph_six:
+          'Together, these two pieces of information make up an <span className="font-bold">Outpoint</span>. Eventually we will pass in the txid and vout values that came from executing the <span className="text-green p-1 font-mono bg-[#00000033] m-1 text-base">listunspent</span> command in the previous exercise. ',
+        paragraph_seven:
+          "The second two arguments are the value of the output we want to spend (in satoshis) and something called a scriptcode. That data is not needed until later so let's temporarily store it as an empty byte array.",
```
satsie, line 1372:
```suggestion
          "The second two arguments are the value of the output we want to spend (in satoshis) and something called a scriptcode. That data is not needed until later so let's temporarily use an empty byte array.",
```

i18n/locales/en.ts
satsie: Thanks for catching this!

Issue #1070, opened 2024-08-30
Chapter 6, put-it-together-3-normal: stale lesson name in instructions


satsie:
In Chapter 6's put-it-together-3-normal challenge, the instructions mention something called "Populate the Input". I couldn't find it in the list of challenges for this chapter. Is this an old challenge name? Is it supposed to be a link? An easy fix could be to replace it with "an earlier exercise".

![Screenshot from 2024-08-30 14-51-25](https://github.com/user-attachments/assets/c1f84da6-60e1-49db-950a-bda8bdeee91e)

benalleng:
I just changed the text to be "(the UTXO we identified in <ins>The input class</ins>)"

satsie:
That works for me. Closing!

PR #1073, opened 2024-08-30
Fix stale translations for chapter 6

### LINKED ISSUE #1069: Stale translation file in CH6PUT1_NORMAL
Stale translation file on the success message for `chapter-6/put-it-together-1-normal`

image

### LINKED ISSUE #1072: Chapter 6: Duplicate challenge names for hard mode
When I toggle hard mode on for chapter 6, these are the challenge names. Are they supposed to be duplicates?

Screenshot from 2024-08-30 15-20-01

Normal mode:
- put-it-together-1-normal, "Build the transaction digest"
- put-it-together-2-normal "Build the witness class"
- put-it-together-3-normal "Sign the transaction"

Hard mode:

- put-it-together-1-hard, "Build the witness class". This is one challenge that is not in normal mode.
- put-it-together-2-hard, "Build the witness class". I think this should be "Serialize a transaction" or "Finish the transaction class" because it has the user writing the same serialize() method that is part of put-it-together-3-normal.
- put-it-together-3-hard, "Sign the transaction". This challenge has the user finishing the transaction digest method, the same as put-it-together-1-normal, but without all the hints in the code comments. This is a great way of turning a normal challenge into a hard one.
- put-it-together-4-hard, "Sign the transaction". I think this should be "Implement the signing algorithm" because it has the player writing the same ECDSA code as put-it-together-2-normal
- put-it-together-5-hard, "Populate the witness". Same as the remaining steps in put-it-together-2-normal (call methods written earlier, encode the signature to DER, add the sighash type, create the witness item and add it to the stack)
- put-it-together-6-hard, "Putting it all together". I can't access this one but it appears to be the second (easy) half of put-it-together-3-normal, the part where you take some funds out of the change address to pay the mining fees. In put-it-together-3-normal, the player has to write a transaction serialization method, but in hard mode this was already done as part of put-it-together-2-hard.

Is this the final version of chapter 6 hard mode? There are a lot of similarities between normal and hard mode, and in some ways hard mode is easier because it's nearly the same amount of work, just divided into smaller steps. Not sure if this is the plan, but if the intention is to remove the solution manual from hard mode, it would be a nice way to differentiate it from normal mode.

benalleng:
Closes #1069 and closes #1072 

also addresses #1070

### DIFF
--- a/content/lessons/chapter-6/hard/in-out-4.tsx
+++ b/content/lessons/chapter-6/hard/in-out-4.tsx
@@ -230,7 +230,7 @@ class Input:
           {t('chapter_six.in_out_four.hard.paragraph_one')}
         </Text>
         <CodeExample
-          className="mt-4 font-space-mono"
+          className="mt-4 whitespace-break-spaces font-space-mono"
           code={`from_output(txid: str, vout: int, value: int, scriptcode: bytes)`}
           language="shell"
         />
--- a/content/lessons/chapter-6/normal/put-it-together-1.tsx
+++ b/content/lessons/chapter-6/normal/put-it-together-1.tsx
@@ -383,7 +383,7 @@ class Transaction:
         config={config}
         saveData
         lessonKey={metadata.key}
-        successMessage={t('chapter_six.put_it_together_two.success')}
+        successMessage={t('chapter_six.put_it_together_one.normal.success')}
       >
         <LessonInfo>
           <Text className="font-nunito text-2xl font-bold text-white">
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -1424,7 +1424,7 @@ const translations = {
       },
       hard: {
         title: 'The ins and outs',
-        nav_title: 'The input class',
+        nav_title: 'Build the input class',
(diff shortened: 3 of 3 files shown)


satsie (approved):
Looks good! RE: #1072 I had some questions about the differences between normal and hard mode in chapter 6. I was in the process of updating the ticket when you were working on this. Let me know if you want me to break that out to a separate issue. It may or may not be actionable but I wanted to point out that currently hard mode is not very much different from normal mode.

satsie (approved):

benalleng:
> I wanted to point out that currently hard mode is not very much different from normal mode.

Yeah, we initially used this chapter as the first chapter to add a normal and hard mode because we released chapter 6 but then realized it was too difficult/tedius so we reverted to what is now normal mode but then added hard mode back in by just grabbing the old code from the git history of this chapter

benalleng:
Feel free to make another issue based on ways you think normal and hard mode could/should be different!

satsie:
will do. at minimum we could just make normal mode easier. This one's tough because the challenges are not mapped 1-to-1

PR #1084, opened 2024-09-16
Home page image update

GBKS:
New home page image with the new font baked in. Ensuring the full text in the image is visibly at various screen aspect ratios is not simple. I ended up mapping out lots of aspect ratios in design and then creating multiple images, and using some Javascript to then load the right image.

My coding approach is probably totally wrong, since I am not very familiar with React. There currently is an ugly effect where the page loads with the default image, and then only refreshes with the correct image. That should be improved so there is no "flash of wrong image".

For the new logo in the header, I simply pasted the SVG code in the Topbar component. There's probably also a more complicated way of doing this, by having it a file somewhere and pulling it in via React magic.

Let's first take a look and see if we all like this image. Then we can fix any code stuff if we do.

🧚‍♀️[Check the preview](https://saving-satoshi-git-fork-gbks-feature-home-2e2f35-savingsatoshi.vercel.app)🧚🏼‍♂️


components/Hero.tsx
```diff
@@ -14,25 +16,51 @@ export default function Hero() {
   const [account] = useAtom(accountAtom)
   const currentChapter = useAtomValue(currentChapterAtom)
 
+  const [imageMedia, setImageMedia] = useState('/assets/images/main-image.jpg')
```
benalleng, line 19: ahh i see now, we're talking about this

components/Hero.tsx
```diff
@@ -14,25 +16,51 @@ export default function Hero() {
...
-        src="/assets/images/main-image.jpg"
+        src={imageMedia}
         alt="Saving Satoshi"
         fill
         quality={100}
         loading="eager"
         priority
         className="-z-20 h-full w-full object-cover"
+        objectPosition="50% 34%"
       />
```
benalleng, line 58: I did some exploration with this and It seems like If we want to have the responsive image selection to choose different images as opposed to simply selecting different quality we might want to opt for a `<picture>` element like so
```
      <picture className="absolute -z-20 h-full w-full object-cover hero-image-position">
        <source srcSet="/assets/images/main-image-mobile-tall.jpg" media="(max-aspect-ratio: 3/5)"/>
        <source srcSet="/assets/images/main-image-mobile.jpg" media="(max-aspect-ratio: 1)"/>
        <source srcSet="/assets/images/main-image-tablet.jpg" media="(max-aspect-ratio: 13/10)"/>
        <source srcSet="/assets/images/main-image-wide.jpg" media="(min-aspect-ratio: 2)"/>
        <img
          loading="eager"
          src="/assets/images/main-image.jpg"
          alt="Saving Satoshi Landing image"
          className="absolute -z-20 h-full w-full object-cover hero-image-position"
        />
      </picture>
```

GBKS:
Closing this one. #1085 is a cleaner solution to the same thing.

PR #1085, opened 2024-09-16
Home page image refresh trying a picture element instead of next/image

benalleng:
This uses a standard `<picture>` html element instead of next/image for responsive sizing based on aspect ratio

[Preview](https://saving-satoshi-git-fork-benalleng-feature-2db575-savingsatoshi.vercel.app/)


GBKS:
That is a cleaner solution on the HTML/JS side, the downside is that we don't get the automated asset pipeline and compression that the next/image approach gives us. Loading both your and my PRs in the same browser at the same size, the next/image approach loads an optimized 1.1MB image, while this PR loads the original 3.5MB image.

Is there maybe a way to tap into that asset pipeline for the picture element so we get the best of both worlds?

0tuedon:
do we need to change the favicon too @GBKS

benalleng:
I think the favicon is ok for now, we did just change that one pretty recently

GBKS (approved):
ACK. Looks great. Much better technical solution than my PR.

Issue #1111, opened 2024-10-07
Change the Mining Reward Amount from the 51% Attack


satsie:
This is a high priority ticket and something we'd like to do before TABConf. In chapter 3 you and your friends pool your hashpower together to overcome Bitrey's 51% attack. The current reward for all that mining is 6.1 BTC. To avoid any confusion with the meme, we are changing that number to 7.36 BTC.

These are the screens in chapter 3 that need updating:

https://savingsatoshi.com/en/chapters/chapter-3/split-1
![Screenshot from 2024-10-07 12-14-02](https://github.com/user-attachments/assets/29a7e0cb-fba1-4de7-badc-8bd4dd065b8c)

https://savingsatoshi.com/en/chapters/chapter-3/split-2
We have the user getting 21.74% share of the rewards. By some stroke of luck, 21.75% of our original 6.1 BTC is NOT 1.61! So that makes our lives easier! I worked backwards from this to get our new total reward of 7.36 BTC, where 21.74% works out to 1.61 so no need to make any updates to the challenges in chapter 4 :)
![Screenshot from 2024-10-07 12-12-27](https://github.com/user-attachments/assets/e15f8359-311a-4113-a675-28689d36f103)

satsie:
I can take this ticket later this week but if anyone can get to it before me, feel free to grab it!

benalleng:
Closed by #1114

PR #1114, opened 2024-10-08
Chapter 3: Fix mining reward amount

satsie:
This PR addresses issue https://github.com/saving-satoshi/saving-satoshi/issues/1111. I was happily surprised that fixing it only required a few small changes.

### DIFF
--- a/content/lessons/chapter-3/split-2.tsx
+++ b/content/lessons/chapter-3/split-2.tsx
@@ -41,6 +41,7 @@ export default function Split2({ lang }) {
 
   const TOTAL_BLOCKS = 100
   const BLOCK_RATIO = 40
+  const TOTAL_REWARD = 7.41 // The total reward for mining this block
 
   const PROTAGONISTS = [
     {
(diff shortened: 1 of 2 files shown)


Emmanuel-Develops (commented):
hello @satsie, it seems the share was calculated based off `blocks found` rather than `partial solutions` which explains the discrepancies you found.

percentage-block-found


### block found percentage
i.e using (block found percentage) we have it currently as shown in the image above
- user (You): 26.32% of 6.1 -> 1.61
- Mining Maniacs: 15.79% of 6.1 -> 0.96
...
_adding up all the shares give 6.1_

### partial solutions
Following partial solutions instead as you correctly pointed out:
For the user (You) to have `1.61` from a share split where they contributed `21.74%`, the total value to be shared should be `7.4057..` instead of `7.36`.

This however introduces another problem as the value for other miners in the pool have to be recalculated.
i.e we should have
- user (You): 21.74% of `7.4057..` -> 1.61
- Mining Maniacs: 20.05% of `7.4057..` =~ `1.48..`
...

I'm uncertain on what amount of rounding or decimals to use which is why there is `..` at the end

satsie:
ooooo @Emmanuel-Develops you are right! The total reward should be 7.41 BTC (rounded to 2 decimal places). I don't know what I was thinking when I initially changed it to 7.36. I must have been half asleep :laughing: 

I updated the rewards for the other miners to now be calculated using the partial shares (and not blocks found). They are as follows:
- you: 1.61 btc
- Mining Maniacs: 1.49 btc
- Coin Crunchers: 1.41 btc
- Hash Hoppers: 2.9 btc

Also attaching a screencap with the new numbers (btw the screencap you added to your comment is private and I wasn't able to view it):
Screenshot from 2024-10-09 23-31-56

I added up all the individual rewards and as expected, the sum is 7.41. Can you double check my work in this PR? Thank you!

Emmanuel-Develops:
> (btw the screencap you included in your comment is private and I wasn't able to view it):

oops, updated it.

> I added up all the individual rewards and as expected, the sum is 7.41. Can you double check my work in this PR? Thank you!

Nice!
I've looked through it and it all checks out

Emmanuel-Develops (approved):

satsie:
Amazing!! Tysm!

PR #1123, opened 2024-10-15
Revision of images from chapters 1-9

GBKS:
I've been working on another big image revision for some time now. There were some story and character changes, and some images simply weren't that great. It's better now, but still lots of work to do. Thanks to Stacie for those highly expressive pics of Vanderpoole.

- Lots of replacements for completely new images
- Some images are simply upscaled for better quality and detail
- Tweaking some image quality settings in a few files so they don't get so blurry during compression. File size is still good

I'll do a separate PR for chapter 10, since some of those pics are still in the works. Let's get these pics in shape for TABConf.


satsie (approved):
These look amazing Christoph! I still hope to help out more with the next iteration. Thanks for putting this massive PR together!

Issue #1146, opened 2024-10-24
Error displays on repls are being overwritten


benalleng:
Validation errors that are printed to the repl console are immediately overwritten by Image closed Errors as their docker container is closed.

benalleng:
Closed by #1147

PR #1147, opened 2024-10-24
Repl must not have a resulting payload to display image closed …

benalleng:
We were having the image removed error always overwrite the validation error when testing the site during the workshop.

[Preview]()


benalleng:
Closes #1146

PR #1155, opened 2024-10-30
Added Portuguese BR translation

kaiosatiro:
## Added Portuguese BR translation and reviewed up to chp 4
Hi,
So, I finished the Portuguese translation and am now reviewing 'on page'. 

I'll keep an eye on changes in the `saving-satoshi/i18n/locales/en.ts` file. I could translate everything in there. If there is more in other places, just let me know. I also tried to keep the same body format of the JS object in English, to track changes more easily.

I'm thinking now inviting some people to review with me.
Let me know if there is something wrong.

### DIFF
--- a/i18n/config.ts
+++ b/i18n/config.ts
@@ -6,6 +6,7 @@ export const i18n = {
     { locale: 'nl', label: 'Dutch' },
     { locale: 'hi', label: 'Hindi' },
     { locale: 'de', label: 'German' },
+    { locale: 'pt', label: 'Português'}
   ],
 } as const
 
--- a/i18n/locales/index.ts
+++ b/i18n/locales/index.ts
@@ -3,6 +3,7 @@ import es from './es'
 import nl from './nl'
 import hi from './hi'
 import de from './de'
+import pt from './pt'
 
 const locales = {
   en,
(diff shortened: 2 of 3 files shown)


benalleng:
Do you think we should worry about translating `Saving Satoshi` or is it ok to leave that in English?

benalleng:
I am also exploring refactoring the translation to use `.json` like is more traditional for i18n stuff so we can use external tools to help us a little more like key matching etc. That would be a future thing though so don't worry about it in this PR.

kaiosatiro:
I don't think we should be worried about it now. In the future would be cool.
When you refactor into a `.json`, I can easily review the pt-br into it too.

kaiosatiro:
I finished the review of all the chapters.

benalleng (approved):

benalleng (approved):
Looks good! Didn't find any missing keys

kaiosatiro:
hey @benalleng. I realized that many resource links that are shared in the lessons have a Portuguese version, like Wikipedia or language documentation.  So I change the ones that have. Sry if I didn't think of this before.
Should I open another pull request?

Also, I noticed that it was not the last revised pt-br version that went online.

PR #1156, opened 2024-11-01
Retrieve State on Tab Switch

### LINKED ISSUE #1137: Switching sub-tabs (ie between "multisig" and "refund) should keep user progress
> When switching sub-tabs (ie between "multisig" and "refund) the inputs fields reset and i lose my work...

0tuedon:
Should close #1137


utils/data.ts
```diff
@@ -116,12 +116,12 @@ export const transactionTabs: TransactionData = {
       sats: '101,000',
     },
     output_0: {
-      sats: '99000',
+      sats: '1000',
```
benalleng, line 119: I think we didn't actually need to switch these

utils/data.ts
```diff
@@ -116,16 +116,54 @@ export const transactionTabs: TransactionData = {
...
```
benalleng, line 119:
``` suggestion
      sats: '99000',
```

benalleng (approved):

PR #1163, opened 2024-11-06
Translation linting

benalleng:
### TODO
- remove added npm package as it seems like the rule is pretty simple and I'd like to modify it slightly to allow some defaults like hex and certain characters


benalleng:
There will be a lot of warnings now that we have extended the `dirs` that the eslint covers, I did fix all the existing errors though!

PR #1165, opened 2024-11-08
Chapter 10 image update

GBKS:
Updating all the images in chapter 10. Still not 100%, but a solid update, if I may say so myself.


benalleng (approved):
Looks great!

PR #1174, opened 2024-11-12
ch10: misc updates based on qa

satsie:
This PR contains updates I've made after doing QA on the first two parts of chapter 10 ("Celebration" and "The initial funds"). It's mostly copy changes in an effort to clarify some instructions. I will have more updates to chapter 10 in a future PR.

### DIFF
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -2785,7 +2785,6 @@ Stack Hint: To spend before the secret is revealed, Vanderpoole uses his signatu
       hint_one_a: `You and Laszlo want to work together in the future to spend output 0 with <br/> `,
       hint_one_b: `<span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base whitespace-nowrap"> 0 SIG(LASZLO) SIG(YOU) </span>`,
       hint_two: `Don't forget to set aside miner fees`,
-
     },
     opening_a_channel_five: {
       title: 'Refund protection in multisig',
@@ -3047,9 +3046,11 @@ Stack Hint: To spend before the secret is revealed, Vanderpoole uses his signatu
       output_zero_sig: 'Output 0 Signature',
       output_one_sig: 'Output 1 Signature',
       sats_distribution:
-        'Laszlo should receive enough sats for the beer, and is not responsible for the 1000 sat miner fees.',
+        'Laszlo should receive enough sats for the beer, and is not responsible for the 1,000 sat miner fees. The fee comes out of your change.',
       output_script:
         'The locking script on the output should not contain any signatures (i.e. <span className="rounded-sm px-1.5 py-1 font-mono bg-[#0000004D] m-1 text-base whitespace-nowrap">OP_PUSH SIG()</span>). If this is part of your script, simply remove it.',
+      miner_fees_basic: `Don't forget to set aside 1,000 sats for miner fees.`,
+      miner_fees: `Don't forget to set aside 1,000 sats from your change for miner fees.`,
     },
   },
 
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -2528,7 +2528,7 @@ const translations = {
       nav_title: 'Multisig',
       heading: 'Multisig',
       paragraph_one:
-        'Multi-signature policies provide a list of public keys and a number of signatures required for a valid spend. It can be described as "m-of-n" meaning "m number of signatures are required from this list of n public keys". The public keys and the m and n values are typically included in the locking script and the spender only needs to provide the right number of signatures.',
+        'Multisignature policies provide a list of public keys and a number of signatures required for a valid spend. It can be described as "m-of-n" meaning "m number of signatures are required from this list of n public keys". The public keys and the m and n values are typically included in the locking script and the spender only needs to provide the right number of signatures.',
       paragraph_two:
         'Holocat appears with a pre-recorded message from Satoshi Nakamoto!',
       paragraph_three:
@@ -2540,7 +2540,7 @@ const translations = {
       heading:
         '<span className="flex items-center text-[#3DCFEF] w-fit rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-1">OP_CHECKMULTISIG</span>',
       subheading_one:
-        'Processes m-of-n multi-signature by following this algorithm.',
+        'Processes m-of-n multisignature by following this algorithm.',
       multisig_list_one:
         'Pop a single integer off the stack. This is the <span className="text-[#3DCFEF] rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-sm">n</span> value.',
       multisig_list_two:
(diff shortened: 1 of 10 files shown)


i18n/locales/en.ts
```diff
@@ -3022,7 +3025,9 @@ Stack Hint: To spend before the secret is revealed, Vanderpoole uses his signatu
       output_zero_sig: 'Output 0 Signature',
       output_one_sig: 'Output 1 Signature',
       sats_distribution:
-        'Remember that Laszlo should only be receiving enough sats for the beer. Also we need to set aside some sats for the fees incase our channel closes!',
```
satsie, line 3025: I removed the part about setting fees aside for channel closes before I realized this same hint is being displayed on all the chapter 10 resource pages. It doesn't belong in the first few exercises I've done so far, but I can see the need to bring it back for a resource page in a later exercise.

Most of the resource pages need to be filled in but I want to get through chapter 10 QA before taking a closer look.

i18n/locales/en.ts
```diff
@@ -2737,49 +2737,52 @@ Stack Hint: To spend before the secret is revealed, Vanderpoole uses his signatu
       nav_title: 'The off-chain transaction',
       heading_one: 'The off-chain transaction',
       paragraph_one:
-        "Let's create a valid Bitcoin transaction but instead of sending it to the network, we'll just send it directly to Laszlo! Laszlo should trade you a beer for this transaction because he knows he effectively has the money even though the transaction is not broadcasted or confirmed.",
-      paragraph_two:
-        "Remember it's the year 2140! A SHA-256 Stout cost 0.00010000 BTC.",
-      heading_two: 'Hints',
+        "Let's create a valid bitcoin transaction but instead of broadcasting it out to the network, we'll give it directly to Laszlo. In return, he should provide you with a beer because he can broadcast the transaction anytime he wants. Once it's confirmed and mined into a block, the money is his.",
+      paragraph_two: `Remember it's the year 2140! A <span className="italic">SHA-256 Stout</span> costs 0.00010000 BTC.`,
```
pinheadmz, line 2741: should be `0.00001000 BTC` I think -- 1,000 sats?

i18n/locales/en.ts
```diff
@@ -2737,49 +2737,52 @@ Stack Hint: To spend before the secret is revealed, Vanderpoole uses his signatu
...
       paragraph_two: `Do yourself a favor; before signing this, make a refund transaction to know you can recover your money."`,
     },
     updating_the_state_one: {
       title: 'The refund',
       nav_title: 'The refund',
       heading_one: 'The refund',
-      paragraph_one: `A new tab appears: <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base">Refund</span> which starts another TX template on the same screen, with an arrow from the <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base">Multi-Sig</span> output to this TX input.`,
+      paragraph_one: `A new tab appears: <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base">Refund</span> which starts another TX template on the same screen, with an arrow from the <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base">Multisig</span> output to this TX input.`,
```
pinheadmz, line 2785: i noticed this kind of copy later on in the chapter -- its my fault for writing the challenge this way, its a mix of actual copy for the user and "stage directions" for design. But if we act like the right side of the screen actually IS the user's ePhone Infinity, then it makes sense

i18n/locales/en.ts
pinheadmz: yeah i am thinking we should be consistent with fees for all txs now, which means we start with the "deposit" 101,000 sats, pay a fee on the funding tx so our open channel has 100,000 and then all commitment txs pay another fee, meaning the Laszlo + You balance should = 99,000 for the rest of the chapter

pinheadmz (commented):
These are really great updates, thanks. I left a few comments

i18n/locales/en.ts
satsie: Good point! I linked your comment in this related issue that I filed: https://github.com/saving-satoshi/saving-satoshi/issues/1175

i18n/locales/en.ts
satsie: Good catch! Fixed and force pushed [caf365a](https://github.com/saving-satoshi/saving-satoshi/pull/1174/commits/caf365a783a8215082e48cd4319f2aa1ae5fe854)

i18n/locales/en.ts
```diff
@@ -2737,49 +2737,52 @@ Stack Hint: To spend before the secret is revealed, Vanderpoole uses his signatu
...
       paragraph_two: `Do yourself a favor; before signing this, make a refund transaction to know you can recover your money."`,
     },
     updating_the_state_one: {
       title: 'The refund',
       nav_title: 'The refund',
       heading_one: 'The refund',
-      paragraph_one: `A new tab appears: <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base">Refund</span> which starts another TX template on the same screen, with an arrow from the <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base">Multi-Sig</span> output to this TX input.`,
+      paragraph_one: `A new tab appears on your ePhone Infinity: <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base">Refund</span> which starts another TX template on the same screen, with an arrow from the <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base">Multisig</span> output to this TX input.`,
```
satsie, line 2785: @pinheadmz Here's where I updated it to mention your ePhone Infinity :)

i18n/locales/en.ts
satsie: Oooo! I like the idea of mentioning the new tab is on your ePhone Infinity. I updated it as part of https://github.com/saving-satoshi/saving-satoshi/pull/1174/commits/5dbeca111b179538b786d1804a51e4080a6f9aae

satsie:
Thanks so much for the review @pinheadmz! I just finished addressing all your comments.

satsie:
FWIW I want to do another pass of chapter 10 after I finish. I think there are still things I did not catch and want to retest after the various QA updates (from myself and others) have been implemented.

satsie:
@pinheadmz just rebased this and resolved the conflicts. When you have time, may I trouble you for a re-review?

i18n/locales/en.ts
benalleng: You can decide where to delete these fee comments individually here.
https://github.com/saving-satoshi/saving-satoshi/blob/dev/content/resources/chapter-10/making-a-payment-2.tsx#L19

benalleng: that is just one resource page but it is in all of them

i18n/locales/en.ts
pinheadmz: This is great -- I think it would help to remind the user to deduce another 1,000 sat fee. The validation code does check for it (sats==99000 for correct answer)

i18n/locales/en.ts
pinheadmz: Its only incorrect after we buy a second beer, then it could just read "Laszlo should receive enough sats for the beer(s)" ;-) It is true he doesnt pay miner fees, but we could also mention that the miner fee for the commitment comes out of your change (get ready for advanced Saving Satoshi: anchor outputs and package relay!)

pinheadmz (approved):
Good stuff Stacie thanks

i18n/locales/en.ts
satsie: I addressed this by updating the help page to say

> Don't forget to set aside 1,000 sats from your change for miner fees.

i18n/locales/en.ts
satsie: Thanks all! Here are the updates I made to the help pages:

1. Kept the help page for the first exercise (opening-a-channel-2) empty because everything is in the instructions.
2. Updated the help page for the multisig excercise (opening-a-channel-4) to say `Don't forget to set aside 1,000 sats for miner fees.`
3. Updated all refund transaction exercises (updating-the-state-1, updating-the-state-3, updating-the-state-5) so the help page says: `Don't forget to set aside 1,000 sats from your change for miner fees.` I did this because it doesn't make sense to mention Laszlo and the beer(s) in those, and to clarify that the miner fees come from your change.
4. Updated the help page for the exercises where you actually buy the beer (making-a-payment-2, making-a-payment-5) to say `Laszlo should receive enough sats for the beer, and is not responsible for the 1,000 sat miner fees. The fee comes out of your change.`, which is more or less what this PR initially intended.
5. Updated the help page for the last exercise where you buy the second beer to just say `Don't forget to set aside 1,000 sats from your change for miner fees.`. Hopefully at this point the user gets it :)

PR #1185, opened 2024-11-21
feat: update code for chapter 2 scriptung challenge

benalleng:
This cleans up the answer and the lesson for chapter 2 scripting challenge

### TODO
- update copy where it no longer matches

### DIFF
--- a/content/lessons/chapter-2/scripting-2.tsx
+++ b/content/lessons/chapter-2/scripting-2.tsx
@@ -35,7 +35,7 @@ export default function Scripting2({ lang }) {
 const min = 1;
 const max = 100000000;
 const randomNumber = Math.floor(Math.random() * (max - min + 1)) + min;
-const testHash = findHashFromNonce(randomNumber) ?? "Be sure your function returns a value."
+const testHash = hashNonce(randomNumber) ?? "Be sure your function returns a value."
 console.log(testHash.toString())
 console.log("KILL")`,
     defaultFunction: {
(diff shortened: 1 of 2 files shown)


pinheadmz:
Screenshot 2024-11-26 at 11 07 59 AM

Comment indentation is a little awkward here

pinheadmz:
So, this is the solution hint:

```python
        hash = hashlib.sha256(str(nonce).encode()).hexdigest()
```

My issue with this is that we don't explicitly say the nonce is converted to a string using ASCII.
For example if `nonce = 0` then that "0" gets interpreted in ASCII as the byte `0x30` and that byte is what is actually passed into the hash function.

Before looking at the hint I used `bytes(nonce)` which also is a mistake but kinda should work. This returns a byte array of length `nonce`. So it will try `0x00` then `0x0000` then `0x000000`, etc. eventually there will be an all-zero byte array of some length that does pass the challenge but I think we time out before we get there.

Probably the MOST correct line of code would use `nonce.to_bytes(4)` since the nonce in the bitcoin block header is a 32-bit field. I don't think it matters if `byteorder="little"` or `byteorder="big"` is specified necessarily, both should pass the challenge in equal time, theoretically. This gives the best representation:


```
>>> (0).to_bytes(4).hex()
'00000000'
>>> (1).to_bytes(4).hex()
'00000001'
>>> (65530).to_bytes(4).hex()
'0000fffa'
```

This also means the copy should specifically state that the nonce should be interpreted as 32 bit integer and those four bytes are what is explicitly passed into the hash function

PR #1208, opened 2024-11-27
ch 10: qa updates to part 3, the refund

satsie:
These are some updates I made after QA'ing "The refund" section of chapter 10 last week. I played through the updated final challenge and it looks good to me. Everything works as expected. Nice touch that the hints refresh when you go from working on Laszlo's commitment transaction to yours.

### DIFF
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -2864,45 +2864,47 @@ Stack Hint: To spend before the secret is revealed, Vanderpoole uses his signatu
       heading_one: 'Making a payment',
       paragraph_one: "Let's recap:",
       list_one:
-        'You sent 100,000 satoshis to a 2-of-2 multisig output between you and Laszlo',
-      list_two: 'You have, offline, a transaction that spends that output.',
-      list_three: 'That offline output enables EITHER',
-      list_three_sub_one: 'You to get all your money back after 700 blocks',
+        'You sent 100,000 satoshis to a 2-of-2 multisig output between you and Laszlo.',
+      list_two:
+        'You have, offline, a refund transaction that spends that output.',
+      list_three:
+        'That refund transaction specifies two options for how the funds can be spent:',
+      list_three_sub_one:
+        'You to get all 100,000 satoshis back after 700 blocks, or',
       list_three_sub_two:
         'Laszlo gets all the money if he gets the private key <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base">REVOCATION_YOU_1</span> from you',
       list_four:
-        'Laszlo has already signed it, and you can sign it whenever you want to broadcast it.',
+        'Laszlo has already signed this refund transaction, and you can sign it whenever you want to broadcast it.',
       list_five:
-        'Once Laszlo signed the offline refund (child) transaction, you were safe to sign and broadcast the parent transaction that sends the funds to the 2-of-2 multisig.',
+        'Once Laszlo signed this refund (child) transaction, you were safe to sign and broadcast the parent transaction that sends the funds to the 2-of-2 multisig.',
       paragraph_two:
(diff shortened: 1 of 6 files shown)


i18n/locales/en.ts
```diff
@@ -2969,16 +2980,18 @@ Stack Hint: To spend before the secret is revealed, Vanderpoole uses his signatu
...
+      paragraph_one: `The night is still young, and you're not flying your Budgetcopter home. Got time for another beer? Now that you and Laszlo have your asymmetric transactions and a flow worked out with revocation keys, let's run the protocol again.`,
+      paragraph_two: `The two commitment transactions representing the current state where you have paid Laszlo 1,000 sats for a beer are on your ePhone Infinity screen.`,
+      heading_two: `Instructions`,
+      list_one: `Update the amounts and output scripts for Laszlo's new commitment transaction`,
       list_two: "Sign Laszlo's transaction and send it to him",
-      list_three: `Laszlo will revoke his last state by sending you <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base whitespace-nowrap">REVOCATION_LASZLO_2</span>`,
+      list_three: `Update your commitment transaction: Laszlo will revoke his last state by sending you <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base whitespace-nowrap">REVOCATION_LASZLO_1</span>. Update the amounts and output scripts in your commitment transaction.`,
```
benalleng, line 2988: wait is this the second revocation key or the first?

i18n/locales/en.ts
```diff
@@ -2969,16 +2980,18 @@ Stack Hint: To spend before the secret is revealed, Vanderpoole uses his signatu
...
       list_four: 'Send your transaction to Laszlo so he can sign it',
-      list_five: `Once you have Laszlo's signature, send him your <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base whitespace-nowrap">REVOCATION_YOU_3</span> `,
+      list_five: `Once you have Laszlo's signature, send him your previous revocation key, <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base"> REVOCATION_YOU_2 </span> `,
```
benalleng, line 2990: same here should it be 3 or 2?

i18n/locales/en.ts
satsie: It's the first. He gives you his previous revocation key, `REVOCATION_LASZLO_1` to prove that he won't try to settle/close the payment channel with the previous state. You might be getting thrown off because he has no reason to broadcast the prior commitment transaction. The newer state where he has 2,000 sats is more favorable to him.

I think this set of instructions could use some improvements (possibly in a follow up PR after the release) but the change I made here is just for correctness.

i18n/locales/en.ts
satsie: It should be `REVOCATION_YOU_2`. The copy on here is describing the act of sending the revocation keys for the previous commitment transactions, which invalidates them and makes the new set of commitment transactions in this exercise the only ones.

I think you're looking for copy that tells the user about the new set of revocation keys, `REVOCATION_YOU_3` and `REVOCATION_LASZLO_2` which would be an improvement since those are the revocation keys you need to update the commitment transactions with. Unfortunately, that would require a larger rework of these instructions. Based on my track record with this kind of stuff, a larger change to this would not be approved before the deadline we are trying to hit for chapter 10 so I'd prefer to tackle it after we release.

benalleng (approved):
Ahh ok I was definitely confused by the text 
"revoke his last state by sending you REVOCATION_LASZLO_1" specifically "sending you" perhaps "revealing to you". Not sure, otherwise LGTM!

PR #1220, opened 2024-12-03
Boss banner

benalleng:
Adds the Chaincode Boss banner as a link to the application page


satsie (approved):
I tested the preview and it looks great!

PR #1238, opened 2024-12-06
Websocket error catching

benalleng:
I added a new error catch in the websocket connection to make sure we are always displaying an error if the websocket doesn't receive a payload before closing. As well as some minor styling fixes for mobile


satsie (approved):

satsie:
Trying to understand this better, so the repl times out and because of that we close the websocket?

benalleng:
the websocket closes no matter what after the "end" payload received by the server, however we have been repeatedly been getting errors where the "output" type is never sent before the timeout time set by the server which automatically sends the "end" repsonse resulting in nothing to display to the user so we need to catch that. Initially I was looking only for errors that said "[Image] system remover container xxxx" but it seems like that is not always the case. (This docker thing makes me want to pull my hair out lol)

Issue #1138, opened 2024-10-21
Chapter 10 prefill the output field


benalleng:
> Since the scripts are getting more complicated, i think it would be nice if the last script we wrote was still there in the output field so i just have to add the time lock, instead of having to debug the entire script again

benalleng:
Do we want this on all the lessons? I am a little worried about how much of a redesign this would need.

satsie:
Implementing this would help address feedback we have gotten from several testers: It would be nice to be able to copy and edit a previous answer so it can be submitted as solution to the current challenge. However, as discussed in #1182, persisting user answers would require overriding default challenge code with user code. That doesn't seem to be worth the complexity as we can achieve the same thing (users not having to rewrite the same code a bunch of times) with some default challenge code that happens to be the answer to the previous challenge.

satsie:
> Do we want this on all the lessons? I am a little worried about how much of a redesign this would need.

If everyone's challenges are prefilled with the same code, as opposed to individual user code, does this make it less of a redesign?

benalleng:
My follow up now is what page to prefill and what to prefill them with?
here are the individual challenge pages
- [ ] opening a channel - 2
- [ ] opening a channel - 4
- [ ] updating the state - 1
- [ ] updating the state - 3
- [ ] updating the state - 3
- [ ] making a payment - 2
- [ ] making a payment - 5
- [ ] making a payment - 8

satsie:
That list makes sense to me. For a few of the challenges I see that the copy idea you proposed earlier probably does make sense, specifically for these two screens:

**making-a-payment-2**: copy to remind the user to switch to the refund tab to see the previous refund/commitment transaction. (unrelated note: I think it's confusing that in this point of the game we use both terms (refund and commitment) but it makes sense considering the user learned of it as a refund before they were introduced to commitment transactions in making-a-payment-1)

**making-a-payment-5**: copy to tell the user they can look at the commitment_you tab to see the script in their commitment transaction

I'm happy to pick this up if no one else gets to it before me.

benalleng:
| making-a-payment-2: copy to remind the user to switch to the refund tab to see the previous refund/commitment transaction. (unrelated note: I think it's confusing that in this point of the game we use both terms (refund and commitment) but it makes sense considering the user learned of it as a refund before they were introduced to commitment transactions in making-a-payment-1)

Using this code, we can change the name of "refund" to "initial commitment / commitment"
[...]

PR #1239, opened 2024-12-06
Add code defaults for different refunds

benalleng:
I made a list of the particular pages I thought would be good to add prefilled data to explained here #1138


ui/lesson/TransactionsChallenge/OutputScript.tsx
```diff
@@ -187,6 +196,7 @@ const OutputScript: FC<IOutput> = ({
   }
 
   const handleScriptChange = (event) => {
+    console.log(event.target.value)
```
satsie, line 199: Was this added by mistake?

i18n/locales/en.ts
```diff
@@ -3064,7 +3064,7 @@ Stack Hint: To spend before the secret is revealed, Vanderpoole uses his signatu
       payment: 'This is your off-chain payment to Laszlo.',
       multisig:
         'This transaction will fund a multisig output between you and Laszlo.',
-      refund:
+      refund_0:
```
satsie, line 3067: Does anything else need to be changed to account for this updated key name?

satsie (commented):
Tested it an it looks good. Does the prefilled code come from the solution on the previous challenge's hints/resources page? 

Somehow that casing issue is back :sob: 

Screenshot from 2024-12-06 13-44-22

i18n/locales/en.ts
benalleng: No, it was an error as a bare translation key with no string before

satsie (approved):

utils/data.json
```diff
@@ -53,7 +53,7 @@
     },
     "output_0": {
       "sats": "99000",
-      "script": "T1BfUFVTSCBQVUJLRVkoWW91KSBPUF9DSEVDS1NJRw=="
+      "script": "T1BfUFVTSCBQVUJLRVkoWU9VKSBPUF9DSEVDS1NJRw=="
```
benalleng, line 56: this was not capitalized correctly, fixed in this string

Issue #1240, opened 2024-12-06
Create Copy to encourage users to look at other tabs on complex commitment code


benalleng:
It would be nice to have some copy to excourage users to use the previous tabs we have for them to referrence

https://github.com/saving-satoshi/saving-satoshi/issues/1138#issuecomment-2523888174

https://github.com/saving-satoshi/saving-satoshi/issues/1138#issuecomment-2523946416

PR #1247, opened 2025-01-10
Chapter 10 copy: encourage player to revisit previous solutions

satsie:
This PR addresses issue https://github.com/saving-satoshi/saving-satoshi/issues/1240

I added copy to encourage users to look at their previous solutions since we don't do any kind of pre-filling here for the code. I also renamed the 'Refund' tab to 'Initial Commitment' every time the user sees it after learning about the term "commitment transactions" in making-a-payment-1. Probably makes more sense to call this the "Initial commitment" from the start, and rename it to "Refund" for updating-the-state-5 but that doesn't seem like it's worth the effort.

@benalleng hoping I implemented things properly based on your comment here: https://github.com/saving-satoshi/saving-satoshi/issues/1138#issuecomment-2523946416 Have a look at it and LMK.

Test links:

### DIFF
--- a/ui/lesson/TransactionsChallenge/index.tsx
+++ b/ui/lesson/TransactionsChallenge/index.tsx
@@ -200,6 +200,18 @@ const TransactionChallenge: FC<ITransactionProps> = ({
         return { id: tab, text: 'commitment' }
       }
 
+      // Start calling the refund tab "initial commitment" now that the player
+      // has learned about commitment transactions
+      const postCommitmentLessons = new Set([
+        'CH10MAP2',
+        'CH10MAP5',
+        'CH10MAP6',
+        'CH10MAP8',
+      ])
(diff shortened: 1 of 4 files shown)


ui/lesson/TransactionsChallenge/index.tsx
```diff
@@ -200,6 +200,18 @@ const TransactionChallenge: FC<ITransactionProps> = ({
...
+      const postCommitmentLessons = new Set([
+        'CH10MAP2',
+        'CH10MAP5',
+        'CH10MAP6',
+        'CH10MAP8',
+      ])
+      if (postCommitmentLessons.has(progressKey) && tab === 'refund_2') {
+        return { id: tab, text: 'initial commitment' }
+      }
+
```
benalleng, line 214: Nice! good idea!

benalleng (approved):
LGTM! Nice addition!

PR #1248, opened 2025-01-28
German translation

Walpurga03:
Update German translations: Change 'holocat' to 'Holokatze' in text while preserving 'holocat' in URLs

### DIFF
--- a/i18n/locales/de.ts
+++ b/i18n/locales/de.ts
@@ -25,7 +25,7 @@ const translations = {
     subtitle:
       'Wir schreiben das Jahr 2139. In zwei Wochen wird der letzte Bitcoin geschürft. Seit Monaten tickt auf dem Satoshi Square eine Uhr herunter.',
     intro:
-      'Die Welt wartet auf den letzten Block. Dann kommt das Netzwerk plötzlich zum Stillstand.<br><br>Sie erhalten ein Holocat von jemandem mit dem Namen Satoshi Nakamoto. (Es ist wie jedes andere E-Hologramm, aber dieses hat die Form einer Katze). Sie öffnen das Holocat, indem Sie neugierig auf seine Nase klopfen, um zu hören, was es zu sagen hat ...',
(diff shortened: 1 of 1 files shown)


satsie:
Thank you for putting this together! I just assigned it to @GBKS for review.

GBKS:
Heads up, I am happy to help with this review, but it's a huge amount of text so I need to set aside a solid amount of time to go over it all. For the future, it might be easier to split things up in per-chapter PRs to make it easier to go through.

May I ask what the process was for the translation?

Walpurga03:
Yes, it's a very long text to translate, and I decided not to translate it line by line, as it would have been too much work. Instead, I wrote a program to format the original text and make the links, line breaks, simple quotes, and other elements consistent. Then, I counted the links and nesting to ensure everything was correct. The translation itself was done using the DeepL AI, which usually produces good results. After the translation, I went through the text again and checked the links and nesting to ensure everything was correct. I also compared the count of links and nesting in the original text with the count in the translated text to ensure that no elements were lost or duplicated. While playing through "Saving Satoshi" in German, I reviewed my translation again and made it up to chapter 6. I made sure that everything makes sense and that the translation is correct. If I can help in any way, I'm happy to assist, whether it's reviewing further texts or helping with other projects you need.

Walpurga03:
Yes, I think that's a great idea, Chris! Your suggestion to split the translation into per-chapter PRs is very helpful. I could start the translation from scratch and we could work our way through the game step by step. If I proceed in this way, it might be better if we stay in touch, for example over Signal or Discord, to make sure everything is correct. Then we can work together to improve the translation!

satsie:
> Yes, I think that's a great idea, Chris! Your suggestion to split the translation into per-chapter PRs is very helpful. I could start the translation from scratch and we could work our way through the game step by step. If I proceed in this way, it might be better if we stay in touch, for example over Signal or Discord, to make sure everything is correct. Then we can work together to improve the translation!

We have a Saving Satoshi channel in the Bitcoin Design Discord that sounds perfect for this! https://discord.gg/eBGRaCSjg5

PR #1249, opened 2025-02-21
copy: swap out beer for Lightning Lemonade

satsie:
We got some feedback from a high school teacher that the beer theme in chapter 10 makes the game a non-starter for him. I replaced the references in chapter 10 to beer with "Lightning Lemonade" (Mempool Mocktail and Taproot Tea were runners up)

I went through all of chapter 10 and the change actually seems to work well with the images because I'd expect there to be something electric about a Lightning Lemonade and all the drinks Laszlo makes are neon. Even if that wasn't the case, I interpret the images as Laszlo making other drinks while he waits for the payment channel to open.

The only image that we may want to consider updating is this one, which I am happy to attempt if necessary. This image could also be of them holding root beers or sodas so maybe we don't need to change it at all.

Screenshot from 2025-02-21 15-30-35

### DIFF
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -2738,7 +2738,7 @@ Stack Hint: To spend before the secret is revealed, Vanderpoole uses his signatu
       paragraph_one: `Since “Satoshi” initially reached out, you’ve only had to work with on-chain transactions. That’s how you claimed your mining rewards and how you transferred funds to Mika 3000. However, these on-chain transactions aren’t ideal for frequent, everyday use because block space is limited and miner fees can vary.`,
       paragraph_two: `How is it then, that people are able to use bitcoin daily in a way that is scalable? The answer is off-chain payments, something you have probably seen a million times: An orange cab ejecting its passengers and zooming off, someone buying dinner from a halal cart, a kid exiting a store with her holodog and a carton of digimilk. And so on. Money’s use cases are nearly infinite.`,
       paragraph_three: `While bitcoin isn’t designed to handle this much activity on its own, with off-chain payments, it can.`,
-      paragraph_four: `—YOU: “That’s not a problem at all, Mika 3000. I can handle it. I’ll prove it by using an off-chain payment to purchase a beer from Laszlo.”`,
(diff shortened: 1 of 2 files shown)


i18n/locales/en.ts
```diff
@@ -2972,27 +2972,27 @@ Stack Hint: To spend before the secret is revealed, Vanderpoole uses his signatu
           '<span className="font-semibold">Output 1:</span> 98,000 sats to you',
         list_four: `<span className="font-semibold">Miner fees:</span> 1,000 sats`,
       },
-      paragraph_eight: `All three of these transactions are signed and valid, but Laszlo hasn't handed you a beer yet. Why not? Only one thing left to do, send Laszlo your previous revocation key, <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base">REVOCATION_YOU_1</span>!`,
+      paragraph_eight: `All three of these transactions are signed and valid, but Laszlo hasn't handed you a Lightning Lemonade yet. Why not? Only one thing left to do, send Laszlo your previous revocation key, <span className="rounded-sm px-1.5 py-1 h-[28px] font-mono bg-[#0000004D] m-0.5 text-base">REVOCATION_YOU_1</span>!`,
       success_bar: 'Send REVOCATION_YOU_1 to Laszlo',
     },
     making_a_payment_seven: {
-      title: `🍺 Ahhhhhhh, nice.`,
+      title: `🍋🍹 Ahhhhhhh, nice.`,
```
benalleng, line 2979: lol I love it

benalleng (approved):
LGTM!

TimothyShimmin:
Found another reference to beer still in prod. Might want to check the pop out script debuggers in ch 10 challenges. 

> won’t broadcast the transaction after paying for your beer.

https://script.savingsatoshi.com/?lesson=CH10UTS5

Edit: 
This one is a bit more cheeky, but might still qualify:

> The night is still young, and you're not flying your Budgetcopter home. 

https://savingsatoshi.com/en/chapters/chapter-10/making-a-payment-8

Edit: 
This one is a better catch
>did you have one too many Mempool Margaritas?

https://savingsatoshi.com/en/chapters/chapter-10/outro-2

PR #1252, opened 2025-02-24
Minor image update

satsie:
Update chapter-10-outro-1 so the characters are holding Lightning Lemonades. This resolution is slightly smaller than the previous one, but it's what Midjourney gave to me when I asked it to upscale. Pending approval from Christoph.

Also include an update for chapter-5-intro-2 that I did a few months ago. Since this is a composite of two images, the original is relatively small (about 700px). I upscaled it to a width of 2048px using Canva. Let me know if this is okay.

Christoph, I don't have access to the main images Figma to upload these images, but I put them into my copy of the images Figma. Does that work for you?


GBKS:
Nice changes. I'd recommend resizing the images to keep to maintain the standard square 2880x2880 ratio and size we have, and also use [ImageOptim](https://imageoptim.com/) to optimize them before committing (to have almost 4MB).

satsie:
@GBKS done and done! ImageOptim is a cool tool :+1:

GBKS (approved):
ACK. Love it.

PR #1254, opened 2025-02-25
v1.10.5 - Remove alcohol from chapter 10

satsie:
Merge dev into master to cut the v1.10.5 release which includes an update to make chapter 10 non-alcoholic (at least in the English version :smiley: )


benalleng (approved):
LGTM! Make sure to include every commit in these merges into master! I always forget to switch from squashed and it messes me up.

PR #1256, opened 2025-03-13
English syntax improvements

Walpurga03:
This change unifies character encoding and formatting in the English localization file (en.ts):

- All apostrophes between letters have been converted to ' from ´ (e.g.,  "it´s" becomes "it's")
- String definitions have been standardized:
  * All string values are wrapped in backticks (`) by default. This eliminates the need for most, if not all of the escape sequences.

These changes improve code readability and consistency without affecting functionality.
The standardized syntax makes future edits and translations easier and more reliable.

The adjustments were made using a custom JavaScript script (convert-apostrophes.js)
specifically developed for this purpose.

EDITED on May 7 to reflect changes in the PR.


i18n/locales/en.ts
```diff
@@ -524,26 +527,23 @@ const translations = {
...
-        tip: "Can you do the math? Based on 16 possible characters per digit, what do you think the likelihood of finding a hash starting with '00' is? How about '000'?",
-        spoiler: "Try something starting with the letter 's'.",
+        power_of_random_paragraph: `Like a perfectly random scenario, probability-wise the SHA-256 hash function is completely random for each addition of new data. If you hashed the novel \'War and Peace\', then added just one extra letter to the original text, it would result in a completely different hash. Each new character changes the hash in a completely random manner. Considering that there are 16 possible characters in a hexadecimal system (0-9 and a-f), how likely is it to find a hash starting with \'0\'?`,
+        tip: `Can you do the math? Based on 16 possible characters per digit, what do you think the likelihood of finding a hash starting with \'00\' is? How about \'000\'?`,
+        spoiler: `Try something starting with the letter \'s\'.`,
```
satsie, line 532: > For values that already contain single quotes, backticks (`) are used instead
> to avoid the need for escape sequences

I think the script added an unnecessary escape sequence here, as well as in the `power_of_random_paragraph` a few lines above it.

satsie (changes requested):
Wow! This is a cool script and it picked up on a lot of changes. Thank you for including it! 

I'm sorry I didn't specify in my message on Discord what the preferred apostrophe is. I didn't realize there were so many  inconsistencies in the file. `´` should be changed to `'` (e.g., "it´s" becomes "it's" ). Is it possible to rerun the script so it does the opposite? 

I realize this may create situations where the new apostrophe should be escaped. To make it easier, what about wrapping all strings in backticks instead of a mix of backticks and single quotes? That way there is no need to escape any apostrophes. Let me know your thoughts!

i18n/locales/en.ts
Walpurga03: is that ok? This script standardises the strings in the English localisation file (en.ts) by replacing all special apostrophes (‘) with normal apostrophes (’) and enclosing all string definitions uniformly with backticks (`).

Walpurga03: I found another character because I changed something. It's a really big file. I hope I've found them all now.

benalleng:
Concept ACK, nice to see some standardization. I'd ask just that the script be moved into `lib/` or `utils/`

i18n/locales/en.ts
satsie: Thanks for the updates! I will thoroughly review the file to make sure changes didn't break anything. As you mentioned, it's quite large so it might take me a hot minute.

Walpurga03:
I have moved the script to the Utils folder. If there is anything else to do, just let me know.

i18n/locales/en.ts
```diff
@@ -1,2237 +1,1822 @@
...
-        paragraph_two:
-          'If you want to get involved with design, story, or anything else, join us in the #saving-satoshi channel in the <a href="https://discord.gg/B9WURkwF" target="_blank" rel="noreferrer">Bitcoin Design Discord</a>. Mention how you’d like to contribute and we’ll help point you in the right direction.',
+        title: `Contribute`,
+        paragraph_one: `All suggestions are welcome. We are also always looking for help reviewing and QA-ing code changes. For code improvements, you can directly open an issue or submit a pull request on <a href="https://github.com/saving-satoshi/saving-satoshi" target="_blank" rel="noreferrer">GitHub</a>.`,
+        paragraph_two: `If you want to get involved with design, story, or anything else, join us in the #saving-satoshi channel in the <a href="https://discord.gg/B9WURkwF" target="_blank" rel="noreferrer">Bitcoin Design Discord</a>. Mention how you'd like to contribute and we'll help point you in the right direction.`,
       },
```
satsie, line 66: Note to self, this is where I left off on review.

i18n/locales/en.ts
```diff
@@ -1,2237 +1,1822 @@
...
-        spoiler:
-          'Nothing to spoil! Just go ahead and read through this lesson. There will be more opportunities to show off your  mining knowledge in future lessons!',
+        mining_heading: `Mining`,
+        mining_paragraph: `Mining is where these concepts all come together. Bitcoin mining is the process by which new bitcoins are created and transactions are added to the blockchain. Miners compete to solve complex mathematical puzzles by finding a unique, valid 'nonce' in each block of transactions. This proof-of-work process demands significant computational power and energy, making it a secure and decentralized way to validate transactions. Successful miners are rewarded with newly minted bitcoins and transaction fees, and they play a critical role in maintaining the integrity of the bitcoin network.`,
+        difficulty_heading: `Difficulty`,
+        difficulty_paragraph: `The mining difficulty is what allows the block to remain at an average of 10 minutes between each new block.`,
+        spoiler: `Nothing to spoil! Just go ahead and read through this lesson. There will be more opportunities to show off your  mining knowledge in future lessons!`,
       },
```
satsie, line 440: I have reviewed the first 2 chapters, 440/2700 lines => about 15% of the diff

satsie:
I just updated the PR description to reflect some changes that have been made since its initial opening.

satsie (approved):

PR #1262, opened 2025-04-16
Deprecate crypto.Hash function in favor of createHash

### LINKED ISSUE #1261: Chapter 6 - NodeJS crypto.Hash constructor is deprecated
I was doing some story related stuff in Chapter 6 and noticed a new deprecation error for the NodeJS crypto.Hash constructor:

Image

Looks like it was indeed deprecated in NodeJS v23.11: https://nodejs.org/api/deprecations.html#DEP0179

This is the challenge I was playing: https://savingsatoshi.com/en/chapters/chapter-6/put-it-together-2-normal

And I got this error by copy/pasting the solution from the help page. 

Need to do more investigation. The solution is to likely fix the answer code, and it might make sense to upgrade NodeJS on our repl containers. Not sure what version it's running right now.

benalleng:
This is a long needed change as `createHash` has been around for a while and `Hash` is ultimately an old method. This PR does not do anything to check node versions only to replace the `Hash` function with `createHash`.

> Need to do more investigation. The solution is to likely fix the answer code, and it might make sense to upgrade NodeJS on our repl containers. Not sure what version it's running right now.

image

We must uses latest or something, I can't remember

Closes #1261 

[Preview](https://saving-satoshi-git-fork-benalleng-deprecat-6017c5-savingsatoshi.vercel.app/en/chapters/chapter-6/in-out-5?dev=true)

### DIFF
--- a/content/lessons/chapter-6/hard/put-it-together-3.tsx
+++ b/content/lessons/chapter-6/hard/put-it-together-3.tsx
@@ -158,7 +158,7 @@ console.log('KILL')
       name: 'put-it-together-3-hard',
       args: ['args'],
     },
-    defaultCode: `const {Hash} = require('crypto');
+    defaultCode: `const {createHash} = require('crypto');
 
 ${prevData.data.slice(0, -2)}
   digest(input_index) {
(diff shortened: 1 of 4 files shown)


satsie:
Thanks @benalleng! This looks great. I tested the put-it-together-1 exercise, but my account is having issues accessing the put-it-together-3-hard exercise. The change to that challenge is identical though so I think it's fine that I am unable to test it. I'm going to merge this PR.

satsie (approved):

Issue #1271, opened 2025-06-05
Add Korean Language


louis-noncelab:
We are the [Bitcoin Social Layer](http://bsl.pub), a volunteer Bitcoin community based in Korea.
At the request of many Koreans, we created a Korean version of Saving Satoshi.
The initial translation draft is already complete, and we will be committing it soon.

### Scope

- Create new file `i18n/locales/ko.ts`
- Update `i18n/locales/index.ts` ,` i18n/config.ts`

### Completion
- Submit a pull request after completion.

PR #1276, opened 2025-06-09
Add Korean(ko) language #1271

louis-noncelab:
Add a Korean translation files. issue #1271

- Create file i18n/locales/ko.ts
- Update i18n/locales/index.ts and i18n/config.ts

Translated to fit the local Korean context. All UI elements work properly.

One thing that's a bit disappointing is that the top right menu is horizontal.
I hope that appropriate action will be taken.

image


benalleng:
고마워! I think with some pretty simple `word-break: keep-all;` we can prevent the text wrapping in Korean, Japanese, etc. here where individual characters are recognized as different words while preserving it in languages where these translations are actually 2 words.

benalleng:
I added a PR to add this css class for this Topbar, any other places that this might exist we should be able to add `break-keep` to also solve this problem in those places too.

satsie:
Thanks Ben! Just merged PR #1277. @louis-noncelab was that Topbar the only place that you were seeing the text displayed vertically?

satsie (approved):
Thank you for doing this! I rebased your changes with dev and opened a new PR, #1278. Was able to confirm that the vertical text issue was fixed as part of #1277 (thanks Ben!)

PR #1277, opened 2025-06-09
Add word break keep-all to topbar to prevent korean text from wrapping

benalleng:
By using wrod-break: keep-all we can prevent character based languages like korean, japanese, and chinese from wrapping while keeping the wrap behavior on languages where the text is actually 2 words.

Should unblock #1276

### DIFF
--- a/components/Topbar.tsx
+++ b/components/Topbar.tsx
@@ -95,7 +95,7 @@ export default function Topbar() {
                     ? routes.chaptersUrl
                     : routes.aboutUrl
                 }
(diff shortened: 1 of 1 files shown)


benalleng:
I guess the major question is whether we should just use `text-wrap: nowrap;` to prevent wrapping everywhere instead?

satsie:
I think `break-keep` sounds like the right option. The [docs](https://tailwindcss.com/docs/word-break) specifically say works for Chinese, Japanese, and Korean, and behaves like `break-normal` for all other languages.

Thanks for fixing this!

satsie (approved):

PR #1288, opened 2025-08-24
Bug on progress unlocking later lessons

### LINKED ISSUE #1272: Bug on Progress
**Describe the bug**

**To Reproduce**

1. Go to 'the first lesson on each challenge'
2. go back 'then go back to the previous lessong'
3. Then we have the other challenges unlocked

**Expected behavior**
Progress shouldn't be unlocked for other lessons so quickly without having to do any challenge

**Screenshots and/or logs**
https://drive.google.com/file/d/1IUkBJPcGTtXjoftqEhcW-lxXzdyHtINq/view?usp=sharing

0tuedon:
This PR closes #1272

This PR adds an order for all lesson, we used to have this before the state update [here](https://github.com/0tuedon/saving-satoshi/blob/d6ae73a01535edd414fb0e208e1bca7753004f2c/lib/progress.ts).

In the new file `lib/progress.ts` I added a static data for all lesson in saving satoshi, while this might be too manual, it does the job for now.

I also added a new state called presentPage, the reason for this is currentLesson just gives the last uncompleted lesson, and with that we don't know where the user is at that point in time, but with present page we get to know what exact page the user is.

Since this is a really delicate  PR, I would need deep testing to see if this actually breaks something else


satsie (commented):
For the most part I think I understand what you're doing here, but I think I'm still missing some context.

Can you point me to the code that is executed when I am on this page, https://savingsatoshi.com/en/chapters/chapter-1/intro-2 and I click the "Continue" button? That will help me better understand the need for the new `presentPage` state.

benalleng:
I ran through the challenges once, I did get one hiccup in chapter 3 where I had passed through the chapter but on the chapter selection screen chapter 4 was not yet unlocked, after redoing some challenges it fixed itself.

As well I now get this error in chapter 4 Public key 3, its possible this method is deprecated on the new node verison?
```
Result
/usr/app/index.js:11
const encodedPrivateKey = convertedToBigInt(privateKey)
^
ReferenceError: convertedToBigInt is not defined
at privateKeyToPublicKey (/usr/app/index.js:11:27)
```

benalleng:
Nvm, I forgot we used pseudo-code for the resource answers, that is just fake code.

0tuedon:
Hey guys is this good to go? @benalleng, did you have the issue again?

satsie:
I did some basic QA and can confirm that the bug indeed appears to be fixed. Thank you Ben for looking at the behavior more thoroughly.

Regarding the code...
> (NB: currentLesson is just the last completed lesson, I found out) 

This doesn't sound good. There is nothing about the variable names `currentLesson` and `presentLesson` that indicate they could ever have different values. It sounds to me like there could be an underlying bug around how the `currentLesson` variable is set, or, at best, it is a poorly named variable. I expect `currentLesson` to hold the same value as `presentLesson`.

It's also possible `currentLesson` is being made to do too much, hence the need for the new `presentLesson` variable.

Either way, I understand any hesitance to dig deeper and rip existing code apart, especially for fear of unknown side effects.

If you find yourself back in this code again, feel free to explore it deeper and propose a refactor. Something seems off about this, but it probably isn't worth the time unless it comes up again.

Thanks for coming up with a patch that works and doesn't disrupt much of the existing code. I will approve and merge it now!

satsie (approved):

PR #1291, opened 2025-09-05
Update types/node version to 20.0.0

benalleng:
This should placate the vercel builds to not get upset about node 18 being deprecated.

This is in conjunction with updating the vercel node build version

As a note I don't think this is totally required to work but I wanted to get this inline with the overall node version

## Pull Request checklist

- Clean commit history


benalleng:
I transitioned this to 20.0.0 as bumping to 22.0.0 caused a bunch of build errors, 20.0.0 is still in LTS so It shouldn't be a problem for a while. but there are some things to address once 20.0.0 is dropped

0tuedon:
Don't we need to changed the .nvmrc file too?

0tuedon (approved):
Node v20 works, 22 upwards doesn't, I need to get it to work before the next cut off

Issue #1283, opened 2025-07-30
Chapter 4, public-key-3: Use accordion for some instructions


nandini584:
The issue deals with instructional clarity. Reducing the extra information apart from the question itself to an accordion so that users can visit those as per their requirements.

Advantages:
1. Enhances the text-heavy content.
2. Improves user focus on what the coding problem is.
3. Avoid getting them overwhelmed by the problem.

Here's the [Figma File](https://www.figma.com/design/qHNmvIn1zyUS63RIyyqwiQ/Saving-Satoshi--Copy-?node-id=1535-19614&t=eZDd6lfT15zB06Ft-1), checkout 
Page Chapter 4 Specs 
Frame Handover

<img width="661" height="764" alt="Image" src="https://github.com/user-attachments/assets/738cc5b3-9934-4131-b920-a3179d0b8c34" />
<img width="657" height="441" alt="Image" src="https://github.com/user-attachments/assets/1b81c2f4-8cb8-4780-8857-0f2630f2609f" />

satsie:
Thanks for filing this issue @nandini584! This is for https://savingsatoshi.com/en/chapters/chapter-4/public-key-3, right? These wireframes you provided encompass more than just the accordion change so it's a little confusing for anyone that hasn't been on the weekly calls. Each ticket should have enough context and clarity such that any contributor can pick it up. 

Regarding the wireframes, one example of an unrelated change is the instructions no longer contain an explanation for the generator point. I know we talked about instruction clarity as an area for improvement but I think we'd be doing the player a disservice to not mention some basics of how elliptic curve mathematical operations work.

For this ticket I ask that the only things that get implemented are:

- [ ] Move the type-conversion hints to an accordion
- [ ] Create a new accordion for examples. Here is an example to use:

```
private key (hex): 3af12b445641e7c1c49322df968e67ff24a952d669df3b618cabbb41a600a670
private key (integer): 26660253977894295461109986997779781205809396931917713711877026925737131681392
public key (hex): (ecbd3207bb765efbd815f8ede94af1d4f22b026c1272106497bc96260de77ba0, feb8973b7f46d79863a1e5c92d0ff650f2c16ff5662df4462ebef3fe5dbfd8f3)
public key (integer): (107080111656673042052127869876037415003757155714704583955391137403493119654816, 115213607167807907920389156886193951148164516778797468829883371017257917339891)
```

satsie:
[...]

PR #1293, opened 2025-09-08
Accordion component

0tuedon:
This PR creates the #1283  Accordion component for use, from the discussion only the type-conversions and examples on Chapter-4 public-key-3  was created with the accordion component.

[preview](https://saving-satoshi-q5e5wgsi2-savingsatoshi.vercel.app/en/chapters/chapter-4/public-key-3?dev=true)

EDIT: updated preview link


i18n/locales/en.ts
```diff
@@ -631,6 +631,7 @@ const translations = {
       nav_title: `Calculate the public key`,
       paragraph_one: `To derive a public key from a private key, we perform an elliptic curve operation repeatedly with the generator point. The generator point is a specific point on the curve. Its value is part of the secp256k1 standard and it's always the same:`,
       paragraph_two: `Mathematical operations on an elliptic curve are similar to addition. Therefore, repetition of those operations is similar to multiplication. We use the * symbol to describe the algorithm, where 'k' is the private key and 'P is the corresponding public key: `,
+      paragraph_three: `Some type-conversion hints`,
```
satsie, line 634:
```suggestion
      paragraph_three: `Type-conversion hints`,
```

After seeing it live I think we can shorten the heading of this accordion

content/lessons/chapter-4/public-key-3.tsx
```diff
@@ -155,24 +156,32 @@ def privatekey_to_publickey(private_key):
         <Text className="mt-4 font-nunito text-xl text-white">
           {t(`chapter_four.public_key_three.${language}.paragraph_three`)}
         </Text>
-        <Text className="font-nunito text-xl italic text-white">
-          {t('chapter_four.public_key_three.paragraph_four')}
-        </Text>
-        <CodeExample
-          className="mt-4"
-          code={`# Python
+        <Accordion
```
satsie, line 159: Can we add an itty bitty amount of space just above this accordion?

satsie (changes requested):
I am fine with merging this and doing a follow up PR for the accordion with the example. I just have 2 small changes to request for this.

content/lessons/chapter-4/public-key-3.tsx
satsie: I think this should fix it

```suggestion
        <Accordion className="mt-4"
```

satsie (approved):

content/lessons/chapter-4/public-key-3.tsx
```diff
@@ -157,7 +157,7 @@ def privatekey_to_publickey(private_key):
           {t(`chapter_four.public_key_three.${language}.paragraph_three`)}
         </Text>
 
-        <Accordion
+        <Accordion className="mt-4"
```
benalleng, line 160: Pretty sure you'll need to add a corresponding param to accept this as `Accordian` is a handrolled component not any default element.

benalleng (changes requested):

content/lessons/chapter-4/public-key-3.tsx
satsie: Ooof. Sorry. That was me. I thought I could sneak a quick change in before the merge, hah. I'll look for examples in the code that do what you described. Thanks for diagnosing this so quickly!

0tuedon: On it 👀

PR #1298, opened 2025-10-07
Menu Redesign for Lessons

### LINKED ISSUE #1285: Remove numerical navigation
Problem:

The number navigation can be difficult to use and we can only navigate through the current chapter using it while if we want to refer back to things from previous chapter, its required to go back to the home page, search for the chapter amd then for the subchapter which is a frustrating process to be not able to see the list in one go.

Here's the [Figma File](https://www.figma.com/design/qHNmvIn1zyUS63RIyyqwiQ/Saving-Satoshi--Copy-?node-id=11248-2683&t=1MDgGpjbDvgEFVxS-1), checkout Page "Chapter 4 Specs"

Frame Handover
Remove Numbered Navigation

Image

### LINKED ISSUE #1287: Enhanced Menu bar for improved navigation
Whenever user wants to navigate through the current chapter or previous chapters, they can click on the navigation menu icon and a side bar open up with the current page items, now they can either navigate in the current chapter or close it to select some other chapter or challenge for navigation. Icons of "add" and "subtract" can be found in bitcoin icons and the "menu" icon can also be found in the link given below.

Here's the [Figma File](https://www.figma.com/design/qHNmvIn1zyUS63RIyyqwiQ/Saving-Satoshi--Copy-?node-id=11248-2683&t=1MDgGpjbDvgEFVxS-1), checkout Page "Chapter 4 Specs"

Frame Handover
"Additional Navigation menu"

The assets can be found on [bitcoin icons](https://bitcoinicons.com/)

Image
Image
Image

0tuedon:
This PR closes #1287 and closes #1285 ,  This PR creates a new style of menu and shown i issue #1287, 

- Create Navigation Drawer with Chapters and Lessons (Done)
- Handle Hard mode and Easy Mode on Chapter 6 (Pending)

### DIFF
--- a/components/Navbar/NavbarDesktop/ChapterAccordion.tsx
+++ b/components/Navbar/NavbarDesktop/ChapterAccordion.tsx
@@ -160,8 +160,6 @@ const ChapterAccordion = ({
       {isOpen && (
         <div className="flex flex-col">
           <div>
-            {hasDifficulty && <DifficultySelection chapterId={id} />}
-
             <Title className="">
               <span className="text-white/50">1. </span> {t('navbar.intro')}
             </Title>
--- a/components/Navbar/NavbarDesktop/ChapterLesson.tsx
+++ b/components/Navbar/NavbarDesktop/ChapterLesson.tsx
@@ -6,6 +6,7 @@ import React from 'react'
 import Icon from 'shared/Icon'
 import {
   currentChapterAtom,
+  getLessonKey,
   isLessonCompletedUsingLessonName,
 } from 'state/progressState'
 import { CourseProgress } from 'types'
@@ -27,6 +28,7 @@ const ChapterLesson = ({
   const { isDevelopment } = useEnvironment()
   const currentChapter = useAtomValue(currentChapterAtom)
   const { lessonId: lessonName } = usePathData()
+  const lessonKey = getLessonKey(`chapter-${chapterId}`, id)
   const slug = `chapter-${chapterId}`
   const link = `/${lang}/chapters/${slug}/${id}${
     isDevelopment ? '?dev=true' : ''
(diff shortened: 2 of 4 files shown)


satsie:
Here's the first issue I spotted. If you go backwards to a lesson you have already done, or a screen you have already unlocked, the menu behaves like you have never progressed past the current screen. 

For example, in this video I am on the first screen of the game. Then I progress to the second screen. Everything looks good. However when I go back to the first screen and open up the menu, it incorrectly shows the second screen as locked. However the chapters page reflects the correct progress.

https://github.com/user-attachments/assets/3e5bbece-2e76-4caa-807d-5f4c799d09f3


The second issue is some weirdness with Chapter 6's easy/hard mode. I see in this PR description that you have the functionality marked as pending but I'm not sure if that information is outdated. I'm also not sure if this is existing behavior but I will describe it anyway. As far as I can tell this PR does not touch any of the lesson progress logic so what I'm seeing could very well be unrelated to this PR. Can you confirm that this PR does not change any lesson progress logic?

I have completed chapter 6 in normal mode. I decided to toggle on hard mode and go to the first screen in chapter 6. From there I press continue, but I get a 404 because I have already played chapter 6 in normal mode.

https://github.com/user-attachments/assets/7428f988-3c00-4108-8c69-a82f348894d5

Another thing I saw is I can revisit chapter 6 in normal mode then toggle hard mode on using this new menu. Once I do that the game gets very confused and I end up going down the same code path that triggers the 404 I just mentioned.

Maybe we should remove the normal/hard mode toggle from the menu. The way the game is right now, this is only meant to be set once, when the user first begins the chapter. I think having it in the menu it makes it too easy for players to turn it on and off.

Side note - at some point we should revisit the logic for this since I can see a scenario where a user selects hard mode, then they find it is too difficult, and they want to go to normal mode.

https://github.com/user-attachments/assets/adc8dd86-7546-4d1c-9d12-77542ba05f0a

0tuedon:
Hmm, I think I will make the release with #1301  that bug fix before this

PR #1312, opened 2025-11-26
BOSS Challenge 2026 - Call BOSS bot to generate next clue

satsie:
Call the BOSS bot to generate a new clue for the user at the end of the game.

I will need to set the following environment variables/secrets before this goes live. Does this happen in Vercel? Right now the preview link doesn't work because these env vars are not set. I just requested access to the Saving Satoshi Vercel.
- `NEXT_PUBLIC_BOSSBOT_API_ENDPOINT`
- `NEXT_PUBLIC_BOSSBOT_AUTH`

We should merge this by the end of the month so that we can cut a new release and have this go live on December 1. BOSS Challenge announcement is Dec 2 in Mauritius, which is sometime around 2am EST on Dec 2.

Test instructions:
- Go to `/en/chapters/chapter-10/outro-6` [Preview link](https://saving-satoshi-git-fork-satsie-bc-ending-bossbot-savingsatoshi.vercel.app/en/chapters/chapter-10/outro-6?dev=true)
- Verify that you get the clue. XOR the clue with your private key and check that it decrypts.

Screenshot from 2025-11-25 23-38-32

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,8 @@
 NEXT_PUBLIC_API_ENDPOINT=<url-of-your-locally-running-backend|https://api.savingsatoshi.com>
 NEXT_PUBLIC_WS_ENDPOINT=<url-of-your-locally-running-backend|wss://api.savingsatoshi.com>
+NEXT_PUBLIC_BOSSBOT_API_ENDPOINT=something
+NEXT_PUBLIC_BOSSBOT_AUTH=something
+
+# If running the backend locally, the env vars will look something like this:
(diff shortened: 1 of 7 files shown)


0tuedon (approved):
LGTM! is it possibe to make sure they don't get it when dev true is on?

satsie:
> LGTM! is it possibe to make sure they don't get it when dev true is on?

Answering this question for posterity :smile: I know I am going to come back to this PR a year from now and have the same question. The answer is dev mode is not activated in production, and even if it was, you would need a secret key to use it.

Issue #1316, opened 2025-12-02
Bug Report: Lottie SSR Crash + Manifest Syntax Error


aniket866:
---

## Describe the bug

The application fails to render due to a server-side crash caused by `lottie-web` being executed during SSR. The server logs show:

ReferenceError: document is not defined  
node_modules/lottie-web/build/player/lottie.js (30:0) @ createTag

This occurs because `lottie-web` expects browser-only globals (`document`, `window`) which are not available in the Node.js server environment used by Next.js during SSR.

Additionally, the browser console reports:

Manifest: Line: 1, column: 1, Syntax error.

This indicates a malformed or invalid `site.webmanifest` file.

---

## To Reproduce

1. Clone this repository.  
2. Run the project using `npm install` followed by `npm run dev`.  
3. Open `http://localhost:3000` in a browser.  
4. Open DevTools → Console.  
5. Navigate to any page that imports from:  
   `content/lessons/chapter-4/public-key-2.tsx`  
6. Observe the SSR crash and console errors.

A screenshot was included showing the main page loaded with console errors visible on the right side.

---

## Expected behavior

- The server should render pages without crashing.  
- Lottie animations should only run in the browser (client-side).  
- The manifest file should load without syntax errors.  

---

## Screenshots and/or logs

- Server error: `ReferenceError: document is not defined`  
- Browser console shows:  
  - Manifest syntax error  
  - React experimental warning (non-critical)

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/a6a22119-b529-48a3-b0fa-5cda25bc5328" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/4826b5c2-73bb-4fae-8a26-264969c32914" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/2a214304-c1eb-49b5-b1ac-6c07c3a6db3d" />

---

## Desktop (please complete the following information)

- OS: Windows 10  
- Browser: Chrome  
- Version: Latest (as shown in screenshot)  

---

## Smartphone (please complete the following information)

Not tested on mobile.

---

## Additional context

The root cause is that `lottie-web` is imported at the top level of a file that Next.js attempts to server-render. Since `lottie-web` touches `document` immediately upon import, SSR fails.

The correct fix is to:

- Move all Lottie logic into a **client-only component**  
- Dynamically import `lottie-web` inside a `useEffect()`  
[...]

PR #1318, opened 2025-12-03
fix: avoid document usage during SSR in public-key-2 lesson

### LINKED ISSUE #1317: 🚨 Bug Report — `yarn build` fails with `ReferenceError: document is not defined` during prerender
### Describe the bug

While running `yarn build`, the static export build fails during the **prerendering** phase with the following error:
This stops the entire build process and prevents successful deployment to production.

---

### Root Cause (Detailed Reasoning)

**Next.js** attempts to prerender pages on the **server**, where browser globals like `document`, `window`, and `navigator` do not exist.

Some component imported into `/` or `/_not-found` is calling **browser-only APIs** (e.g., `document.createElement`) during render or module import, which forces this code to execute on the server, causing a crash.

The stack trace:

ReferenceError: document is not defined at createTag (....next\server\chunks\757.js:1:193878) ...

suggests usage of DOM manipulation inside a utility or animation-related component included on the homepage.

This is commonly caused by:
* Importing browser-only code at the top of a **React Server Component**
* DOM access outside `useEffect`

---

### To Reproduce
 **Run production build**
    ```
    yarn build
    ```
    Build crashes while generating static pages

### Expected behavior

* `yarn build` should complete successfully without server execution errors
* Home page (`/`) and `_not-found` should be statically prerendered
* Browser-only code should run strictly on **client side only**

---

### Actual behavior

The `yarn build` command executes the Next.js build process. However, during the **prerendering** phase for static export, the following occurs:

1.  The build process encounters code that attempts to access the browser-only global object, `document`, while executing on the **Node.js server environment**.
2.  This attempt fails, causing a `ReferenceError: document is not defined`.
3.  The prerendering of the pages `/` (home page) and `/_not-found` fails.
4.  The entire build process is halted and terminates with a non-zero exit code (`error Command failed with exit code 1.`).
5.  No static output files are successfully generated, preventing deployment.

**Log excerpt confirming this behavior:**

Image
Image

#### Why this happens ⬇

When building for production, Next.js executes all **server-side code** without DOM. If any imported component or utility interacts with:

* `document`
* `window`
* `navigator`
* `localStorage`
* CSS Injection that depends on DOM

…in a **Server Component** or at the module top-level, prerendering fails.

---

aniket866:
### 📝 Summary

This PR resolves a build failure (`ReferenceError: document is not defined`) that occurred during the static prerendering phase of `yarn build`. This error was caused by the `react-lottie` library attempting to access browser-specific APIs (like `document`) when executed on the Next.js server.

The fix ensures that the `Lottie` component, which depends on browser APIs, is only imported and rendered on the client side.

### 🔍 Root Cause

The `Lottie` component was imported statically, forcing it to be executed during the server-side rendering (SSR) or prerendering phase, where the DOM object (`document`) does not exist.

### 🛠️ Changes Implemented

The `react-lottie` library is now imported dynamically using `next/dynamic` with Server-Side Rendering explicitly disabled (`ssr: false`).

The minimal patch changes are:

1.  **Added** `import dynamic from 'next/dynamic'`.
2.  **Replaced** the static import `import Lottie from 'react-lottie'` with:
    ```javascript
    const Lottie = dynamic(() => import('react-lottie'), { ssr: false })
    ```

This ensures the animation library is deferred until the browser environment is available.

### ✅ Testing Steps and Verification

1.  Clone the repository and switch to this branch.
2.  Install dependencies: `yarn install`
3.  Run the production build command: `yarn build`
4.  **Expected result:** The build should complete successfully without encountering the `ReferenceError: document is not defined` error.

#### Successful Build Log (After Fix):

image

@satsie , @GBKS  review this PR

Closes #1317 and  #1316

### DIFF
--- a/content/lessons/chapter-4/public-key-2.tsx
+++ b/content/lessons/chapter-4/public-key-2.tsx
@@ -1,8 +1,8 @@
 'use client'
-
+import dynamic from 'next/dynamic'
 import { useMediaQuery, useTranslations } from 'hooks'
 import { HolocatQuestion, TextImage } from 'ui'
-import Lottie from 'react-lottie'
+const Lottie = dynamic<any>(() => import('react-lottie'), { ssr: false })
 import Animation1 from 'public/assets/animations/Animation1.json'
 import focusellipse from 'public/assets/images/focus-ellipse.png'
 import Circle from 'shared/icons/Circle'


ekzyis:
Fwiw, can confirm this fixes #1317

0tuedon (approved):

PR #1330, opened 2025-12-08
Disable the book a call screen

satsie:
Now that this is part of the BOSS Challenge, we can remove this screen.

@0tuedon are there any issues with the way I have commented out the screen? If someone's most recent progress state is the book-a-call-1 screen, will their account be in a buggy state?


satsie:
> hmm, for this if the last progress is then yes, but we can just forward the book-a-call screen to the next lesson

Does that require new code? What would that look like?

0tuedon (approved):
LGTM! I don't think it will break, althought the users progress will have this but it won't break because we don't rely on the users progress for the next lesson.

PR #1331, opened 2025-12-09
Fix: Clear stale error state in Chapter 9 Proposal-3 execution

### LINKED ISSUE #1178: ch 9, proposal-3: Error does not clear
This issue was reported to us by a user in discord: https://discord.com/channels/903125802726596648/1089879043626778655/1306594337894826077

> I entered the wrong script once, got the error and then corrected the script and submitted the solution but I kept getting the same error. After i refreshed the page and pasted the solution again, it worked.

It sounds like once you get an error, it does not clear and you keep seeing it for subsequent tries. 

Note that this has not been reproduced yet, so I am tagging it as a "good first issue" for any new contributors that want to dip their toes in some QA! It would be very helpful to get confirmation that this is indeed a bug, and what the steps are to reproduce it.

rajanarahul93:
## Summary

This pull request fixes a long-standing issue in **Chapter 9 → Proposal-3**, where an error encountered during script execution would continue appearing even after the user corrected their input. The executor state was not fully reset between runs, causing stale stack data, conditional state, and error flags to persist.

The result:  
Users who entered an incorrect script once would continue seeing the same error until a full page refresh.

This PR ensures the executor is reconstructed cleanly for every run, preventing stale state from carrying over.

---

## What was happening

- `LanguageExecutor` reused internal state (`stack`, `conditionalState`, `negate`, `currentIndex`) across runs.
- `OpRunner` recreated the executor, but React reused the same instance reference, so the internal state was not reset.
- The error produced on the first execution persisted even after the script was fixed.

This exactly matches the behavior described in Issue #1178.

---

### 1. Properly reset `LanguageExecutor` internal state
- Store a clean `stackInitial`
- Restore stack, conditional flags, and pointer on each `execute()` call
- Remove unintended `.execute()` call in the error handler

### 2. Force executor re-instantiation in the UI
- Clear executor with `setExecutor(null)`  
- Re-create it on the next frame using `requestAnimationFrame`
- Reset the UI state (`stateHistory`, `startedTyping`) before execution

This guarantees React does *not* reuse the previous executor instance.

---

## Result

- Errors no longer persist between runs
- Executor resets correctly every time the user clicks “Run the Script”
- Behavior now matches user expectations and matches the pattern of earlier chapters

---

## Linked Issue

Fixes #1178

---

## Checklist

- [x] Verified build locally  
- [x] Tested execution flow for success, failure, and edge cases  
- [x] Clean commit history  
- [x] No changes outside the intended Chapter 9 components

### DIFF
--- a/ui/lesson/OpCodeChallenge/LanguageExecutor.ts
+++ b/ui/lesson/OpCodeChallenge/LanguageExecutor.ts
@@ -11,6 +11,7 @@ import {
 class LanguageExecutor {
   tokens: T
   stack: StackType
+  stackInitial: string[]
   height?: number | null
   nSequenceTime?: number | null
   conditionalState: Array<boolean>
@@ -63,7 +64,7 @@ class LanguageExecutor {
     initialStack?: string[],
     height?: number,
     nSequenceTime?: number
-  ): LanguageExecutor | null {
+  ): LanguageExecutor {
     const parsableInput = this.rawInputToParsableInput(script)
     const filteredStack = initialStack?.filter((arg) => {
       arg.trim()
@@ -82,15 +83,12 @@ class LanguageExecutor {
 
       return langExecutor
     } catch (err) {
-      const langExecutor = new LanguageExecutor(
+      return new LanguageExecutor(
         [],
         filteredStack ?? [],
         height,
         nSequenceTime
       )
-      langExecutor.execute()
-
-      return langExecutor
     }
   }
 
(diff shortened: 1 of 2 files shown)


satsie:
@rajanarahul93 can you describe the steps to reproduce the error? If you could include a video that would be very helpful. Issue #1178 first calls for some QA to verify that we know how to trigger the error state. That needs to be done before submitting a PR. Feel free to attach any relevant investigations directly to that issue!

PR #1334, opened 2025-12-11
Fix: accumulate chunked outputs before display

### LINKED ISSUE #1295: Timeouts in Chapter 8 RPC commands
In the [first challenge in chapter 8](https://savingsatoshi.com/en/chapters/chapter-8/building-blocks-3), the user must print the network difficulty by using the "getinfo" RPC command.

```
console.log(Bitcoin.rpc('getinfo'))
```

However, when I execute this command, the response I get back is in varying states of completeness:

Image

Image

Image

Image

Image

I suspect this may have something to do with the timeouts we implemented for the Docker containers on the backend. More investigation is needed

jrakibi:
Closes #1295

Currently, to display the code runner result, Docker streams the output in multiple chunks. The problem with the current implementation is that we close the socket after we receive the first chunk (which may or may not contain the full output, as you can see from the images in #1295)

Instead of displaying just the first chunk and closing the socket, we now accumulate the chunks in `outputBuffer` and only display the result after we have collected all chunks ( after the `'end'` message is received), then we close the socket.

### DIFF
--- a/ui/lesson/ScriptingChallenge/Runner/index.tsx
+++ b/ui/lesson/ScriptingChallenge/Runner/index.tsx
@@ -81,6 +81,7 @@ export default function Runner({
   const [loading, setLoading] = useState<boolean>(false)
   const [isRunning, setIsRunning] = useState<boolean>(false)
   const hasResult = useRef(false)
+  const outputBuffer = useRef<string[]>([])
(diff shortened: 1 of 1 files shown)


0tuedon (approved):
LGTM!

PR #1344, opened 2025-12-17
ch.9, proposal-3: you can spend before the time, VP can spend after

pinheadmz:
Sorry 😬


satsie (approved):
Thanks for the fix! This will go out very soon in the release I'm about to cut

satsie:
Omg I merged after reviewing the code but before I saw that second comment :see_no_evil: Will look into it now.

PR #1345, opened 2025-12-17
add XOR clue so people dont give up at the last step

adamjonas:


satsie (approved):
This is a good addition. Merging!

PR #1346, opened 2025-12-18
ch.9, proposal-3: fix validation

satsie:
This is a follow up PR to https://github.com/saving-satoshi/saving-satoshi/pull/1344

### DIFF
--- a/content/lessons/chapter-9/proposal-3.tsx
+++ b/content/lessons/chapter-9/proposal-3.tsx
@@ -40,7 +40,7 @@ export default function Proposal3({ lang }) {
         success={success}
         setSuccess={setSuccess}
         initialHeight={6930299}
-        initialStackScript={['SIG(ME)', '0']}
+        initialStackScript={['SIG(ME)', '1']}
         nextStepMessage={t('chapter_nine.proposal_three.next_step_message')}
       >
         <LessonInfo>
--- a/content/lessons/chapter-9/proposal-3.tsx
+++ b/content/lessons/chapter-9/proposal-3.tsx
@@ -40,7 +40,7 @@ export default function Proposal3({ lang }) {
         success={success}
         setSuccess={setSuccess}
         initialHeight={6930299}
-        initialStackScript={['SIG(ME)', '1']}
+        initialStackScript={['SIG(VANDERPOOLE)', '0']}
         nextStepMessage={t('chapter_nine.proposal_three.next_step_message')}
       >
         <LessonInfo>


satsie:
There is something buggy about this code that is causing this tweak to the exercise to fail. This conditional should be evaluating to true but it is not: https://github.com/saving-satoshi/saving-satoshi/blob/6dbb762fc4a11c488e82c8f6f71cab728dc69d50/ui/lesson/OpCodeChallenge/OpRunner.tsx#L323C5-L333C15

image

pinheadmz (approved):
ACK nice catch, seems to work correctly

image

weikengchen:
:) I was stuck here and will wait for this PR to be merged before continuing.

Also the hints given weren't right

0tuedon (approved):
tACK

satsie:
Wait was anyone able to test the second stage of this exercise? i tried the patch that you committed @0tuedon and it broke on the second phase of this exercise. Will test now

0tuedon:
> Wait was anyone able to test the second stage of this exercise? i tried the patch that you committed @0tuedon and it broke on the second phase of this exercise. Will test now

Oh I did test it, and @pinheadmz did also, maybe we missed something?

Testing again

satsie:
Ok I just tried it and it works! I am so confused because I tried what your commit does last night and it wasn't working for me locally. Either way, happy to have this fixed! Thank you! 
[Screencast from 2025-12-18 16-25-22.webm](https://github.com/user-attachments/assets/ab1e5c0d-0d70-4b68-870a-219e7432ea9a)

weikengchen:
for the record, this has not been updated to https://savingsatoshi.com/ yet, right?

weikengchen:
Oh I can see that one has to do "SIG(VANDERPOOLE) 0" instead of "SIG(VANDERPOOLE) 1" for the check to pass.

satsie:
> for the record, this has not been updated to https://savingsatoshi.com/ yet, right?

It is! It went out with the release earlier today: https://github.com/saving-satoshi/saving-satoshi/releases/tag/v1.10.14

otaliptus:
After finishing the first part I kept clicking RESET and spent some time trying to find what I'm doing wrong. Now watching the video I realized what to do (also my solution was a bit different but that's ok)

Maybe the exact play button could be shown better :)

satsie:
> After finishing the first part I kept clicking RESET and spent some time trying to find what I'm doing wrong. Now watching the video I realized what to do (also my solution was a bit different but that's ok)
> 
> 
> 
> Maybe the exact play button could be shown better :)

Thanks for sharing your experience! This is good feedback. I agree that at minimum we can swap the play and reset buttons, but maybe there's something more we can do with the design.

@aassoiants @nandini584 can you add this to the feedback you are collecting?

pro258b:
> for the record, this has not been updated to https://savingsatoshi.com/ yet, right?

don't think so

guyHouri:
It is not live unfortunately. It works only if your sig is on the if statement and vanderpoole's is on the else statement.

pinheadmz:
@pro258b @guyHouri it is live. You can spend before the timelock, VP can spend after the timelock -- are you finding that is not the case?

PR #1350, opened 2025-12-25
Fix CH9 proposal_three hint (spender before/after 6930300)

beihaili:
## Summary
Fixes a reversed hint in Chapter 9 (resources → proposal_three) about who can spend before/after block height 6930300.

## Problem
Several locales incorrectly stated the spender roles around height 6930300 (reversed compared to the lesson logic / en copy), which can mislead learners in the timelock + conditional challenge.

## Fix
Align the “before/after 6930300” and corresponding stack-hint roles with the intended lesson behavior:
- Before 6930300: you (ME) can spend (use your signature)
- After 6930300: Vanderpoole can spend (use his signature and provide 0 after the CLTV path)

## Locales updated
- zh-cn
- vi
- pt
- ko
- ja
- fr


satsie:
Oh wow, thank you for this! The translations were missed when https://github.com/saving-satoshi/saving-satoshi/pull/1344 was merged. Appreciate you catching this and taking care of all the other locales.

satsie (approved):

Issue #1328, opened 2025-12-07
Chapter 4, Calculate public key accepts any GE


cassiot:
It's hard to know if the posted answer is really correct or not, because simply logging or returning the provided G returns a successful message.

<img width="1014" height="757" alt="Image" src="https://github.com/user-attachments/assets/5b2d987a-a800-46d2-a0db-45034a9e1c86" />

jamaljsr:
Fixed in #1354

PR #1354, opened 2025-12-25
Script validation in chapter-4 public-key-3

0tuedon:
This PR fixes issue #1328 , where a JavaScript cheat allowed users to bypass the challenge by calling console.log(G). This was possible because the challenge did not strictly validate that the correct uncompressed public key was generated.

The fix uses the @noble/secp256k1 library to generate the expected uncompressed public key and compare it with the user’s output. Since each public key coordinate is 32 bytes, we extract and format the coordinates to match the console output format, making sure key is generated correctly.

This issue wasn't present in python

[LINK TO TEST](https://saving-satoshi-egllhi4im-savingsatoshi.vercel.app/en/chapters/chapter-4/public-key-3?dev=true)

### DIFF
--- a/content/lessons/chapter-4/public-key-3.tsx
+++ b/content/lessons/chapter-4/public-key-3.tsx
@@ -9,11 +9,28 @@ import { getLanguageString } from 'lib/SavedCode'
 import { useAtom } from 'jotai'
 import { accountAtom, currentLanguageAtom } from 'state/state'
 import Accordion from 'ui/common/Accordion'
+import * as secp256k1 from '@noble/secp256k1'
 
 export const metadata = {
   title: 'chapter_four.public_key_three.title',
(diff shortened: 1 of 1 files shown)


pinheadmz:
image

I'm guessing this is because in the test link you provided there is no actual private key value? So it's probably multiplying G by 0 which would be "infinity". 

Other solutions like G*2 or just G did log actual curve points that were correctly evaluated as wrong answers

0tuedon:
> I'm guessing this is because in the test link you provided there is no actual private key value? So it's probably multiplying G by 0 which would be "infinity".

Yh this is why,  I guess  I could just return undefined, so that it passes, on production you can't get there without a private key anyways (being logged in)

pinheadmz (approved):
Fully tested by logging in and using dev mode with an actual private key

PR #1368, opened 2025-12-28
Progress issue with hard and normal mode unsynced

### LINKED ISSUE #1342: Selecting hard mode in chapter 6 leads to page not found error
**Describe the bug**
Selecting hard mode in chapter 6 leads to page not found error 

**To Reproduce**
1. Go to Chapter 6 in overview- https://savingsatoshi.com/en/chapters
2. Set slider to hard mode
3. Click "Start Chapter 6"
4. In https://savingsatoshi.com/en/chapters/chapter-6/intro-1 click "Continue"
5. The intro-2 page briefly flashes
6. Page not found error is shown: https://savingsatoshi.com/en/chaptersundefined

- OS: Ubuntu 25.10
- Browser: Brave 1.84.132

**same issue on smartphone**

- Device: Samsung S25
- OS: Android
- Browser : Brave
- same issue

0tuedon:
This PR should close #1342 . The issue can be reproduced by completing Normal mode in Chapter 6 and then attempting Hard mode in the same chapter. The root cause is that lessons without both Normal and Hard modes are not being properly synced.

This PR ensures those lessons are synced correctly, and it prevents the 404 error from appearing on the page.

### DIFF
--- a/hooks/useProceed.ts
+++ b/hooks/useProceed.ts
@@ -34,16 +34,19 @@ export default function useProceed() {
 
   const Proceed = () => {
     let route
-    if (!currentLesson?.completed && nextLessonUsingCurrentRoute) {
+
+    if (nextLessonUsingCurrentRoute) {
       route =
-        routes.chaptersUrl + nextLessonUsingCurrentRoute?.path + queryParams
-      markLessonAsComplete(currentLessonId)
+        routes.chaptersUrl + nextLessonUsingCurrentRoute.path + queryParams
+
+      if (!currentLesson?.completed) {
+        markLessonAsComplete(currentLessonId)
+      }
     } else {
-      route =
-        routes.chaptersUrl + nextLessonUsingCurrentRoute?.path + queryParams
+      route = routes.chaptersUrl + nextLessonPath + queryParams
       progressToNextLesson()
     }
-    console.log(route, nextLessonUsingCurrentRoute?.path)
+
     router.push(route, { scroll: true })
   }
 
--- a/state/progressState.ts
+++ b/state/progressState.ts
@@ -508,18 +508,65 @@ const syncCourseProgressEffectAtom = atomEffect((get, set) => {
     combinedProgressAndAccountAtom
   )
   if (isAuthLoading || !isProgressLoaded) return
+
+  const syncedProgress = syncSharedLessonCompletion(courseProgress)
+
   if (account) {
-    setProgress(courseProgress)
+    setProgress(syncedProgress)
   } else {
-    setProgressLocal(courseProgress)
+    setProgressLocal(syncedProgress)
   }
 })
+
+// function to sync shared lesson completion across difficulties
+function syncSharedLessonCompletion(
+  courseProgress: CourseProgress
+): CourseProgress {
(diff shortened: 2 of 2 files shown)


pinheadmz:
Can't reproduce the issue.
My steps:
- Dev branch f4585a2f, backend master branch  5e19afa with all chapters "unlocked"
- Played chapters 1-6 locally without dev=true, left ch 6 in easy mode
- From chapter 7 intro screen, return to home
- switch chapter 6 to hard and click Start Chapter 6
- I get http://localhost:3000/en/chapters/chapter-6/intro-1
- click continue
- I get teleported to http://localhost:3000/en/chapters/chapter-7/intro-1 and brwoser console log says "lesson locked"

In other words, I am unable to play ch 6 in hard mode after competing in easy mode. I don't get a 404 and I don't get the flash of chapter 2 as described.

From this state, I switched front end repo locally to this branch 213d6808 and still can not access chapter6 hard mode.

Is there some kind of database migration necessary? I dont understand the issue very well but I dont see how a front end change would fix access to  locked lesson?

0tuedon:
> In other words, I am unable to play ch 6 in hard mode after competing in easy mode. I don't get a 404 and I don't get the flash of chapter 2 as described.

> From this state, I switched front end repo locally to this branch [213d680](https://github.com/saving-satoshi/saving-satoshi/commit/213d6808785211b5ba80a7c528b62c49ab82d8e2) and still can not access chapter6 hard mode.
> 
> Is there some kind of database migration necessary? I dont understand the issue very well but I dont see how a front end change would fix access to locked lesson?

The reason why a frontend change would affect this is because the backend is literally just a store, it stores the progress, the frontend handles the access to these  hard mode challenge logic, I have just pushed a new commit, using the `syncedProgress` instead of the `courseProgress` atom. This worked for me but would need you to try it.

jamaljsr:
I also wasn't able to reproduce the 404 error as mentioned in the issue. I did get redirected to ch 7 after completing ch 6 in normal mode then trying to start hard mode. I also saw the "lesson locked" log that @pinheadmz mentioned. I was using the dev branch to reproduce this behavior.

When I switched to this PR branch, instead of jumping to ch 7, I was sent to the 6th lesson `/chapter-6/in-out-4-hard`. Is this the intended behavior or should users start hard mode from lesson 1? To me it feels a bit strange to jump into the middle of the chapter even if I have already completed the same first 5 lessons in normal mode. I would expect to just redo the common ones all over again. If this is the desired outcome, then I can proceed with reviewing the code. I'd just like to confirm first.

Update:
I've opened a PR #1439 that uses a different approach than this. Instead of syncing the completed lessons between the normal/hard tracks, which causes the skip-ahead I mentioned above, the other PR fixes the root cause of `CH6INT1` HARD not being marked completed when you click Continue on that lesson. That's what causes the redirect to ch 7. With my fix, none of the common lessons in the HARD track will be auto-marked as completed. They'll have to go through the whole chapter again.

I'm throwing it out there as another way to solve the problem and to get your thoughts. I'm open to either solution depending on the desired UX.

0tuedon:
Closed in favour of #1439

Issue #1124, opened 2024-10-15
Lost part of my progression after loging out. Stuck in chapter 6


kaiosatiro:
I stepped on this issue and thought it was a good idea to share it with you all.

**Describe the bug**
After I logged out and logged in again using my private key, my progression after chapter 6 was lost.
I had saved all my code answers with me, so I could paste them to recover. However, when I finished chapter six again, the next one was not liberated, it just sent me to the chapter page.

So I tried in hard mode, but after submitting the code for the "Input Class" exercise, the page sent me to a 404 page.

**To Reproduce**
In easy mode:
1. I finished all the steps and could see the congrats page.
2. Backing to the chapters page, the ch 7 is still blocked.

In hard mode:
1.  When submitting the "Chapter 6, The ins and outs input class", I got a successful answer.
2. Clicking the next button sends me to a 404 page.

**Expected behavior**
Normal advancing through the chapters

**Screenshots and/or logs**
The easy mode issue: (Chp 7 is still blocked)
![image](https://github.com/user-attachments/assets/b8cc7c1b-a1f7-4732-ac7f-5b3efd673dac)

Hard mode:
![image](https://github.com/user-attachments/assets/3e4e5a2c-71ab-4bf4-a5bd-fed56cc7d805)
![image](https://github.com/user-attachments/assets/046e1fbd-e613-44a0-8b6d-f18a995c828c)

**Desktop (please complete the following information):**
I'm using Firefox and Brave in Arch Linux.

**Additional context**
This happened after I logged in again using the private key. Before I logged out, it happened a similar thing the issue #1071.

benalleng:
Do you mind sharing the account key for this?
I want to see what in the progress object has broken to cause this behaviour.
Also I will likely be able to go in and hand edit the progress object to fix it for you so you can restore your progress

benalleng:
also do you have the exact link that was the 404 page?

kaiosatiro:
Sure!

kaiosatiro:
Sure !
The page: https://savingsatoshi.com/en/chaptersnull
The key: ef588ef6e333ddf29dfcd9905edfec478da3bc0c4fa653c863b4813028d8c516

I wish I could provide more information about it.

benalleng:
Thanks! Did you fully complete any chapters beyond chapter 6? or did you finish chapter 6 in `normal` mode and then immediately try again in `hard` mode

kaiosatiro:
[...]

PR #1374, opened 2025-12-29
Loss of progress when going back to previous lesson

### LINKED ISSUE #1347: Loss of progress
I encountered a complete loss of the progress I had.

**Describe the bug**
Game resets to the initial progress state: when accessing the level where I had been previously it redirects to the first level.

I was at chapter 9-6 opcodes.

Image and now I'm back at the 1st chapter: 

Image

**To Reproduce**
I cannot state particular steps to reproduce. It seemed it was completely random. I'm not sure, but it might be when I swiped back and instead of going to the previous level it has reset.

**Expected behavior**
Progress is expected not to be lost.

- OS: Ubuntu 24.04
- Browser firefox
- Version 146.0.1

**Additional context**
Some other users have encountered the very same problem: https://discord.com/channels/903125802726596648/1449453799826784276.

0tuedon:
This has been a long-standing issue dating back to #1124, appearing in different forms. This PR should close #1347, which describes the current version of the problem.

This  bug happens when a user goes back to a completed lesson and clicked Proceed. The app updated progress when it shouldn’t, causing progress to break. This PR fixes that by only updating progress for unfinished lessons, and keeps user progress untouched.

This will need some more test to confirm

### DIFF
--- a/hooks/useProceed.ts
+++ b/hooks/useProceed.ts
@@ -3,14 +3,12 @@
 import { useRouter } from 'next/navigation'
 import { useLocalizedRoutes, usePathData } from 'hooks'
 import useEnvironment from './useEnvironment'
-import { useAtom, useAtomValue, useSetAtom } from 'jotai'
+import { useAtomValue, useSetAtom } from 'jotai'
 import {
   getLessonKey,
   getNextLessonUsingChapterIdAndLessonName,
   isLessonCompletedUsingId,
   markLessonAsCompleteAtom,
-  nextLessonPathAtom,
-  progressToNextLessonAtom,
   syncedCourseProgressAtom,
 } from 'state/progressState'
 
@@ -19,9 +17,7 @@ export default function useProceed() {
   const router = useRouter()
   const { isDevelopment } = useEnvironment()
   const routes = useLocalizedRoutes()
-  const nextLessonPath = useAtomValue(nextLessonPathAtom)
-  const [courseProgress, setCourseProgress] = useAtom(syncedCourseProgressAtom)
-  const progressToNextLesson = useSetAtom(progressToNextLessonAtom)
+  const courseProgress = useAtomValue(syncedCourseProgressAtom)
   const queryParams = isDevelopment ? '?dev=true' : ''
   const nextLessonUsingCurrentRoute = getNextLessonUsingChapterIdAndLessonName(
     chapterId,
@@ -48,7 +44,7 @@ export default function useProceed() {
         routes.chaptersUrl + nextLessonUsingCurrentRoute?.path + queryParams
       progressToNextLesson()
     }
-    console.log(route, nextLessonUsingCurrentRoute?.path)
+
     router.push(route, { scroll: true })
   }
 
--- a/hooks/useProceed.ts
+++ b/hooks/useProceed.ts
@@ -3,12 +3,14 @@
 import { useRouter } from 'next/navigation'
 import { useLocalizedRoutes, usePathData } from 'hooks'
 import useEnvironment from './useEnvironment'
-import { useAtomValue, useSetAtom } from 'jotai'
+import { useAtom, useAtomValue, useSetAtom } from 'jotai'
 import {
   getLessonKey,
   getNextLessonUsingChapterIdAndLessonName,
   isLessonCompletedUsingId,
   markLessonAsCompleteAtom,
+  nextLessonPathAtom,
+  progressToNextLessonAtom,
   syncedCourseProgressAtom,
 } from 'state/progressState'
 
@@ -17,7 +19,9 @@ export default function useProceed() {
   const router = useRouter()
   const { isDevelopment } = useEnvironment()
   const routes = useLocalizedRoutes()
-  const courseProgress = useAtomValue(syncedCourseProgressAtom)
+  const nextLessonPath = useAtomValue(nextLessonPathAtom)
+  const [courseProgress, setCourseProgress] = useAtom(syncedCourseProgressAtom)
+  const progressToNextLesson = useSetAtom(progressToNextLessonAtom)
   const queryParams = isDevelopment ? '?dev=true' : ''
   const nextLessonUsingCurrentRoute = getNextLessonUsingChapterIdAndLessonName(
     chapterId,
@@ -44,7 +48,7 @@ export default function useProceed() {
         routes.chaptersUrl + nextLessonUsingCurrentRoute?.path + queryParams
       progressToNextLesson()
     }
-
+    console.log(route, nextLessonUsingCurrentRoute?.path)
     router.push(route, { scroll: true })
   }
 
--- a/state/AuthFunctions.ts
+++ b/state/AuthFunctions.ts
@@ -26,9 +26,9 @@ export const useAuthFunctions = () => {
       const account = await getSession()
       setAccount(account)
 
-      // Load progress for the logged-in account
       const progress = await getProgress()
-      setCourseProgress(mergeProgressState(defaultProgressState, progress))
+      const finalProgress = progress || defaultProgressState
+      setCourseProgress(mergeProgressState(defaultProgressState, finalProgress))
 
       return true
     } catch (ex) {
@@ -43,7 +43,13 @@ export const useAuthFunctions = () => {
     window.location.pathname !== '/' &&
       router.push(routes.chaptersUrl, { scroll: true })
     await logout()
+
+    const progress = localStorage.getItem('SavingSatoshiProgress')
     localStorage.clear()
+    if (progress) {
+      localStorage.setItem('SavingSatoshiProgress', progress)
+    }
+
     setCourseProgress(defaultProgressState)
     setAccount(undefined)
   }
--- a/state/AuthFunctions.ts
+++ b/state/AuthFunctions.ts
@@ -27,8 +27,9 @@ export const useAuthFunctions = () => {
       setAccount(account)
 
       const progress = await getProgress()
-      const finalProgress = progress || defaultProgressState
-      setCourseProgress(mergeProgressState(defaultProgressState, finalProgress))
+      if (progress) {
+        setCourseProgress(mergeProgressState(defaultProgressState, progress))
+      }
 
       return true
     } catch (ex) {
--- a/api/local/getProgress.ts
+++ b/api/local/getProgress.ts
@@ -1,18 +1,16 @@
-import { defaultProgressState } from 'state/progressState'
 import { CourseProgress } from 'types'
 
-export default async function getProgressLocal(): Promise<CourseProgress> {
+export default async function getProgressLocal(): Promise<CourseProgress | null> {
   try {
     const localStorageProgress = localStorage.getItem('SavingSatoshiProgress')
 
     if (!localStorageProgress) {
-      throw new Error('Could not read progress from LocalStorage')
-    } else {
-      const res = JSON.parse(localStorageProgress)
-      return res
+      return null
     }
+
+    return JSON.parse(localStorageProgress)
   } catch (errors) {
     console.error(errors)
-    return defaultProgressState
+    return null
   }
 }
--- a/api/local/setProgress.ts
+++ b/api/local/setProgress.ts
@@ -2,9 +2,10 @@ import { CourseProgress } from 'types'
 
 export default async function setProgress(progress: CourseProgress) {
   try {
-    localStorage.setItem('SavingSatoshiProgress', JSON.stringify(progress))
+    const serialized = JSON.stringify(progress)
+    localStorage.setItem('SavingSatoshiProgress', serialized)
   } catch (errors) {
-    console.error(errors)
+    console.error('Failed to save progress to localStorage:', errors)
     return undefined
   }
 }
--- a/api/progress/getProgress.ts
+++ b/api/progress/getProgress.ts
@@ -1,8 +1,7 @@
-import { defaultProgressState } from 'state/progressState'
 import { CourseProgress } from 'types'
 import { get } from 'utils'
 
-export default async function getProgress(): Promise<CourseProgress> {
+export default async function getProgress(): Promise<CourseProgress | null> {
   try {
     const res = await get({
       url: '/v1/progress',
@@ -12,6 +11,6 @@ export default async function getProgress(): Promise<CourseProgress> {
     return res.progress_state
   } catch (errors) {
     console.error(errors)
-    return defaultProgressState
+    return null
   }
 }
--- a/api/progress/setProgress.ts
+++ b/api/progress/setProgress.ts
@@ -15,7 +15,7 @@ export default async function setProgress(
 
     return res
   } catch (errors) {
-    console.error(errors)
+    console.error('Failed to save progress to backend:', errors)
     return undefined
   }
 }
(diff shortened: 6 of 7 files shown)


hooks/useProceed.ts
```diff
@@ -19,9 +17,7 @@ export default function useProceed() {
   const router = useRouter()
   const { isDevelopment } = useEnvironment()
   const routes = useLocalizedRoutes()
-  const nextLessonPath = useAtomValue(nextLessonPathAtom)
-  const [courseProgress, setCourseProgress] = useAtom(syncedCourseProgressAtom)
-  const progressToNextLesson = useSetAtom(progressToNextLessonAtom)
```
jrakibi, line 24: Just an early review, I haven’t dug deep into the code yet (I’ll do that tomo), but if we remove this (`progressToNextLesson`), does it mean we don’t need that whole block anymore?
https://github.com/saving-satoshi/saving-satoshi/blob/a2aa53d0c69949dc8663b62bc6a83115aaa43e43/state/progressState.ts#L730-L807

If so, let’s remove it.

pinheadmz:
> This PR fixes that by only updating progress for unfinished lessons, and keeps user progress untouched.

You mean only updating progress for *finished* lessons?

0tuedon:
> > This PR fixes that by only updating progress for unfinished lessons, and keeps user progress untouched.
> 
> You mean only updating progress for _finished_ lessons?

No, uncompleted lessons, take for example a user is in CH10INT1 (Chapter 10 intro 1) and goes back to CH9INT1 to check somethings, the progress will not update when the user goes back to chapter 9 because the user has completed that lesson

pinheadmz:
Not sure how to test this. I tried reproducing by completing chapters 1 & 2, then going back to chapter 1 and completing a challenge, but I didnt lose my progress (chapter 3 was still unlocked)

0tuedon:
Hi @pinheadmz and @jrakibi ,  I noticed another issue with progress handling.
When the backend request fails, this line falls back to the default progress:
https://github.com/saving-satoshi/saving-satoshi/blob/a2aa53d0c69949dc8663b62bc6a83115aaa43e43/api/progress/getProgress.ts#L15

In the new commit, instead of using the default progress (which will reset users progress if there are errors), we now read the last saved progress from localStorage, and also we save the progress to the backend and local

This is implemented in commit:
https://github.com/saving-satoshi/saving-satoshi/pull/1374/commits/ba96f4a32b20f7d54d8a22cdde3722a43176a9bd

We’ll need another round of testing, specifically to simulate backend failures and confirm this behavior.

0tuedon:
So I also do have a video of testing what was currently on dev and in this PR, by mocking a 500 error on the backend for every request and that resets the progress,

On Dev ATP
https://github.com/user-attachments/assets/af30fae1-8667-4314-8316-516b36cfd110

In this PR 
https://github.com/user-attachments/assets/a0504dd3-6357-4533-8f82-5ac13823d271


This does make it work and we don't see a reset.

pinheadmz:
Couldnt reproduce this locally either. Completed chapters 1-6, went back to chapter 1 and completed it again, then was sent to chapter 7 as expected

0tuedon:
> Couldnt reproduce this locally either. Completed chapters 1-6, went back to chapter 1 and completed it again, then was sent to chapter 7 as expected

Hmm, to reproduced, checkout to dev branch, login and then throw a 500 error in your local backend and keep pressing continue, that should reset the progress to chapter 1 @pinheadmz

pinheadmz:
Okay that worked, but switching to this branch didnt fix that. On both dev and PR branches, progress was lost when backend was unreachable, and was not restored until it became reachable again.

Do I need to be clearing browser cache or swithcing to fresh accounts for every trial?

0tuedon:
> Okay that worked, but switching to this branch didnt fix that. On both dev and PR branches, progress was lost when backend was unreachable, and was not restored until it became reachable again.
> 
> Do I need to be clearing browser cache or swithcing to fresh accounts for every trial?

Okay, so here's how to differentiate  it, on dev if i redirects you to chapter-1 and you click continue and then bring the backend back to normal and click continue it should return youe progress back to chapter one 🥲, when you bring the backend back don't refresh

state/AuthFunctions.ts
```diff
@@ -26,9 +26,9 @@ export const useAuthFunctions = () => {
       const account = await getSession()
       setAccount(account)
 
-      // Load progress for the logged-in account
       const progress = await getProgress()
-      setCourseProgress(mergeProgressState(defaultProgressState, progress))
+      const finalProgress = progress || defaultProgressState
```
jamaljsr: When I start off unauthenticated, then complete all the lessons in chapter 1, then sign up, all my chapter 1 progress is lost. I am redirected to `/chapter-1/intro-1` instead of `/chapter-2/intro-1`. I believe this line may be the culprit but I'm not 100% sure.

jamaljsr (changes requested):
I have been reviewing these changes. It appears to me that there are multiple different but related issues that are occurring:

### 1. Progress is lost when a user goes back to previous lesson and clicks Proceed

I was not able to reproduce this locally using the `dev` branch. I authenticated and completed CH3 then went back to CH1 and clicked Continue on a few lessons. I still had my expected CH3 progress intact even after reloading the page. Can you provide more detailed steps to reproduce this, if you are able to?

### 2. Progress is lost when there is a backend outage 
> (ex: `GET /v1/progress` returns HTTP 500 error)

I was able to reproduce this in the `dev` branch by following these steps:
- completed all lessons in chapters 1-3 (`"currentLesson":"CH4INT1"`)
- updated the `GET /v1/progress` API endpoint to return a HTTP 500 error 
- reload the page
- remove the HTTP 500 error from `GET /v1/progress` 
- reload the page again
- my progress on the server is reset back to chapter 1 (the default state) with no way to recover

After switching to this PR branch, I repeated the same steps and did not lose my progress. So your changes do resolve this specific situation.

### 3. Progress is lost when opening multiple tabs

I saw in #1124 that some users mentioned facing issues when opening new tabs to view previous lessons. I was testing this case myself and was able to reproduce the issue in the `dev` branch by following these steps:
- tab-1: completed all lessons in chapters 1-3 (`"currentLesson":"CH4INT1"`)
- tab-2: go to chapter 2, lesson 1
- tab-2: click Continue a few times
- tab-1: reload the page
- my progress was reset back to where I left off in chapter 2 in tab-2. I lost the chapter 3 lessons that were completed

Progress is being reset back to which ever tab clicked Proceed last.

After switching to this PR branch, I repeated the same steps and did not lose my progress. 

---

There is one functional regression I noticed that should be addressed. I commented inline where I think the problem is.

jamaljsr:
From reading the referenced issues, it still feels like there are cases where user's progress is lost, but it's not clear how to reproduce the exact scenario that they faced. These could be due to backend outages or network issues on the user's side. I do believe that the changes in this PR are an improvement and should be merged. I also think we can do a bit more.

Currently, the backend `PUT /v1/progress` API endpoint will overwrite the user's progress with whatever data is received. So if somehow the progress in the browser resets back to a previously completed lesson (CH1), the backend will happily overwrite the data in the DB, even if the DB data has more progress completed. 

I would suggest updating the [/v1/progress route](https://github.com/saving-satoshi/saving-satoshi-backend/blob/master/src/routes/v1/progress.ts#L74) to compare the incoming data with what's stored in the DB and only update the DB if progress is moving forward, not backwards. This would make it impossible for users to lose any completed lessons, no matter what wackiness is happening on the client or how many tabs they have open. 

The only way the user would lose progress would be if the backend API was offline for an extended period of time and the user completed lessons that are not being saved to the DB. If they also lose their local progress somehow, before the backend comes back online, then they would get reset to the latest completed lesson from the backend when it's back online. I'm not sure if there's anything the can/should be done to address this edge case but just wanted to acknowledge it.

In summary, I believe that with this PR and my suggested backend changes, the lost progress issues should be pretty much resolved going forward. I'd like to hear your thoughts on this. I would be happy to open a PR on the backend repo if you agree.

0tuedon:
> ### 1. Progress is lost when a user goes back to previous lesson and clicks Proceed
> I was not able to reproduce this locally using the `dev` branch. I authenticated and completed CH3 then went back to CH1 and clicked Continue on a few lessons. I still had my expected CH3 progress intact even after reloading the page. Can you provide more detailed steps to reproduce this, if you are able to?

This happens, when a user goes back and when there is an error it resets when next the user clicks continue

0tuedon:
>2. Progress is lost when there is a backend outage
> ex: GET /v1/progress returns HTTP 500 error)

> I was able to reproduce this in the dev branch by following these steps:

> completed all lessons in chapters 1-3 ("currentLesson":"CH4INT1")
updated the GET /v1/progress API endpoint to return a HTTP 500 error
reload the page
remove the HTTP 500 error from GET /v1/progress
reload the page again
my progress on the server is reset back to chapter 1 (the default state) with no way to recover

Good, this happens because of how it is structure with JSON and when the FE sends a request with the new state it resets

0tuedon:
>3. Progress is lost when opening multiple tabs
>I saw in https://github.com/saving-satoshi/saving-satoshi/issues/1124 that some users mentioned facing issues when >opening new tabs to view previous lessons. I was testing this case myself and was able to reproduce the issue in the dev >branch by following these steps:

>tab-1: completed all lessons in chapters 1-3 ("currentLesson":"CH4INT1")
>tab-2: go to chapter 2, lesson 1
>tab-2: click Continue a few times
>tab-1: reload the page
>my progress was reset back to where I left off in chapter 2 in tab-2. I lost the chapter 3 lessons that were completed
>Progress is being reset back to which ever tab clicked Proceed last.

> After switching to this PR branch, I repeated the same steps and did not lose my progress.

I believe the best solution is always relying on the backend, but hopefully on the call later we are able to discuss this, I do think the way the backend stores progress is one of the issues, did this happen on this PR? or on dev?

state/AuthFunctions.ts
0tuedon: Okay, I will check this, it could be this

0tuedon:
> Currently, the backend `PUT /v1/progress` API endpoint will overwrite the user's progress with whatever data is received. So if somehow the progress in the browser resets back to a previously completed lesson (CH1), the backend will happily overwrite the data in the DB, even if the DB data has more progress completed.
> 
> I would suggest updating the [/v1/progress route](https://github.com/saving-satoshi/saving-satoshi-backend/blob/master/src/routes/v1/progress.ts#L74) to compare the incoming data with what's stored in the DB and only update the DB if progress is moving forward, not backwards. This would make it impossible for users to lose any completed lessons, no matter what wackiness is happening on the client or how many tabs they have open.


Yh I do, the backend structure basically does a garbage in and out, so it does trust the frontend in the progress which leads to this issue, what happening here is to try as much as possible to not let that happen of the Frontend, but the lasting solution is to create a structure where the backend checks for this.

jamaljsr:
> > After switching to this PR branch, I repeated the same steps and did not lose my progress.
> 
> I believe the best solution is always relying on the backend, but hopefully on the call later we are able to discuss this, I do think the way the backend stores progress is one of the issues, did this happen on this PR? or on dev?

I was able to repro the issue in the `dev` branch. The problem is resolved when switching to this PR branch.

satsie:
Sorry for being a little late to the conversation. I'm not sure what the current state of discussion is, but FWIW I also agree with the following idea that @jamaljsr proposed:

> I would suggest updating the [/v1/progress route](https://github.com/saving-satoshi/saving-satoshi-backend/blob/master/src/routes/v1/progress.ts#L74) to compare the incoming data with what's stored in the DB and only update the DB if progress is moving forward, not backwards. This would make it impossible for users to lose any completed lessons, no matter what wackiness is happening on the client or how many tabs they have open.

I hope you don't mind Jamal but I made a ticket in the backend and assigned it to you: https://github.com/saving-satoshi/saving-satoshi-backend/issues/95 I'll let you decide how it should be prioritized in relation with the other tickets.

The loss of progress due to a backend outage (500) reminds me of [this bug fix](https://github.com/saving-satoshi/saving-satoshi/pull/1445) you guys did recently. Sensing a bit of a theme :laughing: Great investigating. Hopefully there aren't too many other places that need to be updated to handle errors on the backend.

I am going to review the code in this PR now!

state/AuthFunctions.ts
satsie: I just tested the latest preview link for commit 26c18437bce18c9193936544639419b59574c4e9 and this issue is still present :(

satsie:
Results of my QA on this branch:

### :heavy_check_mark:  1. Progress is lost when a user goes back to previous lesson and clicks Proceed
Based on [Tuedon's comment](https://github.com/saving-satoshi/saving-satoshi/pull/1374#issuecomment-3764158739), I played all of chapters 1 and 2. Once I had chapter 3 unlocked I went back to chapter 2 and changed my answer for the "Build a hash function" (chapter-2/scripting-2) exercise to something that would not pass. I also went back to earlier lessons and clicked continue. No loss of progress was observed. 

### :heavy_check_mark: 3. Progress is lost when opening multiple tabs

I followed [Jamal's steps from this comment](https://github.com/saving-satoshi/saving-satoshi/pull/1374#pullrequestreview-3674012633) and saw no loss of progress.

If I have time I will try to circle back and test the scenario where the backend returns a 500

state/AuthFunctions.ts
```diff
@@ -43,7 +43,13 @@ export const useAuthFunctions = () => {
     window.location.pathname !== '/' &&
       router.push(routes.chaptersUrl, { scroll: true })
     await logout()
+
+    const progress = localStorage.getItem('SavingSatoshiProgress')
```
satsie, line 48: Might want to make `SavingSatoshiProgress` a const so it becomes harder to accidentally change.

satsie:
The code seems fine to me :+1:  The only outstanding issue is [the regression that Jamal found](https://github.com/saving-satoshi/saving-satoshi/pull/1374#discussion_r2701078107).

0tuedon:
Updated this @jamaljsr and @satsie , Please review when you have got the time.

state/AuthFunctions.ts
satsie: I just tested this and as of commit 7099dee0eb17b6fa7b6e6edb250d385cd3407d69 it is now fixed!

state/AuthFunctions.ts
satsie: I just realize this comment is a little vague. I was thinking something like `const PROGRESS_KEY = 'SavingSatoshiProgress'` however I don't know what the code conventions are for this repo off the top of my head. No need to waste time by updating this unless you happen to come back to this file.

satsie:
[TACK](https://github.com/saving-satoshi/saving-satoshi/pull/1374#discussion_r2870449569) but I will let jamaljsr give the final approval :)

satsie:
Also, I need to say how beautiful it is that this PR #1374 resolves issue #1347

jamaljsr (approved):
I retested all 3 scenarios I [previously](https://github.com/saving-satoshi/saving-satoshi/pull/1374#pullrequestreview-3674012633) posted and confirmed no progress is lost. I also confirmed the chapter 1 regressions is resolved. Thanks for finishing this PR up. tACK LGTM 👌

state/AuthFunctions.ts
jamaljsr: Confirmed this is fixed as well.

PR #1384, opened 2025-12-30
Fix language switching being disabled incorrectly in Chapter 6

### LINKED ISSUE #1380: Code editor unexpectedly switches from Python to Javascript in Chapter 6
Description:

When I am on Chapter 6 (started with Python) https://savingsatoshi.com/en/chapters/chapter-6/put-it-together-2-normal
I can't switch language, there is a prompt:  "Language disabled since you've started this chapter in Javascript"

**To Reproduce**

1. Go to my latest progress
2.  switch language from Js to Python

**Expected behavior**

I always used Python to complete the previous steps, until at this task randomly clicked "Javascript" then "Python" trying to restore it to my original setting.
not working.
Even if I used a new browser with my private key, still the same prompt.

**Screenshots and/or logs**

Image

- OS: [PC]
- Browser [ chrome]
- Version [143.0.7499.169 ]

**Additional context**

aniket866:
## Pull Request checklist

- [x] Clean commit history

# Fix Language Switching Being Disabled Incorrectly in Chapter 6

## Summary
This PR fixes an issue where language switching in Chapter 6 appeared enabled in the UI but did not actually update the editor state. The root cause was incorrect conditional logic and block structure inside the language selection handler, which prevented the switching logic from executing as intended.

---

### Observed Behavior
- Language tabs were clickable and visually updated
- Console logs indicated language selection events
- Editor language and code content did not change

### Expected Behavior
- Users should be able to switch languages freely until the challenge is completed
- Once the challenge is successfully completed, language switching should be disabled

### Incorrect Conditional Guard
Language switching logic was gated incorrectly:
- Switching was allowed only when `challengeSuccess === true`
- This inverted the intended behavior

### Invalid Block Structure
An extra standalone `{}` block was placed after a `return` statement, resulting in misleading and unreachable logic.
---

## Fix Implemented

- Corrected the conditional guard to block switching **only after** challenge completion
- Removed the invalid block structure
- Ensured `onSelectLanguage` is optional and does not gate core switching logic
- Restored consistent behavior with other working chapters

---

### Before (Broken)
```
const handleSetLanguage = (value) => {
  if (challengeSuccess) {
    setLanguage(value)
    onSelectLanguage?.(value)
    setCurrentLanguage(getLanguageFromString(value))
    setCode(config.languages[value].defaultCode?.toString())
    setHiddenRange(config.languages[value].hiddenRange)
    setConstraints(config.languages[value].constraints)
  }
}

```

```
const handleSetLanguage = (value) => {
  if (challengeSuccess) return

  onSelectLanguage?.(value)
  setLanguage(value)
  setCurrentLanguage(getLanguageFromString(value))
  setCode(config.languages[value].defaultCode?.toString())
  setHiddenRange(config.languages[value].hiddenRange)
  setConstraints(config.languages[value].constraints)
}
```

[screen-capture (3).webm](https://github.com/user-attachments/assets/502a425c-5b26-4c2d-94f4-3087ba91beaa)

@0tuedon @benalleng  Please review this PR and can be merged ,this Fixes a crucial bug  and Closes #1380

### DIFF
--- a/ui/lesson/ScriptingChallenge/index.tsx
+++ b/ui/lesson/ScriptingChallenge/index.tsx
@@ -224,7 +224,7 @@ export default function ScriptingChallenge({
           )}
         >
           <LanguageTabs
-            languageLocked={!onSelectLanguage}
+            languageLocked={false}
             languages={config.languages}
             value={language}
             onChange={handleSetLanguage}
--- a/ui/lesson/ScriptingChallenge/index.tsx
+++ b/ui/lesson/ScriptingChallenge/index.tsx
@@ -102,14 +102,14 @@ export default function ScriptingChallenge({
   const isSmallScreen = useMediaQuery({ width: 767 })
 
   const handleSetLanguage = (value) => {
-    if (!challengeSuccess && onSelectLanguage) {
-      setLanguage(value)
-      onSelectLanguage(value)
-      setCurrentLanguage(getLanguageFromString(value))
-      setCode(config.languages[value].defaultCode?.toString())
-      setHiddenRange(config.languages[value].hiddenRange)
-      setConstraints(config.languages[value].constraints)
-    }
+    if (challengeSuccess) return
+
+    onSelectLanguage?.(value)
+    setLanguage(value)
+    setCurrentLanguage(getLanguageFromString(value))
+    setCode(config.languages[value].defaultCode?.toString())
+    setHiddenRange(config.languages[value].hiddenRange)
+    setConstraints(config.languages[value].constraints)
   }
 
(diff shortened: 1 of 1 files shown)


0tuedon (changes requested):
Hi @aniket866, thanks a lot for your PR.

This logic is used in chapters that rely on a user’s previous lesson code, so they can reuse what they already wrote to solve the final lesson challenge. Removing it would likely introduce more bugs, especially for Python users.

I think a better approach is to investigate why detectLanguage isn’t working on page refresh. I’m currently looking into this, but feel free to update your PR with a fix in that direction.

PR #1397, opened 2025-12-31
Fix typo in French localization for output message

BTCMcBoatface:
Should say 'output' in sentence "outut #1 value (40 BTC or 4,000,000,000 satoshis)"


0tuedon (approved):
ACK! Doesn't affect Behaviour

PR #1398, opened 2025-12-31
Fix typo in comment for output #1 value

BTCMcBoatface:
Changed outut to output in "outut #1 value (40 BTC or 4,000,000,000 satoshis)"


0tuedon (approved):
ACK!, doesn't affect behaviour

satsie:
Thank you for fixing this!

PR #1406, opened 2026-01-01
fix: window resize causes page refresh and state loss

### LINKED ISSUE #1338: Page automatically refreshes when browser tab size changes
**Describe the bug**
When solving the *Saving Satoshi Mysteries*, resizing the browser tab (changing the window width/height) causes the page to automatically refresh. This interrupts progress and resets the current state.

**To Reproduce**

1. Open any *Saving Satoshi Mysteries* challenge
2. Start solving the problem
3. Resize the browser window or change the tab size
4. Observe that the page refreshes automatically

**Expected behavior**
Resizing the browser window should not trigger a full page refresh. The current state and progress should remain intact.

**Actual behavior**
The page refreshes automatically whenever the browser tab size changes.

**Screenshots and/or logs**

https://github.com/user-attachments/assets/b1180238-3805-4e50-b8a6-24fc941d673e

* OS: Windows 11
* Browser: Chrome
* Version: Latest

**Smartphone (please complete the following information):**
Not tested.

**Additional context**
I wanted to confirm whether this behavior is **intentional** (e.g., tied to a resize/layout listener) or an unintended side effect.
If it’s not expected behavior, I’d be happy to help investigate or fix it.

jrakibi:
When resizing the browser window, users would lose progress and the current state would reset, as described in https://github.com/saving-satoshi/saving-satoshi/issues/1338#issue-3733566358 

The issue was that the `Lesson` component rendered two separate `<div>` elements based on screen size. When the window was resized, React unmounted one `<div>` and mounted the other, which reset all child component state.

This was fixed by using a single `<div>` with a conditional className

Closes https://github.com/saving-satoshi/saving-satoshi/issues/1338

### DIFF
--- a/ui/lesson/Lesson.tsx
+++ b/ui/lesson/Lesson.tsx
@@ -37,12 +37,15 @@ export default function Lesson({
 
   return (
     <LessonContext.Provider value={context}>
-      {direction === LessonDirection.Horizontal && (
-        <div className="flex grow flex-col md:flex-row">{props.children}</div>
-      )}
-      {direction === LessonDirection.Vertical && (
-        <div className="flex w-full grow flex-col">{props.children}</div>
-      )}
+      <div
+        className={
+          direction === LessonDirection.Horizontal
+            ? 'flex grow flex-col md:flex-row'
+            : 'flex w-full grow flex-col'
+        }
+      >
+        {props.children}
+      </div>
     </LessonContext.Provider>
   )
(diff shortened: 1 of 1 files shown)


satsie (approved):
Nice fix Jamal!! I tested the scenario in the bug report and can confirm this fixes it. It makes the code cleaner too which is an extra win :tada:

PR #1409, opened 2026-01-02
Refactor/migrate editor state to deterministic hydration model

aniket866:
# PR: Refactor Chapter 6 Editor State to Deterministic Hydration Model

## Summary
This Pull Request fundamentally re-engineers the initialization and code generation logic for Chapter 6 (`PutItTogether2` - Normal & Hard). The previous implementation suffered from "Language Contamination," where Python code would leak into the JavaScript editor (and vice-versa) due to shared state variables.

This refactor introduces a **Deterministic Hydration Model** that respects the user's global language preference, prioritizes saved drafts correctly, and uses AST-lite parsing for surgical code injection.

# Witness Logic Refactoring

### 1. JavaScript Witness Logic (`jsWitness`)
* **Static Definition:** I replaced the dynamic code generation with a strictly defined `const jsWitness`. It contains the full `class Witness` implementation using standard Node.js `Buffer` methods for serialization.
* **Type Safety:** This constant is strictly typed and only injected when `targetLang === Language.JavaScript`. This guarantees that even if the user's previous data was Python, the editor forces a clean reset to this valid JavaScript class structure.

### 2. Python Witness Logic (`pyWitness`)
* **Static Definition:** Defined a matching `const pyWitness` that implements the `class Witness` using Python's `struct.pack` library, mirroring the functionality of the JS version but with correct Python indentation and syntax.
* **Context Isolation:** The injection logic (`getLanguageCode`) explicitly selects this constant only when the Python tab is active. It acts as a firewall, preventing any JavaScript syntax (like curly braces or `this.items`) from polluting the Python environment.

### 1. Architectural Overhaul: State Hydration & Atom Integration
I have moved away from hardcoded defaults to a reactive state model that integrates with the application's global store.

* **Global State Awareness:**
    * Integrated `useAtom(currentLanguageAtom)` to read the user's preferred language from the navbar immediately upon component mount.
    * **Before:** Component blindly defaulted to `useState('python')`, forcing a layout shift or incorrect syntax highlighting.
    * **After:** `startLanguage` initializes with `getLanguageString(currentLanguage)`, ensuring the editor opens in the context the user expects.

* **Language-Based Hydration Logic:**
    * Implemented a strict "Hierarchy of Truth" for determining the active language:
        1.  **Highest Priority:** Existing draft in `localStorage` for *this specific lesson* (e.g., `CH6PUT2-javascript`).
        2.  **Medium Priority:** Completed code from the *previous lesson* (`CH6PUT1`), but only if no current draft exists.
[...]

PR #1416, opened 2026-01-05
Language disabled fix for python users on Refresh

### LINKED ISSUE #1380: Code editor unexpectedly switches from Python to Javascript in Chapter 6

0tuedon:
This PR fixes #1380 

It introduces a custom hook called `usePrevLessonLanguage`. The hook takes a `lessonId`, fetches the previous lesson’s code from the backend, and then detects the correct language from that code.

The issue with the old approach was that when the page first loads, the previous lesson data may not be available yet. As a result, `detectLanguage` returns `unknown` and the app falls back to JavaScript by default. 

To test and compare, check out from dev and start chapter 6 in-out-5 using Python, then complete the challenge. After that, refresh the next lesson page (putting-it-together-normal-1). Before this fix, the lesson defaults to JavaScript. With this fix, it correctly defaults to Python.


content/lessons/chapter-6/hard/put-it-together-1.tsx
```diff
@@ -15,22 +13,7 @@ export const metadata = {
...
-        lesson_id: 'CH6INO5',
-        data: data?.code?.getDecoded(),
-      })
-    }
-  }
-
-  useEffect(() => {
-    getPrevLessonData().finally(() => setIsLoading(false))
-  }, [])
+  const { isLoading, detectedLanguage } = usePrevLessonLanguage('CH6INO5')
```
satsie, line 16: Do we no longer need the previous lesson data? I guess we had been using it to detect the language?

content/lessons/chapter-6/hard/put-it-together-2.tsx
```diff
@@ -15,22 +13,7 @@ export const metadata = {
...
-        lesson_id: 'CH6INO4',
-        data: data?.code?.getDecoded(),
-      })
-    }
-  }
-
-  useEffect(() => {
-    getPrevLessonData().finally(() => setIsLoading(false))
-  }, [])
+  const { isLoading, detectedLanguage } = usePrevLessonLanguage('CH6INO4')
```
satsie, line 16: Should this be `CH6PUT1_HARD` (not `CH6INO4`)?

content/lessons/chapter-6/hard/put-it-together-2.tsx
```diff
@@ -301,9 +284,7 @@ class Transaction:
 
   const config: EditorConfig = {
     defaultLanguage:
-      detectLanguage(prevData.data) === Language.JavaScript
-        ? 'javascript'
-        : 'python',
+      detectedLanguage === Language.JavaScript ? 'javascript' : 'python',
```
satsie: Would it make sense to store 'javascript' and 'python' in constants?

content/lessons/chapter-6/normal/put-it-together-1.tsx
```diff
@@ -367,10 +354,7 @@ class Transaction:
   }
 
   const config: EditorConfig = {
-    defaultLanguage:
-      detectLanguage(prevData.data) === Language.JavaScript
-        ? 'javascript'
-        : 'python',
+    defaultLanguage: language,
```
satsie: The language detection is done differently in all of the hard files (`content/lessons/chapter-6/hard/*`). Those files use:

```
      detectedLanguage === Language.JavaScript ? 'javascript' : 'python',
```

Is there a reason why they can't all use the same approach? I like that this way (`useAtom(currentLanguageAtom()` with `getLanguageString(currentLanguage)`) doesn't use strings, but I don't know if the jotai atom uses additional computation power or if it complicates anything. This is totally up to you, but I would prefer if we could be consistent about this across files.

content/lessons/chapter-6/normal/put-it-together-2.tsx
```diff
@@ -22,24 +21,9 @@ export const metadata = {
...
-        lesson_id: 'CH6PUT1_NORMAL',
-        data: data?.code?.getDecoded(),
-      })
-    }
-  }
-
-  useEffect(() => {
-    getPrevLessonData().finally(() => setIsLoading(false))
-  }, [])
-
```
satsie, line 42: I think you forgot to update the `defaultLanguage` in the `EditorConfig` down after line 400

hooks/usePrevLessonLanguage.ts
```diff
@@ -0,0 +1,41 @@
...
+        const decodedData = data.code.getDecoded()
+        setPrevData({
+          lesson_id: prevLessonId,
+          data: decodedData,
+        })
+
+        const detected = detectLanguage(decodedData)
+        setDetectedLang(detected)
+
+        if (detected !== Language.Unknown) {
```
satsie, line 30: What is the difference between detected language and current language? If detected language is unknown, current language defaults to Javascript? 

I'm not sure I understand the need for two abstractions, as we could default to Javascript anyway when detected language is unknown, even without a second abstraction.

I understand if this is baked too deeply into the code for us to change it, but it's something to consider. If it doesn't need to be there, removing it could simplify the code, making future work easier.

state/DataFunctions.ts
```diff
@@ -1,12 +1,7 @@
 import { useSetAtom, useAtom } from 'jotai'
 import { isLoadingDataAtom, dataAtom, currentLanguageAtom } from './state'
 import { getData, setData } from 'api/data'
-import {
-  Language,
-  getLanguageFromString,
-  getLanguageString,
-} from 'lib/SavedCode'
-import { useEffect } from 'react'
+import { Language } from 'lib/SavedCode'
```
satsie, line 4: Is this import used at all?

state/DataFunctions.ts
```diff
@@ -47,17 +42,8 @@ export const useDataFunctions = () => {
 
   const setCurrentLanguage = (language: Language) => {
     setCurrentLanguageState(language)
-    localStorage.setItem('language', getLanguageString(language))
   }
 
-  // Initialize language from localStorage
```
satsie, line 53: Does this impact other chapters? Were any other chapters relying on the language to be set from localStorage here?

satsie (changes requested):
I have some questions about the code before I sit down to do a thorough QA on this. I also need to look at https://github.com/saving-satoshi/saving-satoshi/pull/1409. Not sure what the relationship is between this PR and that one. Do they both fix the same problem but in different ways?

content/lessons/chapter-6/hard/put-it-together-1.tsx
0tuedon: for this lesson, no it's been used to detect the language. and what happened before was it gives unknown because the data at that time hasn't loaded.

content/lessons/chapter-6/hard/put-it-together-2.tsx
0tuedon: No, the code used with the default is CH6IN04, which isn't related to hard or normal challenges, the previous value was CH6IN04

state/DataFunctions.ts
0tuedon: no, we use the same key for our jotai atom, so it should be able to get it regardless.

state/DataFunctions.ts
0tuedon: Yh

hooks/usePrevLessonLanguage.ts
0tuedon: So if there isn't a data?.code this defaults to the last clicked language, from "language" in localstorage

content/lessons/chapter-6/hard/put-it-together-1.tsx
satsie: Is it a timing thing? The `prevData` hasn't loaded yet, but it will eventually? Sorry to just notice this but wouldn't that be the most straightforward fix? Wait for the `prevData` to be loaded? Not sure if that is even possible.

content/lessons/chapter-6/hard/put-it-together-2.tsx
satsie: How come in the normal version of this file, `content/lessons/chapter-6/normal/put-it-together-2.tsx` it uses "CH6PUT1_NORMAL"? https://github.com/saving-satoshi/saving-satoshi/pull/1416/changes#diff-0a160ab220b9a1fc037f6b72b78897ebea4a98eb488c3b4ce9cc69ef02a302aaR26

Is there a mismatch between the normal and hard challenges for this chapter? Do they not go in the same order?

satsie: I tested this PR for the normal flow (f6157853fe4346b2db7445ceb2d66867367c72ff) and it looks great. However, I am unable to test it for the hard mode because my account gets a 404 everytime I try to switch to hard mode, even with dev=true and even though I completed half of chapter 6 in hard mode (I also finished it in normal mode). I can start a new account from scratch to test hard mode if needed but wanted to ask first in case you have tested hard mode.

hooks/usePrevLessonLanguage.ts
```diff
@@ -0,0 +1,43 @@
...
+          setCurrentLanguage(detected)
+        }
+      } else {
+        setDetectedLang(language)
+      }
+      setIsLoading(false)
+    }
+
+    fetchPrevLesson()
+  }, [prevLessonId, setCurrentLanguage])
```
jamaljsr: nit: `language` should be in the deps array since it is referenced in the function.

jamaljsr (approved):
This is a good clean PR that resolves the issue and does a nice refactoring to reduce duplication in the components. I just have one small nit, but otherwise looks good.

I did a bit more digging to see if this issue impacts other chapters. I found that chapter 6 is the only place that the tabs are locked. I did find other instances where the default selected tab would switch back to JS after a page reload, ex: chapter 8 `building-blocks-3 -> building-blocks-4`. I guess this isn't a big issue since the user can switch to Python. I'm curious if you think it's worth finding and addressing the behavior in the other lessons with scripting challenges?

satsie:
@jamaljsr Thanks for the thoughtful review! If this is happening in other chapters I think it's worth at least filing an issue for it. Tuedon may not be able to work on it immediately because he has been fixing blocking issues in another project (BTC Transcripts), but if there are other contributors that want to give it a go, we can at least point them to this PR to show how it should be done. Good to know there is at least a manual workaround because the tabs are not locked!

content/lessons/chapter-6/hard/put-it-together-2.tsx
0tuedon: > https://github.com/saving-satoshi/saving-satoshi/pull/1416/changes#diff-0a160ab220b9a1fc037f6b72b78897ebea4a98eb488c3b4ce9cc69ef02a302aaR26
> Is there a mismatch between the normal and hard challenges for this chapter? Do they not go in the same order?

No, there isn’t a mismatch.

This can also happen if code from CH6PUT1_NORMAL is used as a default in the challenge, When that code is present, it can still be used to detect the language, which may make it look like the normal and hard challenges are out of order  but they aren’t.

0tuedon: > I tested this PR for the normal flow ([f615785](https://github.com/saving-satoshi/saving-satoshi/commit/f6157853fe4346b2db7445ceb2d66867367c72ff)) and it looks great. However, I am unable to test it for the hard mode because my account gets a 404 everytime I try to switch to hard mode, even with dev=true and even though I completed half of chapter 6 in hard mode (I also finished it in normal mode). I can start a new account from scratch to test hard mode if needed but wanted to ask first in case you have tested hard mode.


This is a Bug related to #1439 and  #1368

0tuedon:
> I did a bit more digging to see if this issue impacts other chapters. I found that chapter 6 is the only place that the tabs are locked. I did find other instances where the default selected tab would switch back to JS after a page reload, ex: chapter 8 `building-blocks-3 -> building-blocks-4`. I guess this isn't a big issue since the user can switch to Python. I'm curious if you think it's worth finding and addressing the behavior in the other lessons with scripting challenges?

Yes, I think a follow-up PR would be good to address this. It is an issue, because in cases like that the Python code can end up loading into the JavaScript tab, and then potentially get cleared when the user switches languages.

After this PR gets merged, we can use the current setup as a base to  fix this.

content/lessons/chapter-6/hard/put-it-together-2.tsx
satsie: Ok I trust you! The two paths for normal and hard mode in this chapter still make confused every time I look at them. I recall from playing both modes that the challenges are not a 1:1 match, and that hard mode is arguably easier than normal mode, haha. I am going to retest this in hard mode once it's merged to dev and has the changes from #1439. Will approve and merge now!

satsie (approved):
Thanks for sticking with this Tuedon, even with my long review cycles. Appreciate Jamal taking a look at this too. Merging!

content/lessons/chapter-6/hard/put-it-together-2.tsx
satsie: Addressed in https://github.com/saving-satoshi/saving-satoshi/pull/1441

Closing this comment!

PR #1435, opened 2026-01-12
Clarify that change address is pre-loaded in code (fixes #946)

### LINKED ISSUE #946: Chapter 6 "Sign the Transaction" challenge: Address discrepancy
In Chapter 6, challenge 2, "Sign the transaction", the instructions say that our change address is `bc1qm2dr49zrgf9wc74h5c58wlm3xrnujfuf5g80hs`:

Screenshot from 2024-05-15 22-53-20

But in the pre-loaded code, the change address is different (`bc1qkg62ae0wwntkzhq8td87s87c4nj5zdlj2ga8j7`) and there is a comment that we will be doing address reuse:
Screenshot from 2024-05-15 22-54-45

If I change the value of `addr` to the new address from the instructions, my solution does not pass. 

Easiest fix is probably to change the instructions. It could be an opportunity to link to Holocat/ChatBTC about why address reuse is bad. Address reuse could even come up later in the story! Maybe you get identified because of the address reuse. What if Mike Ramen's transaction history is revealed and someone is able to connect the dots that the person that gave Ramen funds for his flight is the same person that got mining rewards when BitRey was defeated (because maybe mining pools distribute funds from known addresses)... 

Anyways back to the issue. Was it originally intended for the player to have to replace the change address? Or should it just be corrected in the pre-loaded code? I found myself here because, similar to Jonas's feedback, I also got stuck on this exercise. It ended up being an oversight on my part. I was trying to pay 1 sat in fees instead of 1,000 :laughing:. My brain ignored the wonderful `...Don't forget to reduce our change from 61,000,000 to 60,999,000...` warning because the first time I saw it, it did apply to me, only I changed the value to 60,999,999 and thought I fixed the problem :facepalm: 

_Additional thoughts:_

I started littering the entire thing with debug log statements. I tore the transaction hex apart, reversed little endian buffers, and manually compared bytes, it was awful. I don't have a solution for this, but debugging some of these larger exercises is very difficult.

I'd like to go through some of these chapters and reword the instructions so it's abundantly clear what needs to be done. A checklist of things for the challenge may have helped me notice my oversight faster.

_Bottom line:_

What is the desired fix? I think these are the options:

1.  Update the address value in the pre-loaded code. Recompute the expected transaction hex and update the code that checks if the user's solution is correct.
2. Don't update the address value in the pre-loaded code because the user should have to do it. Recompute the expected transaction hex and update the code that checks if the user's solution is correct.
3. Change the instructions to remove the bit about the new address. Potentially add some resources about address reuse.

benknab:
I found this confusing when doing the challenge. Translations done by AI and verified with Google Translate since I don't speak these languages.

### DIFF
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -1595,7 +1595,7 @@ const translations = {
           },
         },
         bullet_one: `Mika 3000 gets 100,000,000 satoshis to bc1qgghq08syehkym52ueu9nl5x8gth23vr8hurv9dyfcmhaqk4lrlgs28epwj`,
-        bullet_two: `You get 61,000,000 back to your address bc1qm2dr49zrgf9wc74h5c58wlm3xrnujfuf5g80hs`,
+        bullet_two: `You get 61,000,000 back to your address (pre-loaded in the code)`,
         success: `You've done it! You've built a transaction!`,
       },
       hard: {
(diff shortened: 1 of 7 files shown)


satsie:
I'm not entirely sure if this is how we want to fix https://github.com/saving-satoshi/saving-satoshi/issues/946

satsie:
Well, enough time has passed that I've accepted we aren't going to resolve issue #946 with an enhanced user flow that explains why address reuse is bad... at least not anytime soon. Your changes have been incorporated into https://github.com/saving-satoshi/saving-satoshi/pull/1483 and since this PR has conflicts, I am going to close it in favor of #1483. Thanks for your patience regarding this!

PR #1437, opened 2026-01-29
Implement Global Toast Notifications for Network and Server Errors (#1421)

### LINKED ISSUE #1421: Feat: Implement User-Facing Error Feedback (Toasts) for Network & Server Issues
# Bug Reproduction: Silent Failure on Login

**Issue:** The application "swallows" network and server errors, leaving the user with no feedback when login fails.

## Steps to Reproduce
1. **Open Sign In:** Click the "Sign In" button or profile icon.
2. **Fill Private key with No internet or wrong private key:**
3. **Attempt login: No response or indication provided either NO internet, Invalid credentials or something**

## Observed Results
- UI: The loading spinner appears briefly and then vanishes, makes user confused and frustrated, as they are left unsure if their request was processed or why it silently failed.
- **Feedback:**  NO error message is displayed.
- **Console:** A red `NetworkError` appears in the console, proving the error occurred but was hidden from the user.

# Proposed Enhancement:
 **Currently we are receiving the error message in the form of object in console , my moto is to parse this error object and reason of error and display respective message this makes user experience more transparent and actionable, allowing them to immediately understand and resolve the issue (e.g. check internet vs. retry password) without guessing.**

https://github.com/user-attachments/assets/4bc4a2ac-3a2f-45bb-a773-3e889c4d07c7

 @satsie , @0tuedon 
 I have committed locally, Looking forward for your approval to raise PR

pranitaurlam:
This PR adds toast notifications to show clear error feedback to users.

Previously, network failures and server errors were either silent or only visible in the console. With this change, users now see immediate feedback when actions like sign-in or saving progress fail.

The toast system covers:

Network connectivity errors

4xx and 5xx API errors

Validation messages returned by the API

Toasts are managed globally, triggered from the fetch utility, and automatically dismiss after a short delay (or can be closed manually).

### DIFF
--- a/components/Layout.tsx
+++ b/components/Layout.tsx
@@ -4,6 +4,7 @@ import AccountModal from 'components/Modals/Account'
 import SignInModal from 'components/Modals/SignIn'
 import SignUpModal from 'components/Modals/SignUp'
 import HelpModal from 'components/Modals/Help'
+import { ToastContainer } from 'components/ToastContainer'
 import { useAtom, useSetAtom } from 'jotai'
 import { useEffect } from 'react'
 import { useAuthFunctions } from 'state/AuthFunctions'
@@ -45,6 +46,7 @@ export default function Layout({ children }: { children?: React.ReactNode }) {
         onClose={() => close(Modal.Account)}
       />
       <HelpModal state={modals[Modal.Help]} onClose={() => close(Modal.Help)} />
+      <ToastContainer />
     </>
   )
 }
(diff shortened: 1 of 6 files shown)


satsie:
Can you please include a video of how this works? This is probably something that should be run by design first.

aniket866:
> Can you please include a video of how this works? This is probably something that should be run by design first.

I have this committed locally and tested and also had attached video there #1421

PR #1443, opened 2026-02-07
Fix incorrect wording on CH6 Prepare the message

### LINKED ISSUE #1418: Python: Chapter 5, Validate the signaturePrepare the message
The instructions ask to convert the hash into an integer but the final output needs to be a hex string.

>Then we will double SHA-256 hash that blob of data, and convert that hash into an integer. Complete the function encode_message(). It should return a 32-byte hex value.

If we need to convert the integer back into a hex string it seems like we can skip that step?

```
import hashlib
text = "I am Vanderpoole and I have control of the private key Satoshi\n"
text += "used to sign the first-ever Bitcoin transaction confirmed in block #170.\n"
text += "This message is signed with the same private key."

def encode_message(text):
    prefix = "Bitcoin Signed Message:\n".encode()
    text_bytes = text.encode()
    msg_bytes = len(prefix).to_bytes() + prefix + len(text_bytes).to_bytes() + text_bytes
    hash1 = hashlib.sha256(msg_bytes).digest()
    hash2 = hashlib.sha256(hash1).hexdigest()
    
    hash2_int = int(hash2, 16) # convert to int
    
    return hash2 # but I need to return the hexdigest

```

Should remove the instruction to convert to an integer:
>Then we will double SHA-256 hash that blob of data, and return a 32-byte hex value.

jamaljsr:
Closes #1418 

On `/chapter-5/validate-signature-1`, the copy below is not accurate. 
> Then we will double SHA-256 hash that blob of data, and convert that hash into an integer. Complete the function encode_message(). It should return a 32-byte hex value.

I have updated it to remove the portion about converting to an integer because that is not necessary.
> Then we will double SHA-256 hash that blob of data. Complete the function encode_message(). It should return a 32-byte hex value.

image

## Pull Request checklist

- [x] Clean commit history

### DIFF
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -924,7 +924,7 @@ const translations = {
         pre_link: `Then we will`,
         highlighted: `double SHA-256 hash`,
         question: `Why do we double hash in bitcoin?`,
-        post_link: `that blob of data, and convert that hash into an integer. Complete the function <span className="text-green p-1 font-mono bg-[#0000004D] m-0.5 text-sm">encode_message()</span>. It should return a 32-byte hex value.`,
+        post_link: `that blob of data. Complete the function <span className="text-green p-1 font-mono bg-[#0000004D] m-0.5 text-sm">encode_message()</span>. It should return a 32-byte hex value.`,
       },
       success: `Nicely Done`,
     },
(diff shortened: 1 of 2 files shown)


0tuedon (approved):
Copy changes LGTM!

PR #1444, opened 2026-02-07
Fix misleading instructions on CH4 - Get a P2WPKH address

### LINKED ISSUE #1417: Chapter 4, Address Get a P2WPKH address Python
**Describe the bug**
The instructions say to append `0` to the hash and then convert the witness program using the bech32 library:
>First we need to append a witness version number of '0' to the hash. These resulting 21 bytes are known as the witness program.
>Start by making the witness program, then convert the program to an address by using the bech32 library that has already been imported for you.

However, if you add the '0' to the hash before passing it to the `encode` function it fails. The encode function takes the witness version '0' as an argument. You can see it takes the public key hash bytes directly: 

```
# encode
public_key_hash = bytes.fromhex("8bf743f3fd7f46804c39711af735b121747ef6b4")
assert bech32.encode('bc', 0, public_key_hash) == "bc1q30m58ula0argqnpewyd0wdd3y968aa457pksa3"
```
[https://github.com/saving-satoshi/bech32py/blob/main/test.py](https://github.com/saving-satoshi/bech32py/blob/main/test.py)

**To Reproduce**
Adding the 0 to create the witness program first does not work:
```
def hash_to_address(hash):
    witness_program_hex = hash.hex() + '00'
    witness_program_bytes = bytes.fromhex(witness_program_hex)
    return encode('tb', 0, witness_program_bytes)
```

Just need to pass the hash straight to the function with the correct arguments:
```
def hash_to_address(hash):
    return encode('tb', 0, hash)
```

**Expected behavior**
I think the instructions are a bit misleading when they ask "Start by making the witness program"

Suggest removing that part and simply changing it to:
>Convert the program to an address by using the bech32 library that has already been imported for you.

Unless it's supposed to work differently?

jamaljsr:
Closes #1417

On `/chapter-4/address-3`, there are misleading instructions that can be confusing. It makes it seem like the user must append the version before encoding which isn't correct. The `bech32` encoder will do the concatenation for them. They just need to supply the parameters correctly.

I have updated the copy to be a bit more clear about what the user is expected to do.

### Before

image

### After

image

## Pull Request checklist

- [x] Clean commit history

### DIFF
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -685,7 +685,7 @@ const translations = {
       title: `Address`,
       nav_title: `Get a P2WPKH address`,
       paragraph_one: `There are multiple types of bitcoin addresses. In the previous exercise, we created a 20-byte compressed public key hash. Now, we would like to encode that hash into a Pay-to-Witness-Public-Key-Hash (p2wpkh) address on the Testnet network.`,
-      paragraph_two: `First we need to append a witness version number of '0' to the hash. These resulting 21 bytes are known as the <span className="font-bold">witness program</span>.`,
+      paragraph_two: `First, note that for a p2wpkh address the witness version is '0', and the witness program is the 20-byte public key hash from the previous step.`,
       paragraph_three: `Then, the witness program is encoded into a human-friendly format called <Link href="https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#user-content-Specification" target="_blank" className="underline">bech32</Link>. Doing this appends a human-readable prefix and a checksum to the data.`,
       paragraph_four: `The prefix is determined by the network:`,
       table_heading: {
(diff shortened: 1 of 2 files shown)


0tuedon (approved):
LGTM!

satsie:
@jamaljsr Thanks for getting to the bottom of this! Can we bring back a sentence that tells the user what a witness program is? It's terminology that they will likely run into later (hopefully in the wonderful world of BOSS) and I want to make sure we define it at some point in the game.

jamaljsr:
@satsie So the original message was actually incorrect. It stated: 

> First we need to append a witness version number of '0' to the hash. These resulting 21 bytes are known as the witness program.

Technically, the first byte is the version and the remaining 2-40 bytes are defined as the witness program". In this case, it's 20 bytes denoting it's a `p2wpkh` address where the data is a hash of the public key. 

The text I changed it to more accurately defines the witness program as the 20 bytes following the prefix.

> First, note that for a p2wpkh address the witness version is '0', and the witness program is the 20-byte public key hash from the previous step.

I'd be more than happy to change this copy if you can think of a better way to phrase this.

satsie:
@jamaljsr ah, got it! that makes sense to me and I don't think any further updates are needed then. Thank you for the clarification!

PR #1445, opened 2026-02-08
Prevent progress when saving lesson data fails

### LINKED ISSUE #1353: Blank page and 500 error when progressing Chapter 6 in Hard mode (in-out-4)
**Describe the bug**
When progressing through **Chapter 6 in Hard mode**, the app returned a blank page and failed to save progress at the `in-out-4-hard` step. This was caused by a missing hard-mode progress code (`CH6INO4_HARD`) for the account, which resulted in backend 500 errors and blocked lesson progression.

---

**To Reproduce**

1. Enable **Hard mode**
2. Go to Chapter 6
3. Complete the lesson up to **in-out-4-hard**
4. Attempt to proceed to the next step
5. Page turns blank and progress does not save

---

**Expected behavior**
Progress for `in-out-4-hard` should be saved successfully and the user should be redirected to the next lesson without errors.

---

**Screenshots and/or logs**

Browser console shows repeated 500 errors:

```
Failed to load resource: the server responded with a status of 500 (Internal Server Error)
api.savingsatoshi.com/v1/progress

Invalid `prisma.accounts_progress.update()` invocation
Uncaught Error: Attempting to use a disconnected port object
```

(See attached screenshot)
Image

Image

---

* OS: Ubuntu Linux
* Browser: Chromium-based (Edge)
* Version: Latest (at time of issue)

---

The issue was resolved after a maintainer told me to solve  the missing `CH6INO4_HARD` lesson again. This suggests hard-mode lesson progress records may not always be initialized correctly.

jamaljsr:
Closes #1353

When a user completes certain lessons, the code is saved on the backend so that future lessons can fetch it and pre-populate its initial code with a portion of the code from the prior lesson(s). The way this is implemented now is a fire-and-forget approach. It calls [setData](https://github.com/saving-satoshi/saving-satoshi/blob/dev/ui/lesson/ScriptingChallenge/index.tsx#L189) and ignores any response from the backend. So if the API call fails (500 error) for any reason, the UI would still behave like the lesson succeeded, allowing users to move forward with backend data missing. Later lessons that depend on these saved lessons will then hit the issue described in #1353, blank pages.

This PR makes a small tweak to the flow by gating progression on a successful save to the backend. If there are any errors, an error message will be surfaced to the user immediately and they must try again to proceed. This will prevent users from advancing with unsaved state that will break future lessons. The message below is displayed in the output section. I'm open to suggestions on how to improve the wording.

> Your submission was correct but we were unable to persist the code to the backend. It is needed for future lessons. Please try again.

image

It is my assumption that failures to save data to the backend are transient, possibly due to load, that will typically resolve themselves after waiting some time. We just need to prevent users from progressing for that time to prevent them from getting into a invalid state which requires manual server-side intervention to resolve.

## Steps to test

1. Update the backend `PUT /v1/data` endpoint to throw an error
2. In a browser, navigate to a coding lesson that saves data. Ex: `/chapter-6/in-out-4-hard`
3. Submit a correct solution
4. Confirm the error is displayed
5. Undo the backend change
6. Submit the answer again and confirm it allows you to proceed 

## Pull Request checklist

- [x] Clean commit history

### DIFF
--- a/api/data/setData.ts
+++ b/api/data/setData.ts
@@ -1,11 +1,21 @@
 import { StoredLessonData } from 'types'
 import { put } from 'utils'
 
+const SAVE_FAILURE_MESSAGE =
+  'Your submission was correct but we were unable to persist the code to the backend. ' +
+  'It is needed for future lessons. Please try again.'
+
+export type SaveDataResult = {
+  ok: boolean
+  message?: string
+  data?: any
+}
+
 export default async function setData(
   account_id: number,
   lesson_id: string,
   data: StoredLessonData
-) {
+): Promise<SaveDataResult> {
   try {
     const res = await put({
       url: '/v1/data',
@@ -20,9 +30,18 @@ export default async function setData(
       },
     })
 
-    return res
+    return {
+      ok: true,
+      data: res,
+    }
   } catch (errors) {
-    console.error(errors)
-    return undefined
+    // log the detailed error from the backend for debugging
+    console.error(SAVE_FAILURE_MESSAGE, errors)
+
+    // Return a more helpful message to display to the user
+    return {
+      ok: false,
+      message: SAVE_FAILURE_MESSAGE,
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


0tuedon (approved):
TACK, I did make the backend throw an error  in the code save api route and this worked

Image

PR #1447, opened 2026-02-10
Prevent simple workarounds in CH5VLS4

### LINKED ISSUE #1145: Able to skip past 5.4 with workaround
screen
Thank you 🙇‍♀ for wanting to create an issue in this repository. Before you do, please ensure you are filing the issue in the right place. Issues should only be opened on if the issue **relates to code in this repository**.

If your issue is relevant to this repository, please include the information below:

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots and/or logs**
If applicable, add screenshots and/or logs to help explain your problem.

**Smartphone (please complete the following information):**

**Additional context**

jamaljsr:
Closes #1145 

There's a problem in `/chapter-5/validate-signature-4` where the user could simply use `return key[2]` as their solution and it would pass. I tightened the validation so it’s harder to pass with a hardcoded answer while keeping the lesson flow the same. Instead of validating a single run, it now calls `verify_keys` twice for both JavaScript and Python: one with a valid list that should return Vanderpoole’s key, and one with a different list (same length, but without the real key) that should return no match. The validator only passes when both conditions are true, which blocks easy bypasses like always returning the known key or branching on array length.

image

## Pull Request checklist

- [x] Clean commit history

### DIFF
--- a/content/lessons/chapter-5/validate-signature-4.tsx
+++ b/content/lessons/chapter-5/validate-signature-4.tsx
@@ -1,12 +1,12 @@
 'use client'
 
-import { useTranslations } from 'hooks'
 import { useState } from 'react'
+import { useTranslations } from 'hooks'
+import { useAtom } from 'jotai'
+import { getLanguageString } from 'lib/SavedCode'
+import { currentLanguageAtom } from 'state/state'
 import { EditorConfig } from 'types'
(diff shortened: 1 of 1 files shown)


0tuedon (commented):
I was able to test this in different ways to make sure the bypass doesn't exist, Looks good!

TACK!

0tuedon (approved):

PR #1449, opened 2026-02-12
Improve reset button in CH9 two-step opcode challenges

### LINKED ISSUE #1382: Make assignment in "Chapter 9 - Work with an oracle" clearer
On page: https://savingsatoshi.com/en/chapters/chapter-9/proposal-4

The assigment says: "Provide the initial stack to spend from the script."

The issue here is
a) you have to provide 2 initial stacks, not one
b) only one order is accepted, the other rejected without further explanation:

1. SIG(VANDERPOOLE) 1
2.  SIG(ME) FD3771E8 0
is accepted, but not the reverse order 2.), 1.)

**Describe the enhancement**
Reword the assigment as "Provide the initial stack to spend as VANDERPOOLE first, then the initial stack you would use when spending yourself."
Additionally: accept both orders of initial stacks.

### LINKED ISSUE #1383: Make assignment in "Chapter 9 - Make him wait clearer
On page: https://savingsatoshi.com/en/chapters/chapter-9/proposal-3

Issue: Possible to get stuck here for a bit due to unclear instructions. It is easily overlooked that the block height becomes editable after entering the first initial stack.

**Describe the enhancement**
Reword the assignment "Provide the initial stack to spend from the script" to "Provide the initial stack to spend from the script yourself first. In the next step, provide the script for Vanderpoole to spend, and adjust the block height.

jamaljsr:
Closes #1382 & #1383

Chapter 9 `proposal-3` and `proposal-4` are two-step OpCode Challenges which contain conditional `OP_IF` scripts. First, the user submits their script and initial stack. If it is correct, they are prompted to provide a second stack and execute the script again. The confusion comes in because on the second step there is a large Reset button when the Next button normally lives. So user's instincts are to click this primary CTA, which discards their step 1 progress.

As suggested in https://github.com/saving-satoshi/saving-satoshi/issues/1383#issuecomment-3728562644, the Figma designs show a more subtle Reset button in the Execution Stack section. 

I have updated the OpCode Challenge components to move the Reset button to match the Figma design. When the user completes step 1, the large primary CTA does not become visible. They must update the Initial Stack field above then click the play button again. If they complete step 2, then the large Next button becomes visible and they can proceed to the next lesson.

I have also updated the instructions on the left panel to be more clear.

I have tested all of the lessons in chapter 9 to ensure that these changes did not break any of the other lessons.

## Screenshots

`/chapters/proposal-3` Step 1 completed

image

`/chapters/proposal-3` Step 2 completed

image

`/chapters/proposal-1` Step 1 completed

image

## Pull Request checklist

- [x] Clean commit history

### DIFF
--- a/i18n/locales/en.ts
+++ b/i18n/locales/en.ts
@@ -2163,8 +2163,8 @@ const translations = {
       heading: `Conditional time locked transaction`,
       paragraph_one: `Wait a minute, that doesn't make sense—you don't want to deal with him forever! The new deal is you get all donations for the next two hours while you are still on TV. The Lil Bits Foundation gets anything that comes in afterwards. You look at the bitcoin block clock on the wall in the studio and agree that block height 6930300 will probably be mined in about two hours.`,
       paragraph_two: `Remember Vanderpoole's public key is PUBKEY(vanderpoole) and yours is PUBKEY(me).`,
-      paragraph_three: `Provide the initial stack to spend from the script.`,
-      next_step_message: `Looks good! Now lets try with your own signature.`,
+      paragraph_three: `Provide two initial stacks in sequence: first, enter the stack you would use to spend yourself. After that succeeds, enter Vanderpoole's stack and run again.`,
+      next_step_message: `Nice first spend. Now enter Vanderpoole's stack.`,
     },
     proposal_four: {
       title: `Secret preimage locked transaction`,
@@ -2177,13 +2177,13 @@ const translations = {
         b: ` one second earlier!`,
       },
       paragraph_four: `Remember Vanderpoole's public key is PUBKEY(vanderpoole) and yours is PUBKEY(me).`,
-      paragraph_five: `Provide the initial stack to spend from the script.`,
+      paragraph_five: `Provide two initial stacks in sequence: first Vanderpoole's spending stack, then your stack with the revealed preimage.`,
       tooltip_one: {
         question: `What is a satoshi?`,
         link: `https://chat.bitcoinsearch.xyz/?author=holocat&question=What%2520is%2520a%2520satoshi%253F`,
         highlighted: `satoshi`,
       },
-      next_step_message: `Let's see if we used our signature with the preimage correctly.`,
+      next_step_message: `Great first spend. Now enter your stack with the preimage.`,
     },
     outro_one: {
       title: `Outro`,
--- a/ui/common/StatusBar.tsx
+++ b/ui/common/StatusBar.tsx
@@ -26,6 +26,7 @@ export interface StatusBarType {
   errorMessage?: string
   nextStepMessage?: string
   nextStepButton?: string
+  hideNextStepBtn?: boolean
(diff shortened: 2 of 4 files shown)


i18n/locales/en.ts
```diff
@@ -2163,8 +2163,8 @@ const translations = {
       heading: `Conditional time locked transaction`,
       paragraph_one: `Wait a minute, that doesn't make sense—you don't want to deal with him forever! The new deal is you get all donations for the next two hours while you are still on TV. The Lil Bits Foundation gets anything that comes in afterwards. You look at the bitcoin block clock on the wall in the studio and agree that block height 6930300 will probably be mined in about two hours.`,
       paragraph_two: `Remember Vanderpoole's public key is PUBKEY(vanderpoole) and yours is PUBKEY(me).`,
-      paragraph_three: `Provide the initial stack to spend from the script.`,
-      next_step_message: `Looks good! Now lets try with your own signature.`,
+      paragraph_three: `Provide two initial stacks in sequence: first, enter the stack you would use to spend yourself. After that succeeds, enter Vanderpoole's stack and run again.`,
```
0tuedon, line 2166: I am just worried if this isn't close to what the spoiler says

0tuedon (approved):
Tested, and this works great. Thanks @jamaljsr.

I did leave some comments about the side text being a spoiler, otherwise this looks good to go.

i18n/locales/en.ts
jamaljsr: I wasn't sure how explicit we wanted to be in the instructions. I'd be happy to change it to what you think is best, or revert it back to the way it was. The UX improvement is the primary focus of this PR. The copy changes are less important.

@satsie Do you have any thoughts?

satsie: Since this lesson is the number 1 thing we got constructive feedback on, I think making the instructions as explicit as possible is the right move. If we were to change anything I guess we could remove the redundancy from the spoiler, but if someone still needs spoilers after these instructions, maybe it's good that they are redundant 😅

satsie:
Issue https://github.com/saving-satoshi/saving-satoshi/issues/1382 has a feature request to:

> Additionally: accept both orders of initial stacks.

We aren't doing this, right? It seems like it would really complicate the exercise and the new copy in this PR makes it clear the order that the initial stacks need to be in.

satsie (approved):
tACK! This is great. I tested both lessons and this is a major UX improvement. I didn't realize how problematic the reset button was there until I was able to use this new version. Everything is a lot smoother now and I'm happy to see it reflect Christoph's original designs. It feels good to finally have this issue resolved!

jamaljsr:
> We aren't doing this, right? It seems like it would really complicate the exercise and the new copy in this PR makes it clear the order that the initial stacks need to be in.

@satsie yes, supporting both orders would make this lesson's component far more complex. I feel like the UX would also be more confusing. The copy change feels like the better trade-off in my opinion.

PR #1453, opened 2026-02-14
Fix signature verification in CH6PUT2_NORMAL

### LINKED ISSUE #1376: Chapter 6 Build the witness class unchecked function
**Describe the bug**
The testcase only accounts for compute_input_signature(index, key), sign_input(index, priv, pub, sighash) is never checked to be valid or functional. Completing compute_input_signature() and submitting no sign_input() code passes the challenge due to this validation bug. 

**To Reproduce**

1. Go to Chapter 6
2. Click on Build the witness class
3. Scroll down to sign_input(index, priv, pub, sighash) function

**Expected behavior**
The validation logic should not only verify the ECDSA signature math in compute_input_signature() but also verify that the sign_input() method correctly populates the transaction's witness data. Specifically:

- The sign_input() function should be called during the validation block.
- The validator should check that tx.witnesses[index] contains a valid Witness object.
- The validator should verify that the serialized signature within that witness includes the correct DER encoding and the trailing SIGHASH byte.

**Screenshots and/or logs**
Image

**Additional context**
To fix the problem, navigate to content/lessons/chapter-6/normal/put-it-together-2.tsx and update the javascript.program (and the Python equivalent) to execute and verify sign_input.

You can replace the final console.log logic in the validation block with something like this: 
```// Call the method that was previously unchecked
tx_eagmcued.sign_input(0, priv_dsljfohd, compressed_pub_agfwuebb);

// Check if the witness was actually created
const witness = tx_eagmcued.witnesses[0];
const hasWitness = witness && witness.items.length === 2;

// Verify the signature math
const [r, s] = tx_eagmcued.compute_input_signature(0, priv_dsljfohd);
const hashed_message_bytes = tx_eagmcued.digest(0);
const signature_valid = private_key_faxwmufa.verify(hashed_message_bytes, {r: r.toString(16), s: s.toString(16)});

// Final success condition requires both valid math AND a functional sign_input method
console.log(hasWitness && signature_valid && 'true');```

jamaljsr:
Closes #1376

This fixes a validation issue in `/chapter-6/put-it-together-2` where `sign_input()` was being called but not properly validated.

Previously:
- JavaScript called `sign_input()` but only validated `compute_input_signature()`, so a no-op `sign_input` could still pass.
- Python didn’t call `sign_input()` in the validation block at all.

I updated both JS and Python validation blocks so the challenge now requires both: 1) correct signature math from `compute_input_signature()` and 2) correct witness creation from `sign_input()`. Passing the challenge now requires both signature verification and witness verification to pass.

I tested this with:
- correct spoiler code (passes)
- intentionally incorrect/no-op `sign_input` code (fails)

## Pull Request checklist

- [x] Clean commit history

### DIFF
--- a/content/lessons/chapter-6/normal/put-it-together-2.tsx
+++ b/content/lessons/chapter-6/normal/put-it-together-2.tsx
@@ -150,13 +150,28 @@ const output_uhmhvgcw = Output.from_options(addr_pqvejvea, valueTwo_jhcermcr);
 const tx_eagmcued = new Transaction();
 tx_eagmcued.inputs.push(input_bauoevbs);
 tx_eagmcued.outputs.push(output_uhmhvgcw);
-const [r, s] = tx_eagmcued.compute_input_signature(0, priv_dsljfohd);
+tx_eagmcued.sign_input(0, priv_dsljfohd, compressed_pub_agfwuebb);
 const EC = require('elliptic').ec;
 const ec = new EC('secp256k1');
+const witness_ogyehdag = tx_eagmcued.witnesses[0];
+const witness_shape_valid_ayqszxcg = !!(
+  witness_ogyehdag &&
+  witness_ogyehdag.items.length === 2 &&
+  witness_ogyehdag.items[0].length > 0 &&
+  witness_ogyehdag.items[0][witness_ogyehdag.items[0].length - 1] === 0x01 &&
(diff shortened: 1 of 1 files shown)


0tuedon (approved):
LGTM!, Tested without sign_input and it fails in both languages

PR #1459, opened 2026-02-20
Add unit tests for progressState jotai atoms

jamaljsr:
This is the first of two PRs to refactor `state/progressState.ts`. This one adds test coverage for the existing behavior so we can verify that the upcoming refactor does not break anything.

### Migrate from Jest to Vitest

I replaced Jest + ts-jest with Vitest. The main motivation is that Vitest reads `tsconfig.json` path aliases natively via `vite-tsconfig-paths`, which eliminates the need for manual `moduleNameMapper` configuration. The `describe`/`expect`/`it` API is identical so the existing `lib/SavedCode.test.ts` required no changes beyond a single import. Vitest also natively supports ESM modules, which `jotai` uses, while Jest requires a bit of hacky configuration to get ESM working properly. Overall Jest is a more modern testing framework. I felt it would be ok to switch since we don't really have any existing unit tests in the project.

### Unit tests

I added two test files covering the public API of `progressState.ts`:

**`test/unit/progressState.utils.test.ts`** — Pure utility function tests (no Jotai needed):

- `getLessonById` — finds lessons across chapters, handles difficulty chapters, returns null for unknown IDs
- `isLastLesson` — checks last lesson of last chapter (with and without difficulty)
- `isChapterInProgress` — first lesson completed but not all
- `mergeProgressState` — merges backend state into defaults, handles missing chapters/lessons
- `getLessonKey` — maps chapter+lesson to key
- `isLessonCompletedUsingId` / `isLessonCompletedUsingLessonName` — completion checks
- `isLessonUnlockedUsingId` / `isLessonUnlockedUsingLessonName` — unlock logic including cross-chapter gating
- `getNextLessonUsingChapterIdAndLessonName` — next in chapter, cross-chapter, end of course

**`test/unit/progressState.atoms.test.ts`** — Jotai atom tests using `createStore()`:

- `currentLessonComputedAtom` — returns first incomplete lesson, null when all complete
- `nextLessonAtom` — next lesson including cross-chapter transitions
- `currentLessonPathAtom` / `nextLessonPathAtom` — URL path generation
- `progressToNextLessonAtom` — marks current lesson complete and advances, including the `presentPageAtom` gating behavior
- `markLessonAsCompleteAtom` — marks specific lesson complete and finds next
- `setChapterDifficultyAtom` — updates `selectedDifficulty` for a chapter

Both test files use a minimal `CourseProgress` fixture (`test/unit/fixtures/progressFixtures.ts`) with 3 chapters (one with difficulty, two without) to keep the tests readable.

### CI

I added a Github Workflow file that will run the unit tests on every pull request in order to verify that the PR changes do not break the tests.

## Pull Request checklist

- [x] Clean commit history

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,0 +1,20 @@
+name: Unit Tests
+
+on:
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: yarn
+
+      - run: yarn install --frozen-lockfile
+
+      - run: yarn test
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "lint": "next lint",
     "prettier": "prettier --write --ignore-unknown .",
     "prettier:check": "prettier --check --ignore-unknown .",
-    "test": "yarn prettier:check && yarn lint && yarn jest",
+    "test": "yarn vitest run",
     "postinstall": "husky install"
   },
   "lint-staged": {
@@ -47,7 +47,6 @@
   },
   "devDependencies": {
     "@svgr/webpack": "^6.5.1",
-    "@types/jest": "^29.5.11",
     "@types/node": "22.0.0",
     "@types/react": "18.0.25",
     "autoprefixer": "^10.4.13",
@@ -56,14 +55,14 @@
     "eslint-config-prettier": "^8.6.0",
     "eslint-plugin-local-rules": "^3.0.2",
     "husky": "^8.0.2",
-    "jest": "^29.7.0",
     "json-loader": "^0.5.7",
     "lint-staged": "^13.1.0",
     "postcss": "^8.4.18",
     "prettier": "^2.8.1",
     "prettier-plugin-tailwindcss": "^0.2.1",
     "tailwindcss": "^3.2.2",
-    "ts-jest": "^29.1.1",
-    "typescript": "4.8.4"
+    "typescript": "4.8.4",
+    "vite-tsconfig-paths": "^6.1.1",
+    "vitest": "^4.0.18"
   }
 }
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,4 +1,0 @@
-module.exports = {
-  preset: 'ts-jest',
-  testEnvironment: 'node',
-}
--- a/lib/SavedCode.test.ts
+++ b/lib/SavedCode.test.ts
@@ -1,4 +1,5 @@
 // organizeImports.test.ts
+import { describe, expect, it } from 'vitest'
 import { organizeImports } from './SavedCode'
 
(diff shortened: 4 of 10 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "lint": "next lint",
     "prettier": "prettier --write --ignore-unknown .",
     "prettier:check": "prettier --check --ignore-unknown .",
-    "test": "yarn prettier:check && yarn lint && yarn jest",
+    "test": "yarn prettier:check && yarn lint && yarn vitest run",
```
0tuedon: Could we make this just be test, I think there's a backlog of unformatted files in i18n and in few content folders, I could create a PR for that

0tuedon (approved):
I have been able to look through this test and seen the good work at @jamaljsr , Thank you for your contributions, I do have a comment about the test script having linting inside, i think we have a backlog of linting to do.

package.json
0tuedon: The test breaks for me when I run yarn test because of the linting, would you want me to create a PR for this backlog of unlinted file

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 24
```
0tuedon: Can we make the node version 20? So we have everything in sync just like with the e2e

0tuedon (changes requested):

package.json
jamaljsr: Ok, I can change this to just run the tests. I also saw the linting errors but felt it was out of scope for this PR to fix them.

Yes, that would be good to get all those resolved.

jamaljsr: I have updated the `yarn test` command.

.github/workflows/unit-tests.yml
jamaljsr: I changed the node version to 20.

jamaljsr:
@0tuedon Thank you for the review. I have made the requested changes. Please take another look at this when you have a chance.

0tuedon (approved):
LGTM! Thank you @jamaljsr

package.json
satsie: Thanks for bringing this up @0tuedon. Good catch! I created a ticket for the backlog of unformatted files and assigned it to you. I gave it a P1 priority. It's not immediately urgent, but it's something we should fix sooner than later so we can add the linter back to the test command, and therefore to all PRs that are opened.

https://github.com/saving-satoshi/saving-satoshi/issues/1464

satsie:
> Overall Jest is a more modern testing framework. 

Did you mean to say Vitest is a more modern testing framework?

Post-merge ACK. Great idea to have something like this together before the refactor!

jamaljsr:
> > Overall Jest is a more modern testing framework.
> 
> Did you mean to say Vitest is a more modern testing framework?

Ha! yes, I did. Good catch.

PR #1472, opened 2026-02-28
Fix code persistence in ScriptingChallenge

### LINKED ISSUE #1467: Code not persistent
I've been doing all of my coding locally or in collab because my code is never reliably persistent on the site.

If I open a random completed challenge right now: https://savingsatoshi.com/en/chapters/chapter-6/in-out-5

This is the code:
```
class Output:
    def __init__(self):
      self.value = 0
      self.witness_version = 0
      self.witness_data = b""

    @classmethod
    def from_options(cls, addr: str, value: int):
        assert isinstance(value, int)
        self = cls()
        # YOUR CODE HERE
        return self

    def serialize(self):
        # YOUR CODE HERE
```

Here are my localstorage entries for CH6

Image

Maybe it should be CH6INO5:
```
CH6INO5-python:Array
0:"from struct import pack
from bech32py import bech32
# Use the bech32 library to find the version and data components from the address
# See the library source code for the exact definition
# https://github.com/saving-satoshi/bech32py/blob/main/bech32py/bech32.py

class Output:
    def __init__(self):
      self.value "1:" 0
      self.witness_version "2:" 0
      self.witness_data "3:" b""

    @classmethod
    def from_options(cls, addr: str, value: int):
        assert isinstance(value, int)
        self "4:" cls()
        # YOUR CODE HERE
        return self

    def serialize(self):
        # YOUR CODE HERE
"
```

jamaljsr:
Closes #1467

There is an issue where in-progress code is lost when refreshing the page or navigating to a challenge via Continue/Next. The code is saved to React state (so browser back works) but gets lost from localStorage on remount.

I have found two bugs in the `ScriptingChallenge ` component:

**Bug 1: Autosave overwrites saved drafts on mount.** The autosave effect runs on mount with `debouncedCode` still set to `defaultCode`, writing the starter code to localStorage before the load effect's `setCode` call takes effect.

**Bug 2: Language atom hydration race.** `currentLanguageAtom` uses `atomWithStorage`, which returns JavaScript (the default) on the first render before asynchronously hydrating from localStorage. For Python users, the component reads from the wrong storage key (`CH6INO5-javascript` instead of `CH6INO5-python`), finds nothing, and shows default code. The local `language` state is initialized once via `useState` and never updates when the atom hydrates.

To fix this:

1. I added a `hasLoaded` ref that stalls the autosave effect. It won't write to localStorage until after the initial hydration read completes.
2. I added a sync effect that watches `currentLanguageAtom` and updates local `language` state when the atom hydrates, loading the correct per-language draft.

### Testing

1. Open any scripting challenge (e.g. `chapter-4/public-key-4`), write some code, wait a couple seconds for autosave
2. Refresh the page and confirm draft should be restored
3. Navigate away via chapter menu, then back and confirm draft should be restored
4. Go to the previous lesson (e.g. `public-key-3`), click Continue and confirm the code should be your saved draft
5. Use the editor refresh/reset button and confirm it reverts to starter code
6. Repeat steps 1-4 using Python instead of JavaScript

## Pull Request checklist

- [x] Clean commit history

### DIFF
--- a/ui/lesson/ScriptingChallenge/index.tsx
+++ b/ui/lesson/ScriptingChallenge/index.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { useEffect, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 
 import LanguageTabs from './LanguageTabs'
 import Editor from './Editor'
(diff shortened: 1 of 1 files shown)


satsie:
Wow, this is a really difficult bug to diagnose. I can't believe you were able to find two underlying causes, one of them being a race condition. I tried to reproduce it in prod and I was only able to observe the behavior with Python. Should I have also been able to reproduce it for JavaScript? If so I can try it a few more times.

That being said, I tested this branch with the instructions you provided using both JavaScript and Python and can confirm that my code was persisted correctly each time.

TACK cd22a897943c06c224d2fb65cf8811e5d5bb1cdb

jamaljsr:
@satsie Thanks for the review

> I tried to reproduce it in prod and I was only able to observe the behavior with Python. Should I have also been able to reproduce it for JavaScript? If so I can try it a few more times.

I was only able to repro the the issue with JS in the case where I click Continue from the previous lesson. It's when I started testing Python that the other bugs cropped up. Then I started tracing the code paths with `console.log`s. That's where the race condition became obvious.

0tuedon (approved):
On dev and live I was able to observe the issue there, for me python was cause by the Tab Switching and losing state progress there which can be frustrating, I tested that aspect on this PR and it works really well. Thank you @jamaljsr 
TACK

satsie (approved):
I tried to reproduce the issue again on prod with Javascript and still didn't see it. However I think that between the three of us, we have each observed and tested enough to have the confidence that this PR works as expected and fixes the code persistence. Merging!

PR #1460, opened 2026-02-20
Refactor progress state

jamaljsr:
Depends on #1459 
Closes #1458

`state/progressState.ts` is over 1200 lines and difficult to reason about. It mixes hardcoded lesson data, ~10 copies of identical difficulty-branching logic, a 4-layer atom dependency chain, and pure utility functions all in one file. This PR splits it into focused modules while preserving the exact same behavior and public API.

### New file structure

```
state/progress/
  defaultProgress.ts   — DifficultyLevel enum + defaultProgressState constant
  helpers.ts           — Pure functions that operate on chapters
  atoms.ts             — Core atoms + sync effect
  selectors.ts         — Read-only derived atoms
  actions.ts           — Write-only action atoms
  utils.ts             — Higher-level functions that operate on the full state (`CourseProgress`)
  index.ts             — Barrel re-export of all public symbols
state/progressState.ts — Now just: export * from './progress'
```

### Key changes

**Deduplicated difficulty-branching logic** : The ~10 copies of `if (chapter.hasDifficulty) { find difficulty lessons } else { use chapter.lessons }` scattered across functions are replaced by two shared helpers in `helpers.ts`: `getChapterLessons()` (returns lessons for the selected difficulty) and `getAllChapterLessons()` (returns lessons across all difficulties). Every function that previously inlined this logic now calls one of these helpers.

**Simplified atom chain** : The previous 4-layer chain (`courseProgressAtom → combinedProgressAndAccountAtom → syncCourseProgressEffectAtom → syncedCourseProgressAtom`) is reduced to 3 by eliminating `combinedProgressAndAccountAtom`. The sync effect now calls `get()` on `accountAtom` and `isAuthLoadingAtom` directly.

**Code Comments**: I added a lot of code comments explaining what each function does in detail as well as many inline comments where I felt the code wasn't obvious. This may feel a bit excessive but I initially struggled to understand how all of these worked, so chose to over-explain for future devs.

**No consumer changes**: All consumer component/hook files keep their existing `import { ... } from 'state/progressState'` imports unchanged. The barrel re-export through `state/progressState.ts` to `state/progress/index.ts` preserves the public API. The `progressState.ts` file can be removed in a follow-up PR, if desired, by updating all of the consumers import statements to pull directly from the new `state/progress/` modules. I chose not to do that here to keep this PR focused on the state refactoring only.

I also added tests for the new `helpers.ts` functions in `test/unit/progressState.helpers.test.ts`. All existing unit tests from the previous PR pass without modification.

[...]

PR #1474, opened 2026-03-04
Progress state import paths

### LINKED ISSUE #1473: Remove progressState.ts
This is very low priority but it is tech debt that we should eventually tackle.

From: https://github.com/saving-satoshi/saving-satoshi/pull/1460

> The `progressState.ts` file can be removed in a follow-up PR, if desired, by updating all of the consumers import statements to pull directly from the new state/progress/ modules. I chose not to do that here to keep this PR focused on the state refactoring only.

jamaljsr:
Closes #1473 

This is the follow-up PR mentioned in #1460 that removes the barrel re-export files (`state/progressState.ts` and `state/progress/index.ts`) by updating all consumer imports to pull directly from the specific submodules (`state/progress/actions`, `state/progress/selectors`, etc.).

The barrel re-exports were left in place during the refactor PR to keep that diff focused on the state restructuring. Now that the refactor has landed, these intermediate files just add indirection. Removing them makes it clear where each function actually lives and prevents future imports from going through unnecessary layers.

Each commit updates one category of consumers (app pages, components, hooks, content, types, UI, tests, state) to import directly from the specific `state/progress/` submodule that defines the symbol. The final commit deletes the now-unused `state/progressState.ts` and `state/progress/index.ts` barrel files.

The commits are structured so that each one is self-contained and the build passes at every step.

### Steps to test

This PR has no behavioral changes. Any issues with the changes made would be caught by the build process.

- `yarn build`
- `yarn test`
- `yarn e2e all`

You can also run the app and confirm lessons/chapters still work as expected.

## Pull Request checklist

- [x] Clean commit history


satsie (approved):
woooo hooo!!! looks great! goodbye progressState.ts! 👋 🫡

PR #1485, opened 2026-07-28
fix: remove chatbtc from link

0tuedon:
This PR removes all chatbtc links. since it has been shutdown


0tuedon:
The E2E test failure is related to the new update in github workflows I would create a fix in another PR and link that here

satsie:
Can you do me a favor and rebase now that I merged the fix for the workflow? I want to confirm that the E2E test workflow is fixed

0tuedon:
Rebased and the e2e tests works now @satsie

satsie (changes requested):
I think we should also remove the Holocat icon and speech bubbles. Having the icon and speech bubble with a link to a 404 honestly feels like a better experience than having it there with nothing to click. At least with the 404 the user can understand the reason for the UX.

Example: https://saving-satoshi-git-fork-0tuedon-remove-cha-8013a6-savingsatoshi.vercel.app/en/chapters/chapter-5/validate-signature-1

Image

satsie (approved):
Thanks for removing the Holocat image hover. I am undecided on if we should remove the entire `tooltip_one`, etc. objects in the `i18n/locales` files but maybe we can deal with that cleanup if and when we start adding links to Decoding Bitcoin.

